PR #522, opened 2014-02-10
fix building with musl libc

ncopa:
A set of patches needed to make uwsgi build on musl libc, which is fairly strict.

It is compile tested with musl libc and uClibc.

### DIFF
--- a/core/lock.c
+++ b/core/lock.c
@@ -99,7 +99,7 @@ retry:
 			uwsgi_log("unable to set PTHREAD_PRIO_INHERIT\n");
 			exit(1);
 		}
-		if (pthread_mutexattr_setrobust_np(&attr, PTHREAD_MUTEX_ROBUST)) {
+		if (pthread_mutexattr_setrobust(&attr, PTHREAD_MUTEX_ROBUST)) {
 			uwsgi_log("unable to make the mutex 'robust'\n");
 			exit(1);
 		}
@@ -161,7 +161,7 @@ void uwsgi_lock_fast(struct uwsgi_lock_item *uli) {
 #ifdef EOWNERDEAD
 	if (pthread_mutex_lock((pthread_mutex_t *) uli->lock_ptr) == EOWNERDEAD) {
 		uwsgi_log("[deadlock-detector] a process holding a robust mutex died. recovering...\n");
-		pthread_mutex_consistent_np((pthread_mutex_t *) uli->lock_ptr);
+		pthread_mutex_consistent((pthread_mutex_t *) uli->lock_ptr);
 	}
 #else
 	pthread_mutex_lock((pthread_mutex_t *) uli->lock_ptr);
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -1832,7 +1832,7 @@ void uwsgi_plugins_atexit(void) {
 
 void uwsgi_backtrace(int depth) {
 
-#if (defined(__linux__) && !defined(__UCLIBC__))|| (defined(__APPLE__) && !defined(NO_EXECINFO)) || defined(UWSGI_HAS_EXECINFO) 
+#if (defined(__GLIBC__) && !defined(__UCLIBC__))|| (defined(__APPLE__) && !defined(NO_EXECINFO)) || defined(UWSGI_HAS_EXECINFO)
 
 #include <execinfo.h>
 
--- a/uwsgi.h
+++ b/uwsgi.h
@@ -153,22 +153,7 @@ extern "C" {
 #endif
 #endif
 
-#ifndef _GNU_SOURCE
-#define _GNU_SOURCE
-#endif
-#include <stdio.h>
-#ifdef __UCLIBC__
-#include <sched.h>
-#endif
-#undef _GNU_SOURCE
-
-#include <stdlib.h>
-#include <stddef.h>
-#include <signal.h>
-#include <math.h>
-
-#include <sys/types.h>
-#ifdef __linux__
+#if defined(__linux__) || defined(__GNUC__)
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE
 #endif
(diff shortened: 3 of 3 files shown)


xrmx:
@ncopa any chance you can rebase on top of master?
 My only concern is about "always define _GNU_SOURCE for linux" where you are defining GNU_SOURCE only for __linux__, wondering how that change does affect gnu/kFreeBSD

ncopa:
It is now rebased against current master.

Good catch wrt gnu/kFreeBSD. I created a vm with debian kFreeBSD and indeed, it did break build there. I changed the `#ifdef __linux__` to `#if defined(__linux__) || defined(__GNUC__)`.
I also added an `#undef _GNU_SOURCE` at the end as _GNU_SOURCE causes problems for Python.h as mentioned earlier.

I have compile tested it in both alpine linux (musl libc) and debian kbsd. It builds at least.

xrmx:
@ncopa thanks!
@unbit probably a good time to merge this and catch any build breakage?

a-tal:
hello. I tried building @ncopa's branch today, and while it does compile in alpine with musl, it errors on start with `unable to set PTHREAD_PRIO_INHERIT`.

:+1: for musl support though, let me know if you need anything else

xrmx:
@a-tal thanks for testing!
Indeed pthread_mutexattr_setprotocol is not supported on musl:
http://git.musl-libc.org/cgit/musl/tree/src/thread/pthread_mutexattr_setprotocol.c
What about undefining EOWNERDEAD for musl?, see lock.c line 49

ncopa:
> What about undefining EOWNERDEAD for musl?, see lock.c line 49

I don't like that. What if future version of musl has support for PTHREAD_PRIO_INHERIT?

Also the PTHREAD_PRIO_INHERT will only prevent priority inversion when SCHED_FIFO or SCHED_RR is used. I could not find anything in code indicating that real-time priority threads is used anywhere, so I believe this could be a non-fatal error.
```diff
diff --git a/core/lock.c b/core/lock.c
index 20935b5..d1005bd 100644
--- a/core/lock.c
+++ b/core/lock.c
@@ -97,7 +97,9 @@ retry:
        if (uwsgi_pthread_robust_mutexes_enabled) {
                if (pthread_mutexattr_setprotocol(&attr, PTHREAD_PRIO_INHERIT)) {
                        uwsgi_log("unable to set PTHREAD_PRIO_INHERIT\n");
-                       exit(1);
+                       /* PTHREAD_PRIO_INHERIT will only prevent priority
+                          inversion when SCHED_FIFO or SCHED_RR is used, so
+                          this is non-fatal */
                }
                if (pthread_mutexattr_setrobust(&attr, PTHREAD_MUTEX_ROBUST)) {
                        uwsgi_log("unable to make the mutex 'robust'\n");
```

ncopa:
While at it, we could make a runtime error when setting robust make it fall back to standard pthread mutexes:
```diff
diff --git a/core/lock.c b/core/lock.c
index 20935b5..476a35e 100644
--- a/core/lock.c
+++ b/core/lock.c
@@ -97,11 +97,14 @@ retry:
        if (uwsgi_pthread_robust_mutexes_enabled) {
                if (pthread_mutexattr_setprotocol(&attr, PTHREAD_PRIO_INHERIT)) {
                        uwsgi_log("unable to set PTHREAD_PRIO_INHERIT\n");
-                       exit(1);
+                       /* PTHREAD_PRIO_INHERIT will only prevent priority
+                          inversion when SCHED_FIFO or SCHED_RR is used, so
+                          this is non-fatal */
                }
                if (pthread_mutexattr_setrobust(&attr, PTHREAD_MUTEX_ROBUST)) {
-                       uwsgi_log("unable to make the mutex 'robust'\n");
-                       exit(1);
+                       uwsgi_log("!!! unable to make the mutex 'robust'\n");
+                       uwsgi_log("!!! falling back to standard pthread mutexes !!!\n");
+                       uwsgi_pthread_robust_mutexes_enabled = 0;
                }
        }
 #endif
```

xrmx:
Il 06/08/2015 11:07, ncopa ha scritto:
> While at it, we could make a runtime error when setting robust make it fall back
> to standard pthread mutexes:

I don't think this change of behaviour should be made in this patchset, 
especially if it's changing behaviour on other configurations. I'd just go with 
the minimal changes to get uwsgi compile with musl and then we can reason about 
other changes.

bodgit:
Is there any plan to fix the `PTHREAD_PRIO_INHERIT` problem on musl? Compiling from master (there doesn't seem to be a release with these changes in yet) on Alpine Linux gets a binary provided I omit the ugreen plugin however I'm in the same boat as @a-tal in that the uwsgi binary won't actually run due to this fatal error.

xrmx:
@bodgit could you please check if applying this changes https://github.com/unbit/uwsgi/pull/522#issuecomment-128296571 works for you?

ncopa:
@bodgit I am not aware of any plans to implement `PTHREAD_PRIO_INHERIT`.

Returning `ENOTSUP` is also perfectly [valid](http://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutexattr_getprotocol.html#tag_16_443_05).

bodgit:
@xrmx Yes, making the change referenced in the comment and removing the `exit(1);` to make the `PTHREAD_PRIO_INHERIT` error non-fatal allows the binary to start up and so far I've not seen any further problems.

ncopa:
Why is the `PTHREAD_PRIO_INHERIT` there in the first place?

bodgit:
Is it worth creating a new PR with these additional changes so that it can be discussed separately?

xrmx:
@bodgit sure, i'd change the code to not return error only if ENOTSUP is returned as suggested by @ncopa. And keep his comment, maybe adding explicitly that musl does not implement that. Thanks.

bodgit:
Sure, will do.

PR #730, opened 2014-09-28
avg_response_time var calculation errors

jiathenine:


blkt:
Hi @unbit,

the one uWSGI is calculating **is not** the average between the last two requests' response_time: suppose you're really willing to calculate the average between the last two response_times, and suppose your sequence of timings was

180 190 200 210 300

Then your algorithm calculates  201.25 + 30 / 2 = 250.625, when the real value should be 216.

Also, the correct version of your algorithm is exceptionally useless if you're using uWSGI as dispatcher for, say, REST API (where different requests carry different burdens in terms of execution time) because the order is relevant, while in mean it should not (percentiles are a different matter).

Some very useful metrics to see what's happening behind the curtains are percentiles (50, 95 and 99 usually), of which [folsom](https://github.com/boundary/folsom) provides a very good implementation.

unbit:
Hmm yes, it "was" the the average of the last two responses, then it changed to be cumulative. Btw i am not against adding new computations/metrics, but i would like to avoid changing their meaning (even if the current avg_response_time is pretty dumb ;)

blkt:
No problem, there's still the way to calculate the average :D

PR #736, opened 2014-10-07
use _full write methods in rados plugin to truncate PUT request objects

gdahlm:
Use librados rados_write_full and rados_aio_write_full methods to ensure that PUT requests to existing objects are truncated if they are smaller than the original objects.  Mainly this change will reduce disk utilization on the ceph cluster.  The *_full methods do also offer atomic truncate and update operations.  As the offset is anchored at 0 in the current code there is no difference in behavior for the client.

### DIFF
--- a/plugins/rados/rados.c
+++ b/plugins/rados/rados.c
@@ -147,7 +147,7 @@ static int uwsgi_rados_put(struct wsgi_request *wsgi_req, rados_ioctx_t ctx, cha
                 char *body =  uwsgi_request_body_read(wsgi_req, UMIN(remains, 32768) , &body_len);
                 if (!body || body == uwsgi.empty) goto error;
 		if (uwsgi.async < 1) {
-			if (rados_write(ctx, key, body, body_len, off) < 0) {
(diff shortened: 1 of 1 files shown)


funny-falcon:
No, bug is in plugin.
Quote from documentation:

> rados_write_full
> The object is filled with the provided data. If the object exists, it is atomically truncated and then written.

> rados_aio_write_full
> The object is filled with the provided data. If the object exists, it is atomically truncated and then written. Queues the write_full and returns.

Currently plugin calles it in a cycle `while (remains > 0)`.
So, instead of writing with growing offset, plugin overwrite same object several times.
More over, it is strange that this pull request were merged, cause there is already call to `uwsgi_rados_delete` if file exists.

funny-falcon:
I'll put reverse of this commit into #1228

funny-falcon:
Perhaps, first call should be `rados_write_full` and following are `rados_write_append`,
but it still doesn't solve an issue if someone tries to write into same object simultaneously.
There is no way to solve it.

PR #781, opened 2014-11-24
Removed file permissions, and let umask magic happen instead.

JonathanHuot:
Overwriting umask is not a good idea because it is preventing req-logger to create log file readable by everyone (nxlog, logstash, ... plenty of differents users).

Also, umask is already in global configuration so it will be good to use it !

### DIFF
--- a/plugins/logfile/logfile.c
+++ b/plugins/logfile/logfile.c
@@ -43,7 +43,7 @@ static ssize_t uwsgi_file_logger(struct uwsgi_logger *ul, char *message, size_t 
 				logfile = ul->arg;
 			}
 
-			ul->fd = open(logfile, O_RDWR | O_CREAT | O_APPEND);
+			ul->fd = open(logfile, O_RDWR | O_CREAT | O_APPEND, S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH);
 			if (ul->fd >= 0) {
 				ul->configured = 1;
 			}	


AlexanderDobrodey:
Guys, could you please integrate this pull request because this bug is annoying!
Thank you

unbit:
The patch should go in the logfile plugin too. I just only need to check why umask was not used (maybe it is from times when the umask option did not exist)

unbit:
Hmm, wait not sure to really understand the issue. If we do not pass the mode of the file we ends (with the default 022 umask) with a file with 700 permissions. The umask ensures specific flags are not applied. The "classic" approach would be creating the file with 644 (and then let the umask do its work). (your problem here is that default mode is 640 in uwsgi)

Now, changing to 644 could be a good thing, but Is it not easier to call the "chmod" hook ?

JonathanHuot:
I tried with the following configuration. Let me know your thoughts, I'm not sure about the ideal solution. I think uWSG provide too many ways to setup file permissions really :-) It will be great if all plugins+core were working similarly.

```
$ cat /etc/sysconfig/uwsgi
umask 022
$ cat myapp.ini
[uwsgi]
uid = nginx
gid = nginx
logfile-chown = on
logfile-chmod = 644
log-reopen = on
logto2 = /data/log/api/uwsgi.log
daemonize = /data/log/api/uwsgi.log
req-logger = file:/data/log/api/uwsgi_access.log
(..)
```

Despite this setup above, I was not able to have read permissions for others to log files. It's why I'm using this Pull Request, but let me know if you have another solution available.

unbit:
Personally i prefer hooks (http://uwsgi-docs.readthedocs.org/en/latest/Hooks.html):

```ini
[uwsgi]
logto = foobar.log
hook-as-root = chmod:%(logto) 644
```

JonathanHuot:
OK I see your point.

BTW regarding this pull, don't you think it's better to add it to master? It will also make sense to use *umask* as a global permission rule instead of using a hook for each and every log file.

unbit:
Yes, but the mode should 644 not "undefined". The umask "blocks" permissions, so we need to give the maximum tolerable (and then reduce it with umask or hooks)

PR #795, opened 2014-12-17
OS X clang compiler look up.

estobbart:
OS X Version 10.9.5 with XCode w/ command line tools installed.

```
$ pip install uwsgi
build/uwsgi/uwsgiconfig.py", line 692, in __init__
    gcc_major = int(gcc_version.split('.')[0])
AttributeError: 'NoneType' object has no attribute 'split'
...
uwsgi/uwsgiconfig.py", line 695, in __init__
    raise Exception("you need a C compiler to build uWSGI")
```

### DIFF
--- a/uwsgiconfig.py
+++ b/uwsgiconfig.py
@@ -608,7 +608,7 @@ class uConf(object):
         gcc_version = spcall("%s -dumpversion" % GCC)
         if not gcc_version and GCC.startswith('gcc'):
             if uwsgi_os == 'Darwin':
-                GCC = 'llvm-' + GCC.split('-')[0]
+                GCC = 'llvm-' + GCC.split('-')[0] if '-' in GCC else GCC
             else:
                 GCC = 'gcc'
             gcc_version = spcall("%s -dumpversion" % GCC)


gformich:
Any status on this PR? Wondering if there is a plan to fix compilation on OS X.

Thank you

estobbart:
Some additional info on the problem..

    >>> os.environ.get('CC', sysconfig.get_config_var('CC'))`
    gcc-4.2
    >>> spcall('llvm-gcc-4.2 -dumpversion')
    None

Results in a compiler not found error, when it's really compiler version unknown.

    >>> spcall('llvm-gcc -dumpversion')
   4.2.1

The split fixes the dumpversion.

unbit:
Currently on OSX the best thing you can do is forcing a specific compiler with CC=xxx. Personally i always use 

CC=clang make

but others users could prefer others compilers

hhstore:
### my Mac happens similar error. i find a way fix it.

### my Mac env:

- os: osx 10.11.2
- py2: python2.7
    - pip2
    - virtualenv
- py3: python3.4
    - pip3
    - virtualenv

### when i install `uwsgi`,  it is really weird. pip2 is OK, but pip3 happens error!

```bash
pip2 install uwsgi   # it it OK !

pip3 install uwsgi   # happens error !
```

### fix it. for pip3 (python3.4)
- see here. [stackoverflow](http://stackoverflow.com/questions/11669594/error-while-installing-uwsgi-on-mac)
- use below cmd, it works for me.

```bash
# use this command.
CC=gcc pip3 install uwsgi
```

PR #805, opened 2015-01-07
Add reasons to worker self-shutdown messages.

akx:
This should make it easier to understand why a worker is shutting down.

New hotness:

> ...The work of process 18381 is done (reload-on-rss limit reached (11251712 >= 1048576)). Seeya!
> ...The work of process 18860 is done (max requests reached (1 >= 1)). Seeya!

vs the old and busted

> ...The work of process 13371 is done. Seeya!

We're allocating heap memory for the message, but since the process will unconditionally die just after using the memory, I don't think not cleaning it up is a problem?

## ABI change!

Also, this _does_ involve an ABI change for the `goodbye_cruel_world` function.

I tried to search Github for possible third-party uses of the function (and `gbcw_hook` too) and couldn't find any (other than stray copies of old versions of uWSGI source), so I figure it _should_ be safeish to change.

Also, considering the fact that the `uwsgi_log()` that _might_ access a bogus pointer and cause a segfault is the very last thing that's done before exiting the process, segfaulting _maybe_ isn't that bad?

@unbit, if you deem the ABI change a stupid idea, I'm more than willing to rewrite this in a form that adds a new API function, say, `goodbye_cruel_world_reason(const char *reason)` (and not pass the reason argument to `gbcw_hook()` to retain that ABI compatibility).

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -1196,15 +1196,24 @@ void uwsgi_close_request(struct wsgi_request *wsgi_req) {
 
 	if (uwsgi.max_requests > 0 && uwsgi.workers[uwsgi.mywid].delta_requests >= (uwsgi.max_requests + ((uwsgi.mywid-1) * uwsgi.max_requests_delta))
 	    && (end_of_request - (uwsgi.workers[uwsgi.mywid].last_spawn * 1000000) >= uwsgi.min_worker_lifetime * 1000000)) {
-		goodbye_cruel_world();
+		goodbye_cruel_world("max requests reached (%llu >= %llu)",
+			(unsigned long long) uwsgi.workers[uwsgi.mywid].delta_requests,
+			(unsigned long long) (uwsgi.max_requests + ((uwsgi.mywid-1) * uwsgi.max_requests_delta))
+		);
 	}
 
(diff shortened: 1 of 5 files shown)


unbit:
The API (not ABI ;) change is not a problem (in 2.1), that function is used only in internal plugins and in the core. But i would refactor it to have goodbye_cruel_world() as a vararg function:

```c
void goodbye_cruel_world(char *fmt, ...) {
    char buf[1024];
    vsnprintf(buf, 1024, ....);
    
    uwsgi_log("...The work of process %d is done (%s). Seeya!\n", ... buf ....);
    exit(0);
}
```

so we can directly do

```c
goodbye_cruel_world("reload-on-rss limit reached %llu > %llu", ....);
```

akx:
Sure, that's a good idea! I'll do that tomorrow :)

akx:
New version, @unbit :)

Also added a commit to allow "megabytes" parsing to read floating-point values. (It's erring on the side of caution with new behavior, only using floating-point parsing when a `.` is encountered in the input.)

> $ ./uwsgi --master --workers 1 --reload-on-rss 0.1 --http :2323 --min-worker-lifetime 0
> ...The work of process 20002 is done (reload-on-rss limit reached (11182080 >= 104857)). Seeya!

unbit:
Good work. merged. Thanks

PR #810, opened 2015-01-14
Add :use_ssl=>true to extconf.rb

tsuyo:
It seems the installer is on https://uwsgi.it/install (as of Jan 14, 2015). Without this change, "gem install uwsgi" does almost nothing.


unbit:
thanks a lot

PR #815, opened 2015-01-16
Add machinery to warn on unused return values

xrmx:
We have ~20 issues on coverity regarding missing error values, add machinery to catch them at compile time. We are hiding the config under the coverity buildconf until are warnings (as -Werror is in place) are fixed. gcc only for now, not sure if clang on mac support this, if so we can extend it to clang too.

### DIFF
--- a/buildconf/coverity.ini
+++ b/buildconf/coverity.ini
@@ -1,3 +1,4 @@
 [uwsgi]
 main_plugin = psgi,rack,lua,python,gevent,php,cgi,glusterfs,pty,xslt,msgpack,geoip,v8,pam,ldap,mono,jvm,ring,jwsgi,servlet,rados,coroae,pypy,gccgo
 inherit = base
+cflags = -DUWSGI_WARN_UNUSED_RESULT
--- a/uwsgi.h
+++ b/uwsgi.h
@@ -139,6 +139,11 @@ extern "C" {
 
 #define uwsgi_foreach_token(x, y, z, w) for(z=strtok_r(x, y, &w);z;z = strtok_r(NULL, y, &w))
 
+#ifdef UWSGI_WARN_UNUSED_RESULT
+#define WARN_UNUSED_RESULT  __attribute__ ((warn_unused_result))
+#else
+#define WARN_UNUSED_RESULT
+#endif
 
 #ifndef __need_IOV_MAX
 #define __need_IOV_MAX
@@ -3295,7 +3300,7 @@ int event_queue_add_file_monitor(int, char *, int *);
 struct uwsgi_fmon *event_queue_ack_file_monitor(int, int);
 
 
-int uwsgi_register_signal(uint8_t, char *, void *, uint8_t);
+int uwsgi_register_signal(uint8_t, char *, void *, uint8_t) WARN_UNUSED_RESULT;
 int uwsgi_add_file_monitor(uint8_t, char *);
 int uwsgi_add_timer(uint8_t, int);
 int uwsgi_signal_add_rb_timer(uint8_t, int, int);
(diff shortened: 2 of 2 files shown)


xrmx:
If you are fine with the marked functions i can push the fixes in this branch just ack it instead of merging.

unbit:
is not better to fix them ? If we mark them as returning int and then we do not check for return value, well it is a "bug". Eventually we can change them to void (if needed)

xrmx:
So here's a batch of patches silencing these warnings. Some code had comments clearly asking to not check some functions returns value, this has been preserved. In quite some places i haven't found a clear way to return errors and thus i've added a comment and ignored them. Please review.

unbit:
Wow, those ```if (...) {}``` are pretty horrible :) You convinced me btw, let's mark those function in uwsgi.h. But (if you can/want) remove the ```if (...) {}``` lines (otherwise i can do it after the merge)

xrmx:
If you can comment if any of what i've ignored that should be really checked that would be helpful. By asking to remove the ifs i assume you prefer to use an int variable to store the return value, correct?

unbit:
Nope, there are cases where the function return value is really useless, so just calling it will be more than enough. I will comment those part that do not need it

core/errors.c
```diff
@@ -69,7 +69,8 @@ void uwsgi_redirect_to_slash(struct wsgi_request *wsgi_req) {
 		redirect = uwsgi_concat3n(wsgi_req->path_info, wsgi_req->path_info_len, "/?", 2, wsgi_req->query_string, wsgi_req->query_string_len);
 		redirect_len = wsgi_req->path_info_len + 2 + wsgi_req->query_string_len;
         }
-        uwsgi_response_add_header(wsgi_req, "Location", 8, redirect, redirect_len);
+	// do not care about errors
+        if (uwsgi_response_add_header(wsgi_req, "Location", 8, redirect, redirect_len)) {};
```
unbit: useless

core/routing.c
```diff
@@ -743,8 +743,8 @@ static int transform_fixcl(struct wsgi_request *wsgi_req, struct uwsgi_transform
                 wsgi_req->write_errors++;
                 return -1;
         }
-	// do not check for errors !!!
-        uwsgi_response_add_header(wsgi_req, "Content-Length", 14, buf, ret);
+	// do not care about errors
+        if (uwsgi_response_add_header(wsgi_req, "Content-Length", 14, buf, ret)) {};
```
unbit: useless

core/uwsgi.c
```diff
@@ -4812,8 +4817,8 @@ char *uwsgi_get_dot_h() {
 		free(base);
 		return "";
 	}
-	// add final null byte
-	uwsgi_buffer_append(ub, "\0", 1);
+	// add final null byte, do not care about errors
+	if (uwsgi_buffer_append(ub, "\0", 1)) {};
```
unbit: useless

plugins/cache/cache.c
```diff
@@ -225,7 +225,8 @@ static int uwsgi_cache_request(struct wsgi_request *wsgi_req) {
                                 argc = 3;
                                 if (!uwsgi_parse_array(wsgi_req->buffer, wsgi_req->uh->_pktsize, argv, argvs, &argc)) {
                                         if (argc > 1) {
-						uwsgi_cache_magic_set(argv[0], argvs[0], argv[1], argvs[1], 0, 0, NULL);
+						// do not care about errors
+						if (uwsgi_cache_magic_set(argv[0], argvs[0], argv[1], argvs[1], 0, 0, NULL)) {};
```
unbit: useless

plugins/cache/cache.c
```diff
@@ -233,7 +234,8 @@ static int uwsgi_cache_request(struct wsgi_request *wsgi_req) {
                 case 2:
                         // del
                         if (wsgi_req->uh->_pktsize > 0) {
-                                uwsgi_cache_magic_del(wsgi_req->buffer, wsgi_req->uh->_pktsize, NULL);
+				// do not care about errors
+                                if (uwsgi_cache_magic_del(wsgi_req->buffer, wsgi_req->uh->_pktsize, NULL)) {};
```
unbit: useless

plugins/cgi/cgi_plugin.c
```diff
@@ -323,7 +323,8 @@ static int uwsgi_cgi_parse(struct wsgi_request *wsgi_req, int fd, char *buf, siz
 					}
 				}
 
-				uwsgi_response_add_header(wsgi_req, NULL, 0, key, header_size);
+				// do not care about errors
+				if (uwsgi_response_add_header(wsgi_req, NULL, 0, key, header_size)) {};
```
unbit: useless

plugins/lua/lua_plugin.c
```diff
@@ -822,13 +823,15 @@ static int uwsgi_lua_request(struct wsgi_request *wsgi_req) {
 				}
 
 				http2 = lua_tolstring(L, -1, &slen2);
-				uwsgi_response_add_header(wsgi_req, (char *) http, slen, (char *) http2, slen2);
+				// do not care about errors
+				if (uwsgi_response_add_header(wsgi_req, (char *) http, slen, (char *) http2, slen2)) {};
```
unbit: useless

plugins/lua/lua_plugin.c
```diff
@@ -822,13 +823,15 @@ static int uwsgi_lua_request(struct wsgi_request *wsgi_req) {
...
 				lua_pop(L, 1);
 			}
 		}
 		else {
 			http2 = lua_tolstring(L, -1, &slen2);
-			uwsgi_response_add_header(wsgi_req, (char *) http, slen, (char *) http2, slen2);
+			// do not care about errors
+			if (uwsgi_response_add_header(wsgi_req, (char *) http, slen, (char *) http2, slen2)) {};
```
unbit: useless

plugins/mono/mono_plugin.c
```diff
@@ -169,13 +169,15 @@ static void uwsgi_mono_method_SendStatus(MonoObject *this, int code, MonoString
 	char status_code[4];
 	uwsgi_num2str2n(code, status_code, 4);
 	char *status_line = uwsgi_concat3n(status_code, 3, " ", 1, mono_string_to_utf8(msg), mono_string_length(msg));
-	uwsgi_response_prepare_headers(wsgi_req, status_line, 4 + mono_string_length(msg));
+	// do not care about errors
+	if (uwsgi_response_prepare_headers(wsgi_req, status_line, 4 + mono_string_length(msg))) {};
```
unbit: useless

plugins/mono/mono_plugin.c
```diff
@@ -169,13 +169,15 @@ static void uwsgi_mono_method_SendStatus(MonoObject *this, int code, MonoString
...
 	free(status_line);
 }
 
 static void uwsgi_mono_method_SendUnknownResponseHeader(MonoObject *this, MonoString *key, MonoString *value) {
 	struct wsgi_request *wsgi_req = current_wsgi_req();
-	uwsgi_response_add_header(wsgi_req, mono_string_to_utf8(key), mono_string_length(key), mono_string_to_utf8(value), mono_string_length(value));
+	// do not care about errors
+	if (uwsgi_response_add_header(wsgi_req, mono_string_to_utf8(key), mono_string_length(key), mono_string_to_utf8(value), mono_string_length(value))) {};
```
unbit: useless

xrmx:
Then it won't compile because of -Werror :)

plugins/rack/rack_plugin.c
```diff
@@ -781,7 +781,8 @@ VALUE send_header(VALUE obj, VALUE headers) {
 	}
 
 	if (cnt > 0) {
-		uwsgi_response_add_header(wsgi_req, RSTRING_PTR(hkey), RSTRING_LEN(hkey), this_header, cnt);
+		// do not care about errors
+		if (uwsgi_response_add_header(wsgi_req, RSTRING_PTR(hkey), RSTRING_LEN(hkey), this_header, cnt)) {};
```
unbit: useless

plugins/xslt/xslt.c
```diff
@@ -329,9 +329,9 @@ static int transform_toxslt(struct wsgi_request *wsgi_req, struct uwsgi_transfor
         char *output = uwsgi_xslt_apply( doc, utxc->stylesheet->buf, utxc->params ? utxc->params->buf : NULL, &rlen);
 	if (!output) goto end;
 
-	// do not check for errors !!!
+	// do not care about errors !!!
 	if (ut->round == 1) {
-        	uwsgi_response_add_content_type(wsgi_req, utxc->content_type->buf, utxc->content_type->pos);
+        	if (uwsgi_response_add_content_type(wsgi_req, utxc->content_type->buf, utxc->content_type->pos)) {};
```
unbit: useless

plugins/transformation_gzip/gzip.c
```diff
@@ -38,8 +38,8 @@ static int transform_gzip(struct wsgi_request *wsgi_req, struct uwsgi_transforma
 	utgz->len += ub->pos;
 	uwsgi_buffer_map(ub, gzipped, dlen);
 	if (!utgz->header) {
-		// do not check for errors !!!
-        	uwsgi_response_add_header(wsgi_req, "Content-Encoding", 16, "gzip", 4);
+		// do not care about errors !!!
+        	if (uwsgi_response_add_header(wsgi_req, "Content-Encoding", 16, "gzip", 4)) {};
```
unbit: useless

plugins/transformation_chunked/chunked.c
```diff
@@ -17,8 +17,8 @@ static int transform_chunked(struct wsgi_request *wsgi_req, struct uwsgi_transfo
 	}
 
 	if (ut->round == 1) {
-		// do not check for errors !!!
-        	uwsgi_response_add_header(wsgi_req, "Transfer-Encoding", 17, "chunked", 7);
+		// do not care about errors !!!
+        	if (uwsgi_response_add_header(wsgi_req, "Transfer-Encoding", 17, "chunked", 7)) {};
```
unbit: useless

plugins/rpc/rpc_plugin.c
```diff
@@ -117,8 +117,9 @@ static int uwsgi_rpc_xmlrpc(struct wsgi_request *wsgi_req, xmlDoc *doc, char **a
 		xmlFree(xmlbuf);
 		return -1;
 	}
-        uwsgi_response_add_content_length(wsgi_req, xlen);
-        uwsgi_response_add_content_type(wsgi_req, "application/xml", 15);
+	// do not care about errors
```
unbit: in this part a goto error would be better

xrmx:
I've moved the rpc/plugin change to a newer commit handling the error. To move this branch forward another option to avoid the empty ifs is to add the warning annotation to wrappers of the original functions so a silent version that does not require tricks is available. What do you think?

PR #820, opened 2015-01-23
core/static: print mimetype of file in debug print

xrmx:
Something is off with mimetype handling, to easier debug print it to the file serve debug print


xrmx:
@unbit should we merge or close this?

PR #839, opened 2015-02-13
psgi_loader: fix a one off bug in XS_input_seek()

ikruglov:


plugins/psgi/psgi_loader.c
```diff
@@ -12,7 +12,7 @@ XS(XS_input_seek) {
 	struct wsgi_request *wsgi_req = current_wsgi_req();
 
         psgi_check_args(1);
```
unbit: here we should check for 2 args too (instead of 1)

ikruglov: why would one check params which are not used?

avar: What this does is make sure you have at least 1 (argument), i.e. the object (0) and the position.

Now you can just call $input-seek() and it won't error out.

@unbit: The full interface is $input->seek(0, [$whence]), where $whence is SEEK_{SET,CUR,END}. See "perldoc -f seek", any chance of supporting that?

I don't need it for anything, just curious.

ikruglov:
Please merge the patch! It fixes real bug in psgi plugin.

unbit:
Oh sorry, did not notice you already fixed the check_args value

ikruglov:
Thanks!

PR #848, opened 2015-02-20
Don't try to compress empty responses.

funkybob:
As discussed on IRC, basic testing has shown this functions as expected.


funkybob:
Please also include this in the next 2.0 release :)

Issue #564, opened 2014-03-14
config from plugins (eg. python:// for statically link python)


anthonyrisinger:
i think it would be **REALLY** useful to activate plugins as a config source, specifically:

`--json python://package.module:callable()`

...would use `import package.module; return package.module.callable()` as JSON config, whereas:

`--ini python://package.module:object`

...would use `import package.module; return package.module.object` as INI config (note: this mirrors the `--module`, `--callable`, and `--mount` syntax)

this would allow me to move our entire config to python (with access to `uwsgi.opt` hopefully!)... combined with a writable `uwsgi.opt` (that future config can see!) i'd be ecstatic!

PR #852, opened 2015-02-23
introduce uwsgi.binary_argc

xtfxme:
if uWSGI is embedded in some way, part of argv might not be intended for processing by `uwsgi_commandline_config`, eg. consider the `pyuwsgi` plugin:

```
python2 -c 'import uwsgi as u; u.run()' --master -s@nil
```

in such cases, uWSGI should ignore some number of arguments (set by a plugin's `load()`, if necessary). this patch allows binary reloading to work again.

i have a working patch to `plugins/pyuwsgi/pyuwsgi.c` that makes use of this, but i want to clean it up a little prior to submitting. since this touches `uwsgi.h`, i mainly want to ensure it (or another solution) lands before 2.1 drops.

ASIDE: my work with `pyuwsgi` stems from a desire to do this: https://github.com/unbit/uwsgi/issues/564

`pyuwsgi` allows our apps to use their own settings (in python code) to configure uWSGI. while i think this is particularly elegant (and IMO, preferred), `--early-python-*` options also suggest `python://` as config might be possible now, and would achieve a similar effect. like i said, i prefer the current approach, and if acceptable, i will either add the `pyuwsgi` patch here or another ticket, whichever you prefer. once both of those land, i can update `pyuwsgi` to be python3 compatible.

thoughts?

### DIFF
--- a/core/init.c
+++ b/core/init.c
@@ -62,6 +62,7 @@ void uwsgi_init_default() {
 
 	uwsgi.cpus = 1;
 	uwsgi.new_argc = -1;
+	uwsgi.binary_argc = 1;
 
 	uwsgi.backtrace_depth = 64;
 	uwsgi.max_apps = 64;
@@ -261,11 +262,33 @@ void uwsgi_commandline_config() {
 	int argc = uwsgi.argc;
 	char **argv = uwsgi.argv;
 
+	// we might want to ignore some arguments not meant for us
+	char binary_argv0_pretty[256] = {'\0'};
+	char *binary_argv0_actual = NULL;
+
 	if (uwsgi.new_argc > -1 && uwsgi.new_argv) {
 		argc = uwsgi.new_argc;
 		argv = uwsgi.new_argv;
 	}
 
+	if (uwsgi.binary_argc > 1 && argc >= uwsgi.binary_argc) {
+		char *pretty = (char *)binary_argv0_pretty;
+		strncat(pretty, argv[0], 255);
+
+		for (i = 1; i < uwsgi.binary_argc; i++) {
+			if (strlen(pretty) + 1 + strlen(argv[i]) + 1 > 256)
+				break;
+
+			strcat(pretty, " ");
+			strcat(pretty, argv[i]);
+		}
+
+		argc -= uwsgi.binary_argc - 1;
+		argv += uwsgi.binary_argc - 1;
+
+		binary_argv0_actual = argv[0];
+		argv[0] = (char *)binary_argv0_pretty;
+	}
 
 	char *optname;
 	while ((i = getopt_long(argc, argv, uwsgi.short_options, uwsgi.long_options, &uwsgi.option_index)) != -1) {
@@ -289,6 +312,8 @@ void uwsgi_commandline_config() {
 		add_exported_option(optname, optarg, 0);
 	}
 
+	if (binary_argv0_actual != NULL)
+		argv[0] = binary_argv0_actual;
 
 #ifdef UWSGI_DEBUG
 	uwsgi_log("optind:%d argc:%d\n", optind, uwsgi.argc);
--- a/uwsgi.h
+++ b/uwsgi.h
@@ -2811,6 +2811,9 @@ struct uwsgi_server {
 	int spooler_signal_as_task;
 
 	int log_worker;
+
+	// number of args to consider part of binary_path
+	int binary_argc;
 };
 
 struct uwsgi_rpc {


xtfxme:
as a concrete example of what this fixes (`setup.cpyext.py` builds uwsgi as a python module):

```console
# python2 setup.cpyext.py build
# PYTHONPATH=build/lib.linux-x86_64-2.7 python2 -c 'import uwsgi as u; u.run()' --master -s@nil &
# kill -HUP ${!}
```

the process will fail with something like this:

```
closing all non-uwsgi socket fds > 2 (max_fd = 1024)...
found fd 3 mapped to socket 0 (@nil)
running /usr/bin/python2.7
Unknown option: --
usage: /usr/bin/python2.7 [option] ... [-c cmd | -m mod | file | -] [arg] ...
Try `python -h' for more information.
```

because the `pyuwsgi` plugin currently must skip all arguments between `argv[0]` and the last occurrence of `-c` or `-m`, else `uwsgi_commandline_config` will fail:

https://github.com/unbit/uwsgi/blob/master/plugins/pyuwsgi/pyuwsgi.c#L226

so, does this seem like a sound? patch is minimal :) but i can certainly attach the `pyuwsgi` patch too, if requested.

xtfxme:
would adding a patch to `pyuwsgi` plugin help accelerate toward merge (demonstrating more concretely why it's needed)?

i just want to make sure i don't miss the 2.1 window -- the `pyuwsgi` patch could come whenever, but since this touches `uwsgi.h` i'm trying to make sure it's both the best way and available for use later on.

unbit:
Yes, make a patch for pyuwsgi. Do not be worried for 2.1, it is expected not before the summer :)

xtfxme:
So I realized that summer is here, so time must be short!

aside: `interp_argc` might be clearer than `binary_argc`

I've updated the patch to move `uwsgidecorators.py` to `plugins/pyuwsgi`, and just told `setup*.py` to use that as the root of the project. This cleans up the root dir slightly, and gives a place for additional language-specific integration files.

`plugins/python` may seem like a better place, but my purpose here is to make `pyuwsgi` a first-class plugin, and the default entrypoint for python integration. I also intend to add an actual `uwsgi.py` module there (or a package, eventually `uwsgidecorators.py` could be `uwsgi/decorators.py`), and the proper `python` plugin simply adds objects to that. The point is, `python` plugin is all about how uWSGI interacts with python while *running*, and `pyuwsgi` is about how python itself configures and runs uWSGI.

Ultimately I'm trying to make `setup.cpyext.py` and `pyuwsgi.c` compatible with all pythons uWSGI runs with (2.7, 3.x, pypy, etc), but if that can't be done soon enough, could we add another `UWSGI_*` environment var to control which setup file is used?

This patch makes several new things possible:

* pip installs a proper python extentsion module
* standard setuptools entrypoints are generated
* graceful reloading works again (uWSGI can now be told to ignore some arguments)
* allows uWSGI to be configured by the host process directly (python)
* **makes testing uWSGI easier**, eg. http://hastebin.com/gelibimevu.py
```python
    uwsgi.setup(
            '--uwsgi-socket=fd://{}'.format(server.fileno()),
            '--module=__main__:app',
            )
    uwsgi.register_rpc('rpc', rpc)
    uwsgi.run()
```

This patch makes `uwsgi` *always* importable -- if you are not running under it, you'll have the option to start it up!

I propose this implementation act as a reference API for other language runtimes desiring this feature, to complement the the APIs uWSGI adds after it's running. Currently adds two methods:

```python
_uwsgi.setup([sys.executable, '-muwsgi'], ['--help'])
_uwsgi.run()
```

where `setup` accepts two sequences, which combined form `argv`. The reason for separation is that uWSGI must have a way of knowing `-muwsgi` is an argument for python, not itself.

I wanted to update before going too much further, and get design goals/ideas... like would it better for pypy/cpython compat if everything was done from `uwsgi.py` using ctypes/cffi?

Let me know where is should go from here -- I'm working on python3 compat -- all the legwork is done and we have been using `setup.cpyext.py` builds for a long time now without issue, and I'd like our teams to be able to start using with native pip (even if they must export an environ var). I reallly think people will find the ability to configure/test easier a big win, along with idiomatic integration with python. It could make testing uWSGI itself easier to, by setting up test contexts more directly.

xtfxme:
I haven't added `uwsgi.py` yet to bridge the gap, but you can try it out by running `python2 setup.cpyext.py build` then:

```sh
PYTHONPATH=build/lib.linux-x86_64-2.7 python2 -c 'import _uwsgi; print(_uwsgi.argv); _uwsgi.setup(*_uwsgi.argv); print(_uwsgi.new_argv); _uwsgi.run()' --socket=@nil --master &
```

Then just `kill -HUP $!` and uWSGI will properly perform a self-referential graceful reload.

xtfxme:
it might be possible to create a `uwsgi.py` that can properly expose `uwsgi_setup` and `uwsgi_run` to any Python impl... then we could move everything to `plugins/python` and drop `pyuwsgi` completely. Would this be of interest?

At the end of the day, I just want to be able to initialize and run uWSGI, in two distinct steps, from a host Python application -- how it actually works is not important.

unbit:
Finally got the time to review it. 

I think the only step missing is moving the binary_argc struct field to the end of the structure to retain ABI compatibility. The rest pf the patch is good. Thanks a lot and sorry for the huge amount of time required :)

xtfxme:
@unbit please see final, single, simple patch

i dropped all the shuffling about... i now think it's best to just drop `pyuwsgi` altogether, and instead work on integrating it's functionality into the existing python plugins. I think we can achieve this via CFFI, and at the same time unify how uWSGI is bootstrapped in a Python environment (where `_uwsgi` is a CFFI-based module)... but i'll go on about this somewhere else :)

this pull request lays the groundwork for all that by making it possible for plugins to better control command line processing.

unbit:
so from the cffi point of view the only requirements is setting uwsgi.binary_argc to the right value ?

xtfxme:
yeah, the plugin would likely set `new_argv` as well, but this just gives plugins a way to say "here are *all* the arguments needs to start uWSGI from a cold state, but the first N are not for uWSGI"

this also gives a way to completely disable uWSGI CLI processing by setting `uwsgi.binary_argc = argc`.

when the plugin is loaded it sets `uwsgi.binary_argc` to an appropriate value. in my embedded python example, by this point, python already is running and someone is simply trying to initialize uWSGI in the same process, and they are explicitly passing all of argv (so the function setting up uWSGI *from python* has enough information to set `new_argv`, `new_argc`, and `binary_argc`.

PR #870, opened 2015-03-16
Fix grammar

sammko:
Grammar is important! The reflexive pronoun `I` should always be capitalized, and no spaces should be present before punctuation.


unbit:
Thanks :)

PR #873, opened 2015-03-29
Be more helpful when a request has an unknown modifier1

akx:
Adds a list of "well-known" modifier1 values and changes the

> -- unavailable modifier requested: %d --

message to add a hint about which plugin might be needed to
fix the situation.

(Missing plugins seems to be a common enough problem.)

### DIFF
--- a/core/known_plugins.c
+++ b/core/known_plugins.c
@@ -1,0 +1,70 @@
+#include "uwsgi.h"
+
+typedef struct {
+    int modifier;
+    const char *usage;
+    const char *plugin_name;
+} map_entry_t;
+
+#define BEGIN_KNOWN_MODIFIER_MAP(name) static map_entry_t* name[] = {
+#define END_KNOWN_MODIFIER_MAP() NULL };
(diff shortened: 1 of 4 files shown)


akx:
@unbit Seems this PR has gone stale -- do you think it could get merged if I rebase it?

unbit:
the mapping of modifiers is becaming pretty "opaque" recently with the proliferation of third party plugins. They are mainly a suggestions, and lot of users tend to overwrite the zero one to simplify setups. I do not know if it is a good idea to add hardcoded values. I leave this open for futher discussion.

Thanks as always

akx:
Well, if I add support for the modifier remap table here?

(Aside: I feel it would be a good idea to have a table of known "contrib"/external modifier1s too, to avoid conflicts (even if they are eventually inevitable).)

xrmx:
Not scrictly related to the patch but speaking of plugins, what about adding a plugins registry? Something like a json hosted in this repo that lists plugin name, description, modifier, url, whatever. Plus some script that add automatically the bundled ones. We can then add a human friendly js app on top.

akx:
Closing to clean up my PR list ✌️

PR #875, opened 2015-03-31
 catch exception in executing post fork chain so that exceptions dont

triggity:
Right now, if any function in the fork chain throws an exception, the exception is not caught, and the rest of the functions never get called. while just printing an exception is not desirable, it should allow the rest of the decorators to continue, agree?

### DIFF
--- a/uwsgidecorators.py
+++ b/uwsgidecorators.py
@@ -70,7 +70,11 @@ def manage_spool_request(vars):
 
 def postfork_chain_hook():
     for f in postfork_chain:
-        f()
+        try:
+            f()
+        except Exception as e:
+            print e
+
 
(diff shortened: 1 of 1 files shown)


unbit:
Sorry for the late. Well i am not quite sure about blindly going on. Maybe it should be a flag on the decorator ? @aldur what you think ?

aldur:
Yes, the flexibility of the flag could be our best option here. @triggity, how would you like to proceed?

triggity:
a flag sounds fine. would it be one that determined wether to catch the exception?

aldur:
Exactly :)

triggity:
what would you like the flag to be called?

aldur:
What do you think of `catch_exception=bool`?

PR #881, opened 2015-04-09
plugins/psgi - remove the need for Scalar::Util

JRaspass:
Commit ff9bab4b fixed support for sending IO like handles or GLOBs
via sendfile by first calling fileno on them. This was checked
by calling Scalar::Util::reftype on the reference first, as per
https://metacpan.org/pod/PSGI#Body . This may be sane in pure
perl, but there are better methods in XS/C than string evalling
some Perl, which will eventually also call some XS/C.

This commit introduces more robust checking of whether we have a
"real" filehandle, and also introduces a test that chucks a bunch
of different scalars and references at uWSGI to make sure it doesn't
choke. I wasn't sure how to hook this test up to travis, or wether
that's even desireable, so just left the test in what I believe to
be the right directories.

Because this commit massively simplifies the code, it also had an
effect on object size, on my machine (x64, GCC 4.9), psgi_plugin.o
went from 63360 bytes to 62296 bytes.

The indentation level of the ->getline for loop has been left at
its original level, which is now incorrect, to make the diff
easier to read. A whitespace cleanup commit could follow.

In related news, does anyone know the purpose of the three other
use lines in psgi_loader.c, two of them were added back in 2011
by cf08972e, but I can't work out what their inclusion has to do
with memleak hunting.

Basically I would love to remove these lines as I don't believe
they're needed for uWSGI to function, and if apps running under
uWSGI need them, they should load them themselves.

Also related, would more PSGI refactorings be welcome? I get the
impression that Perl is a bit of a second class citizen in uWSGI
these days, and as both an avid user of uWSGI+PSGI and a hobby XS
developer, I'd love to clean up more of the PSGI plugin.

### DIFF
--- a/plugins/psgi/psgi.h
+++ b/plugins/psgi/psgi.h
@@ -83,7 +83,6 @@ int psgi_response(struct wsgi_request *, AV*);
 
 SV *uwsgi_perl_obj_call(SV *, char *);
 int uwsgi_perl_obj_can(SV *, char *, size_t);
-int uwsgi_perl_obj_isa(SV *, char *);
 int init_psgi_app(struct wsgi_request *, char *, uint16_t, PerlInterpreter **);
 PerlInterpreter *uwsgi_perl_new_interpreter(void);
 int uwsgi_perl_mule(char *);
--- a/plugins/psgi/psgi_loader.c
+++ b/plugins/psgi/psgi_loader.c
@@ -442,7 +442,6 @@ int init_psgi_app(struct wsgi_request *wsgi_req, char *app, uint16_t app_len, Pe
 		perl_eval_pv("use IO::Handle;", 1);
 		perl_eval_pv("use IO::File;", 1);
 		perl_eval_pv("use IO::Socket;", 1);
-		perl_eval_pv("use Scalar::Util;", 1);
 
 		if (uperl.argv_items || uperl.argv_item) {
 			AV *uperl_argv = GvAV(PL_argvgv);
--- a/plugins/psgi/psgi_plugin.c
+++ b/plugins/psgi/psgi_plugin.c
@@ -254,34 +254,6 @@ int uwsgi_perl_obj_can(SV *obj, char *method, size_t len) {
 
 }
 
-int uwsgi_perl_obj_isa(SV *obj, char *class) {
-
-	int ret = 0;
-
-        dSP;
-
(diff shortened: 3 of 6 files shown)


unbit:
Thanks for your patch, i will review it in the week end but at first look it loogs good. 

Language plugins (not only perl) in uWSGI are in a sort of freeze by months, as we are in a state like "they work, it required ages for them to be robust, do not touch them" ;) In 2.1 there is space for new approaches, optimizations, refactoring and features, so if you want to invest a bit in cleaning up (and improving) the psgi plugin, i will merge it for sure in the master branch.

Regarding those use statements, well, i have no idea :)  i think they are related to some corner-case use of Plack years ago. We were in rush as a bunch of customers reported memory leaks and malfunctions after being in production, so we did lot of hacks and brutal tests :( 

Just a couple of note if you plan to heavily work on the psgi plugin:

* unfortunately we need to support threads, the current code/way for managing them is really horrible in the psgi plugin, probably there is space for cleaning up
* we need to support at least perl 5.8 (customers still on ancient centos)
* ensure after any modification the Coro plugin still works (it is used a lot)

Thanks again :)

JRaspass:
Thanks for the quick reply, based on your feedback I'm currently extending the test to test multiple perl versions, from ancient 5.8.9 up to present day, just having a little difficulty getting perlbrew and uWSGI to play nice together :-P

I'll probably amend this commit when the test is ready as I must confess I've only tried it on a modern, non threaded perl. Either way, don't let this halt the review, I'm keen to hear what you (or others) think, my C is certainly more rusty than my Perl :-S

JRaspass:
Just a quick update, I've got the test running under perlbrew, it currently runs threaded and non threaded variants of perl 5.20.2 and 5.8.9, I will be added more in between versions shortly.

The good news is that threaded worked fine, but surprise surprise, 5.8.9 doesn't compile. It doesn't look too serious, probably just needs a define. If all goes well I'll have an updated commit/PR up either tonight or tomorrow.

JRaspass:
Revamped the test, minor code change, all tested perl versions (5.20.2, 5.18.4, 5.16.3, 5.14.4, 5.12.4, 5.10.1, 5.8.9) with and without threads pass :-)

unbit:
All looks fine, thanks

Issue #882, opened 2015-04-10
GridFS plugin is not compatible with mongo-cxx-driver


voxik:
```
python uwsgiconfig.py --plugin plugins/gridfs fedora
using profile: buildconf/fedora.ini
detected include path: ['/usr/lib/gcc/x86_64-redhat-linux/5.0.0/include', '/usr/local/include', '/usr/include']
*** uWSGI building and linking plugin plugins/gridfs ***
[gcc -pthread] gridfs_plugin.so
plugins/gridfs/gridfs.cc: In function 'void uwsgi_gridfs_do(wsgi_request*, uwsgi_gridfs_mountpoint*, char*, int)':
plugins/gridfs/gridfs.cc:45:3: error: 'scoped_ptr' is not a member of 'mongo'
   mongo::scoped_ptr<mongo::ScopedDbConnection> conn( mongo::ScopedDbConnection::getScopedDbConnection(ugm->server, ugm->timeout) );
   ^
plugins/gridfs/gridfs.cc:45:3: note: suggested alternative:
In file included from /usr/include/boost/scoped_ptr.hpp:14:0,
                 from /usr/include/mongo/logger/message_log_domain.h:18,
                 from /usr/include/mongo/logger/logger.h:19,
                 from /usr/include/mongo/util/assert_util.h:27,
                 from /usr/include/mongo/bson/bsontypes.h:20,
                 from /usr/include/mongo/bson/bsonelement.h:25,
                 from /usr/include/mongo/bson/bsonobj.h:27,
                 from /usr/include/mongo/client/bulk_operation_builder.h:20,
                 from /usr/include/mongo/client/dbclientinterface.h:29,
                 from /usr/include/mongo/client/dbclient_rs.h:23,
                 from /usr/include/mongo/client/dbclient.h:59,
                 from plugins/gridfs/gridfs.cc:3:
/usr/include/boost/smart_ptr/scoped_ptr.hpp:41:25: note:   'boost::scoped_ptr'
 template<class T> class scoped_ptr // noncopyable
                         ^
plugins/gridfs/gridfs.cc:45:21: error: 'ScopedDbConnection' is not a member of 'mongo'
   mongo::scoped_ptr<mongo::ScopedDbConnection> conn( mongo::ScopedDbConnection::getScopedDbConnection(ugm->server, ugm->timeout) );
                     ^
plugins/gridfs/gridfs.cc:45:61: error: 'mongo::ScopedDbConnection' has not been declared
[...]

PR #884, opened 2015-04-10
Updated GridFS plugin

WeirdCarrotMonster:
Fixes https://github.com/unbit/uwsgi/issues/882

I haven't implemented pooling, but plugin itself works. I think it would be better to have working version now.


unbit:
I am a bit unsure about forcing c++11, is it truly required ?

WeirdCarrotMonster:
No, i only use it for unique_ptr. It can be replaced with auto_ptr, but as far as i know auto_ptr is deprecated.

unbit:
Yes, auto_ptr is deprecated, at this point i think we can merge it as-is

unbit:
Merged, thanks again

PR #886, opened 2015-04-11
Tidy with flake8

funkybob:
There was also one bug in uwsgiconfig.py with a split( ... , 2) that should have been a 1.
Finally, uwsgiconfig.py refers to plugin_is_remote but it should probably be is_remote_plugin


unbit:
Thanks, the plugin_is_remote/is_remote_plugin has been fixed too

PR #888, opened 2015-04-15
php7 build fix, don't forget to add "default php" as plugin name while b...

mow:
php7 build fix, don't forget to add "default php" as plugin name while building, otherwise an "undefined symbol" error raises (UWSGICONFIG_PHPDIR=/opt/php7 python uwsgiconfig.py --plugin plugins/phpng default php)


unbit:
Hi, instead of adding plugins to the core repository we prefer to split them in different repository. Feel free to create a uwsgi-phpng one and we will add it to the official site.

Remember that the new (robust) way of building plugins is

```sh
uwsgi --build-plugin <dir>
```

mow:
Thank you for fast response :)

mow:
Hi again, here is the separate repository https://github.com/mow/uwsgi-php7

unbit:
Great, starred and spammed on twitter :)

mow:
:D thanks

PR #891, opened 2015-04-16
plugins/psgi - refactor env population

JRaspass:
Use PL_sv_yes & PL_sv_no instead of newSViv(1) & newSviv(0) for bools
as they're interpreter globals that don't need to be created everytime.

Like how psgix.harakiri was already handled, don't bother creating a
$env hash key just to store false, $env->{key} returns false regardless
of whether the key exists if the value is false.

In keeping with that theme, psgi.run_once is now never set as it was
always false.

Added a test and refactored the tests that were added with a1e30157,
all .t files in t/perl are now ran under all major perl versions with
and without threading when t/perl/run is invoked.

On my box, psgi_plugin.o went from 62,296 bytes to 62,112 bytes. Not a
huge saving I'll grant you, but every little helps :-)


unbit:
I do not think removing those "required" keys is a good idea (maybe @miyagawa can give his opinion here) , the PL_sv_yes thing instead is absolutely ok

miyagawa:
@unbit correct, storing the false value for a required key is meaningful in the PSGI spec.

However [the PSGI doc](https://metacpan.org/pod/PSGI) says:

>   When an environment key is described as a boolean, its value MUST conform to Perl's notion of boolean-ness. This means that an empty string or an explicit 0 are both valid false values. If a boolean key is not present, an application MAY treat this as a false value.

For non-required keys, omitting the entry altogether to mean false should be accepted. Although personally, being explicit by saving false is also a good idea.

JRaspass:
I've amended the commit and it's message, it now always set's every key, there's still a slight object size saving.

PR #893, opened 2015-04-18
Show process name in stats.

akash0x53:
PID is not so user friendly hence added process name in stats.


unbit:
uwsgi.procname could be NULL and covers only one of the ways to name processes.

Maybe it is better to manage it in platform dependent way (like reading /proc/self/cmd on linux and similar), or change uwsgi_set_processname to always set a pointer with the processname

PR #902, opened 2015-04-30
Covert most chmod users to fchmod

xrmx:
The plan is to avoid toctouc races, i've converted the user where the fd of the path was nearly available. There are a couple of hooks which plays with chmod which may be hard to convert and chmod on cgroup dir in core/utils.c:222 which can be coverted with some code reordering. Please review.

### DIFF
--- a/core/emperor.c
+++ b/core/emperor.c
@@ -222,7 +222,7 @@ static int on_demand_bind(char *socket_name) {
 	}
 
 	if (!is_tcp) {
-		if (chmod(socket_name, 0666)) {
+		if (fchmod(fd, 0666)) {
 			goto error;
 		}
 	}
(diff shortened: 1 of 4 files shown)


unbit:
I think it is "utopic" to aim at completely solving a problem that is intrinsic in UNIX :) Btw your patches are not invasive, and will not hurt for sure

unbit:
It looks like fchmod on socket is not portable (or better, it looks like is reliable only on Linux), @xrmx can you change it for taking in account only files ?

xrmx:
Commits on sockets are even more bogus because the fchmod must be called before bind otherwise it has no effect (reference: http://openvswitch.org/pipermail/discuss/2011-April/005073.html). So let's just revert the three socket commits. @unbit i can do and push the reverts if you are ok with that.

unbit:
yes its ok

xrmx:
Pushed the three reverts.

PR #905, opened 2015-05-05
Don't define UWSGI_CFLAGS for coverity build profile

xrmx:
coverity cli tools have supposedly a fixed buffer to contain each
compiler invocation. The embedding of UWSGI_CFLAGS of course breaks
that limit. The issue has been reported for more than a year and they
have not fixed that. So add this hack to make sure we don't define
UWSGI_CFLAGS if the build profile ends with 'coverity.ini'.


xrmx:
Please don't apply this, it's now working properly

PR #909, opened 2015-05-05
Fishy coverity fixes

xrmx:
These three are fixes for assigments to variables that are never read later. I think these require a bit of scrutiny since they can be symptoms of misplaced code. Or hopefully just leftovers.


xrmx:
@unbit ping

PR #910, opened 2015-05-05
Fix client certificate validation. Log verification errors

vladimir-didenko:
Hi, guys. 

It looks like client authentication via SSL certificate is completely broken now. 

You provide own verification callback (uwsgi_ssl_verify_callback) to OpenSSL via SSL_CTX_set_verify function. But in this callback uwsgi always returns 1 (success). 

From OpenSSL documentation:
>If verify_callback always returns 1, the TLS/SSL handshake will not be terminated with respect to verification failures and the connection will be established. The calling process can however retrieve the error code of the last verification error using SSL_get_verify_result or by maintaining its own error storage managed by verify_callback.

Since uwsgi doesn't use SSL_get_verify_result function then it accepts ANY client certificate. It can be easily reproducible - configure uwsgi to perform client authentication via SSL certificate and send https request using certificate not signed by trusted CA. 

Also I added some basic logging of validation errors. Tell me if I should remove it. Though from my point of view it is important information that should be logged.

### DIFF
--- a/core/ssl.c
+++ b/core/ssl.c
@@ -30,15 +30,15 @@ void uwsgi_ssl_info_cb(SSL const *ssl, int where, int ret) {
 }
 
 int uwsgi_ssl_verify_callback(int ok, X509_STORE_CTX * x509_store) {
-        char buf[256];
-        X509 *err_cert;
-        int depth;
-        int err;
-        depth = X509_STORE_CTX_get_error_depth(x509_store);
-        err_cert = X509_STORE_CTX_get_current_cert(x509_store);
-        X509_NAME_oneline(X509_get_subject_name(err_cert), buf, 256);
-        err = X509_STORE_CTX_get_error(x509_store);
         if (!ok) {
+                char buf[256];
+                X509 *err_cert;
+                int depth;
+                int err;
+                depth = X509_STORE_CTX_get_error_depth(x509_store);
+                err_cert = X509_STORE_CTX_get_current_cert(x509_store);
+                X509_NAME_oneline(X509_get_subject_name(err_cert), buf, 256);
+                err = X509_STORE_CTX_get_error(x509_store);
                 uwsgi_log("[uwsgi-ssl] client certificate verify error: num=%d:%s:depth=%d:%s\n", err, X509_verify_cert_error_string(err), depth, buf);
         }
         return ok;
--- a/core/ssl.c
+++ b/core/ssl.c
@@ -30,7 +30,18 @@ void uwsgi_ssl_info_cb(SSL const *ssl, int where, int ret) {
 }
 
 int uwsgi_ssl_verify_callback(int ok, X509_STORE_CTX * x509_store) {
-        return 1;
+        char buf[256];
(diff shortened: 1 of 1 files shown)


unbit:
Hi, you mean that specifying the ca file with the ! prefix only checks for the certificate presence but does not validate it against the supplied CA ?

Effectively from the docs it looks like is is somethign that must be managed by hand

vladimir-didenko:
@unbit 
>Hi, you mean that specifying the ca file with the ! prefix only checks for the certificate presence but does not validate it against the supplied CA ?

Yes. It is expected. In another case CA option doesn't have any sense. 

>Effectively from the docs it looks like is is somethign that must be managed by hand

I don't understand - how it can be managed by hand? It is task of uwsgi to authenticate client using SSL certificate.  I don't see any place in doc that claims another.

unbit:
by hand -> something that must be programmed like you did :)

vladimir-didenko:
@unbit 
>by hand -> something that must be programmed like you did :)

How I can handle this on application side if SSL connection is handled by uwsgi and client certificate is not provided to application.

unbit:
Hey, i am saying that you did right :)

vladimir-didenko:
>Hey, i am saying that you did right :)

Sorry, misunderstood :)

core/ssl.c
```diff
@@ -30,7 +30,18 @@ void uwsgi_ssl_info_cb(SSL const *ssl, int where, int ret) {
...
-        return 1;
+        char buf[256];
+        X509 *err_cert;
+        int depth;
+        int err;
+        depth = X509_STORE_CTX_get_error_depth(x509_store);
+        err_cert = X509_STORE_CTX_get_current_cert(x509_store);
+        X509_NAME_oneline(X509_get_subject_name(err_cert), buf, 256);
+        err = X509_STORE_CTX_get_error(x509_store);
+        if (!ok) {
```
unbit: can't we move it higher ? if i understand correctly those data are for logging

vladimir-didenko: Do you mean outside this function ? Full list of verification errors is accessible only from this callback. We can move 'uwsgi_log' outside but still will be forced to store error data in some context. Another question - where to extract error data back from this context. I see only one possibility in this case - store special flag that we have ssl enabled and extract errors after *->proto_accept call.

Also there is another possible solution - move logging under UWSGI_DEBUG macro.

unbit: if mean

if (!ok) {
509 *err_cert;
+        int depth;
+        int err;
+        depth = X509_STORE_CTX_get_error_depth(x509_store);
+        err_cert = X509_STORE_CTX_get_current_cert(x509_store);
+        X509_NAME_oneline(X509_get_subject_name(err_cert), buf, 256);
+        err = X509_STORE_CTX_get_error(x509_store);
uwsgi_log("....
}


regarding logging i would probably add a check for uwsgi.ssl_verbose like we do in https router

unbit:
Backported to 2.0 too, thanks a lot

PR #917, opened 2015-05-12
Tidy use of spaces and tabs

funkybob:


xrmx:
Unfortunately this has drawbacks like trashing the output of git blame which imho has more merits than properly formatted code.

unbit:
@xrmx so what would be the best approach to cleanup the code ? there are 6 years of tab/spaces mixing :)

xrmx:
@unbit I don't know :) One option is to clean it up before touching the code so you the last commit would be valuable. But that's a slow process with no guarantees.

@funkybob please agree with me that "More" is not a great commit message :)

funkybob:
I do agree it would be preferable if people would clean as they go, but evidence is they can't be relied on for that :)

I guess a better long term solution would be a hook to check the code commits...

methane:
@unbit Could you provide recommendation for new code?
I want vim modeline like `/* vim: set ts=4 sts=0 sw=0 expandtab : */`at last line.

unbit:
on top of uwsgi.h you will find the indent command line

PR #928, opened 2015-05-21
Added replica set support to emperor_mongodb and fixed uwsgi_instance attribute setup

WeirdCarrotMonster:
In emperor_mongodb plugin i added support for two arguments: replica and default. The `default` argument is used to set vassal attributes, which is not specified in document. It does not complete the required document fields (name, config and ts).

Also, i tried to make emperor_mongodb and docker plugin work together, and uncovered following problem:
If plugin tries to initialize on `vassal_before_exec` and configure something using vassal attributes, it will fail, because attribute setup is done *after* it's creation:
`uwsgi_emperor_simple_do_with_attrs` calls `uwsgi_emperor_simple_do`, which calls `emperor_add` and so on; then `ui->attrs` was set.
I rewrote `uwsgi_emperor_simple_do` and moved it's logic to `uwsgi_emperor_simple_do_with_attrs`.

I've realized that this pull request covers two different tasks too late, but i can try to split it if needed.

### DIFF
--- a/core/emperor.c
+++ b/core/emperor.c
@@ -2684,35 +2684,6 @@ next:
 }
 
 void uwsgi_emperor_simple_do_with_attrs(struct uwsgi_emperor_scanner *ues, char *name, char *config, time_t ts, uid_t uid, gid_t gid, char *socket_name, struct uwsgi_dyn_dict *attrs) {
-	uwsgi_emperor_simple_do(ues, name, config, ts, uid, gid, socket_name);
-	struct uwsgi_instance *ui_current = emperor_get(name);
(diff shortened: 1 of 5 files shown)


unbit:
All seems fine, i will make some test than i will merge it. Would you like to write a tutorial on uwsgi-docs repository about this particular setup ? (docker + emperor + mongodb)

WeirdCarrotMonster:
Yeah, i'd like to. I'm trying different configurations now, it works, but i think i need to add some options to docker plugin too.

unbit:
Ok, thank you a lot

PR #929, opened 2015-05-22
Wait for initial thread last.

kevinxucs:
Follow up on http://lists.unbit.it/pipermail/uwsgi/2015-May/008012.html

Since after calling pthread_cancel() and pthread_join() on initial thread, the worker process will appear to be a zombie, which prevents worker process to report its own memory stat, and if one of the spawned threads hangs, the worker process will hang forever since evil-reload-on-rss won't work when memory stat is missing.


unbit:
thanks

PR #932, opened 2015-06-01
Changed abs to labs because offset is declared as a long

protoCall7:
uwsgi with the psgi plugin fails to compile on OS X because offset is declared as a long, then passed to abs.

abs is defined as:

     int
     abs(int i);

where labs is defined as:

     long
     labs(long i);


unbit:
thanks

PR #936, opened 2015-06-07
some features

Darvame:
For Lua plugin
- rpc calls from my lua app ('local str = uwsgi.rpc(node, func_name, arg1, arg2, ...);')
- global "uwsgi.req_input_read" alias of local "env.input.read" function
- can yield numbers as well
- can return nil instead of header's table (same as {})
- can return nil or string (or number) instead of coroutine

### DIFF
--- a/plugins/lua/lua_plugin.c
+++ b/plugins/lua/lua_plugin.c
@@ -71,11 +71,46 @@ static int uwsgi_api_log(lua_State *L) {
 	lca(L, 1);
 
 	if (lua_isstring(L, 1)) {
-		logline = lua_tolstring(L, 1, NULL);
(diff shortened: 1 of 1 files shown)


unbit:
Hi, thanks for your great work. Before merging, do you have any specific use case for uwsgi.req_input_read ?

Darvame:
```lua
-- std way
local do_something_with_input = function(env)
    local input = env.input.read();
     ...
end
-- calling
local arg = do_something_with_input(env);
```
-- -- -- -- -- --
```lua
-- my way
local do_something_with_input = function()
    local input = uwsgi.req_input_read();
     ...
end
-- calling
local arg = do_something_with_input();
```

PR #940, opened 2015-06-11
When cheaper_overload=1, decheap worker for every 1 sec

methane:
I want to increase worker as soon as possible when spike comes.

I expect uWSGI spawns `cheaper-step` workers for each `cheaper-overload`.
But when I set `cheaper-overload` to 1, new workers are spawn for every 2 secs.


unbit:
Thanks

PR #942, opened 2015-06-14
add content_encoding to router_redis

nopjmp:
I have used this on my own personal server for a few days for gzip caching. This file has terrible white-spacing and might want to be fixed for easier committing.


unbit:
So you are inserting gzipped blobs in redis and then using uWSGI routing to serve them ?

nopjmp:
Yes, that is what I'm doing. It is https://uwsgi.readthedocs.org/en/latest/Transformations.html using redis instead of the memory caching system as I want the cache to survive uwsgi restart.

unbit:
Thanks

PR #943, opened 2015-06-15
cheaper-spare: More conservative to cheap

methane:
When there are many workers (e.g. 64 or more), 1 idle worker is not enough for "spare" worker.
Instead of one, use cheaper_count as ideal spare count.

### DIFF
--- a/core/master_utils.c
+++ b/core/master_utils.c
@@ -252,7 +252,6 @@ int uwsgi_cheaper_algo_manual(int can_spawn) {
 
 
 int uwsgi_cheaper_algo_spare(int can_spawn) {
-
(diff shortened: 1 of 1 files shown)


unbit:
I am bit scared of changing established behaviour. Would not be more "clean" to add a new algorithm ? (like spare2, or highloadspare, or .....). You can register new algos (without creating new plugins) in core/uwsgi.c (search for uwsgi_register_cheaper_algo() calls)

methane:
OK.  I want more configure options for large workload. (e.g. cheaper-busyness-multiplier like option).
I'll consider about name and design.

Issue #589, opened 2014-04-08
SNI and SPDY


prymitive:
Could uWSGI support SPDY when https, subscriptions and SNI are used?
This could be global option (`--sni-dir-spdy=true`) or per subscription (`--subscribe2=[...],spdy=1`)

This is assuming that SPDY is enabled per SSL context, rather than --https socket.

SamiLehtinen:
I would prefer HTTP/2 support instead of SPDY. https://tools.ietf.org/html/rfc7540 Yet it implements many things where tight interaction between the client, webserver and the application being server is required, like stream dependencies, exclusivity and priorities.

Issue #556, opened 2014-03-04
Warn if (Py3) WSGI app returns Unicode strings instead of bytes


PAStheLoD:
Though I'm largely ignorant of the problem space outside the most naive and simplest case, but I think it'd be great to warn the user if the WSGI application returns a Python3 string instead an iterable of bytes.

jbq:
I've been banging my head several hours on this.  String output is silently swallowed with Python 3 if I do this:

```
def app(environ, start_response):
    start_response('200 OK', [])
    return ["foo bar"]
```

Whereas it correctly processes the response if I use bytes:

```
def app(environ, start_response):
    start_response('200 OK', [])
    return [b"foo bar"]
```

Just for a missing `b` sign my program becomes a nightmare to troubleshoot!  Please issue a warning.

Issue #167, opened 2013-03-02
Don't count pings from nagios plugin in request counter


prymitive:
I'm using nagios plugin to check my apps and as I see those pings are going as normal requests but with modifier1=100, so they are not handled by the app itself, but rather by ping  plugin. This means that they are counted in worker requests counter, but for clarity I would have them ignored there. The reasoning for ignoring them in counters is that it is little misleading sometimes, I can't tell for sure if app had any requests until I check the logs (if I'm logging them), I can't just check the requests counter since it will always be non-zero with nagios checks running.
Request counter is incremented in uwsgi_close_request(), so we would need a piece of logic there, maybe we could add new field to uwsgi_plugin struct? So this:

```
struct uwsgi_plugin ping_plugin = {

        .name = "ping",
        .modifier1 = 100,
        .options = uwsgi_ping_options,
        .request = uwsgi_request_ping,
        .init = ping_init,
};
```

would become this:

```
struct uwsgi_plugin ping_plugin = {

        .name = "ping",
        .modifier1 = 100,
        .options = uwsgi_ping_options,
        .request = uwsgi_request_ping,
        .count_requests = 0,
        .init = ping_init,
};
```

and simple check in uwsgi_close_request():

```
if ( uwsgi.p[wsgi_req->uh->modifier1]->count_requests ) {
    uwsgi.workers[0].requests++;
    [...]
}
```

Is it acceptable? If so just say if `count_requests` is right name for it, I'll do the patch.
If not than I'm fine with that, it's not important, just nice to have.

PR #949, opened 2015-06-26
Fixes errors with Intel compiler for 64-bit

alef:
Fixes some Intel compiler errors for 64-bit on Linux 

`core/spooler.c(105): error #167: argument of type "void (*)()" is incompatible with parameter of type "void (*)(int)"`
`                uwsgi_unix_signal(SIGUSR1, spooler_wakeup);`

`core/fifo.c(97): error #556: a value of type "void (*)()" cannot be assigned to an entity of type "void (*)(int)"`
`        uwsgi_fifo_table['S'] = subscriptions_blocker;`

`plugins/cache/cache.c(208): error #589: transfer of control bypasses initialization of:`
`            variable "cache_dump" (declared at line 275)`

`[thread 5][icc] core/sigplugins/gevent/gevent.c(369): error #810: conversion from "void *" to "PyGILState_STATE={enum <unnamed>}" may lose significant bits`
`        PyGILState_Release((PyGILState_STATE) pthread_getspecific(up.upt_gil_key));`

### DIFF
--- a/core/config.c
+++ b/core/config.c
@@ -785,12 +785,14 @@ char *uwsgi_manage_placeholder(char *key) {
 
 		switch(state) {
 			case concat:
+                            {
 				if (current_value) arg1 = current_value;
 				if (value) arg2 = value;
 				char *ret = uwsgi_concat2(arg1, arg2);
 				if (current_value) free(current_value);
 				current_value = ret;	
 				break;
+                            }
 			case sum:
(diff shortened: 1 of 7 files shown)


xrmx:
I've cherry-picked the ones that makes sense to me in e910098e6b676b681df90bb1437376a6b3904a6e. Please open a new pull request with different commit per kind of fix if you care. Thanks!

PR #951, opened 2015-06-30
Check whether *value is indeed an SvRV before casting it

fabzzap:
I compiled uwsgi with psgi and ran it, and it segfaulted. Apparently, line 288 of plugins/psgi/psgi_loader.c casts *value to a SvRV and tries to get its type, even when *value is not an SvRV.


unbit:
thanks a lot

PR #972, opened 2015-07-18
 Fixed segfault on uems->default is NULL

swuecho:
the format of my change is not quite right, I try to get it right, but failed. 

when uems->defaults is NULL, mongo::fromjson will segfault.

BTW, what does the defaults in uems->defaults mean? thanks.

### DIFF
--- a/plugins/emperor_mongodb/emperor_mongodb.cc
+++ b/plugins/emperor_mongodb/emperor_mongodb.cc
@@ -40,7 +40,6 @@ extern "C" void uwsgi_imperial_monitor_mongodb(struct uwsgi_emperor_scanner *ues
 		}
 		mongo::BSONObj p = builder.obj();
 		mongo::BSONObj q = mongo::fromjson(uems->json);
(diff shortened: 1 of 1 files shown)


unbit:
Honestly i do not remember what defaults is, maybe @WeirdCarrotMonster could help ?

WeirdCarrotMonster:
`uems->defaults` provide default values for emperor attributes. I should have set them to empty dict by default and add few checks for value type.

swuecho:
Thanks. Could you take a look at my pull request or find a better fix？

xrmx:
@swuecho Does this path works for you? https://github.com/xrmx/uwsgi/commit/132a90c5f61e558e043c472670ce253c7c0d480c

swuecho:
@xrmx Should work. Thanks.

xrmx:
@swuecho Fixed in git master 71d657283e8e48dd06dc2cbae4eeb4228f5d3f7d

PR #977, opened 2015-07-22
lua_plugin: lua_newthread instead of new_State

Darvame:
I think it is working proper now.

only ONE lua_State per 1 worker, and one lua thread per one worker's core.
so, 4 workers with 8000 async cores will not create 8000 lua_States, only 4, for each worker.
uwsgi table now also has lua_thread collection, and mywid variable.

Know bugs: too many simultaneously running cores (like async = 8000) without proper gc may cause segmentation fault (out of memory)

May be I should add option like "lua_State per worker" in future.


unbit:
Nice work, thanks

PR #981, opened 2015-07-25
lua_plugin: --lazy-apps, --threads and other

Darvame:
our workers now init lua states on .post_fork:
- better rpc/signal registration logic
- faster loading (mutlicpu)
- avoiding some strange seg_fault event during heavy load (oops that was luajit-2.1alpha's problem)
- worker respawn also respawns lua_state

Also:
- our app can now work without wsapi app, it will just send 500 error code
- detecting dead coroutines in async mode without additional switch
- 'lua-gc-full = 1' option to perform a full gc instead of incremental

### DIFF
--- a/plugins/lua/lua_plugin.c
+++ b/plugins/lua/lua_plugin.c
@@ -1230,7 +1230,7 @@ static void uwsgi_lua_hijack(void) {
 static void uwsgi_lua_init_apps() {
 	int i;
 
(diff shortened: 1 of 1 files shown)


unbit:
Are you sure removing preforking support is a good idea ? --lazy-apps is the standard way to pospone initialization of vm states

Darvame:
hm

PR #982, opened 2015-07-26
plugins/greenlet: hack on uwsgiconfig.py to support venv

stepheny:
Either <greenlet.h> or <greenlet/greenlet.h> is available would be
tested in uwsgiconfig.py. Then tune plugins/greenlet/greenlet.c to
include the correct one by c macro.

With system repo provided python-greenlet-dev or equivalent greenlet.h 
is available at <greenlet/greenlet.h>. In contrast a virtual environment
pip installed greenlet would provide greenlet.h as <greenlet.h>. This is
the reason current uwsgi would always fail to build with 
`UWSGI_PROFILE='asyncio' pip install uwsgi` in a venv.

### DIFF
--- a/plugins/greenlet/greenlet.c
+++ b/plugins/greenlet/greenlet.c
@@ -1,5 +1,9 @@
 #include "../python/uwsgi_python.h"
+#ifndef GREENLET_HEADER_HACK
 #include <greenlet/greenlet.h>
+#else
+#include <greenlet.h>
+#endif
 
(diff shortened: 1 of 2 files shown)


circlingthesun:
Just what I was looking for! :+1:

2mf:
I think I got that covered with #1467

2mf:
@stepheny  @xrmx I think you can close this one

PR #983, opened 2015-07-26
Allow call uwsgi websocket API in other greenlet (NOT the main callable)

yueyoum:
I add this feature for using uwsgi websocket with gevent more easy and convenient.

SEE  tests/websockets_chat_2.py 

Add a new API uwsgi.input_object() for get `uwsgi_Input` object in Python.
Modify uwsgi.websocket_recv/recv_nb/send/send_binary, add a kwarg: `uwsgi_input`.

If not provide the `uwsgi_input` kwarg, This function's behavior not changed.
When provide the `uwsgi_input` kwarg, This function will find `wsgi_req` from the `uwsgi_input`.

In this way, uwsgi websocket functions can be used in other greenlets.

NOTE:  uwsgi.websocket_handshake, uwsgi.input_object can only call from the main callback.

### DIFF
--- a/plugins/python/uwsgi_pymodule.c
+++ b/plugins/python/uwsgi_pymodule.c
@@ -4,6 +4,42 @@ extern struct uwsgi_server uwsgi;
 extern struct uwsgi_python up;
 extern struct uwsgi_plugin python_plugin;
 
+PyTypeObject uwsgi_UWSGIInput = {
+    PyObject_HEAD_INIT(NULL)
+    0,
+    "uwsgi.UWSGIInput",
+    sizeof(uwsgi_Input),
+    0,
+    0,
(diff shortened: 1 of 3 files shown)


unbit:
In the current form it introduces various security problems and unexpected behaviours as the wsgi_req pointer could map to no more running requests or (and this is really bad) a completely different running requests as that memory is constantly reused (it is indeed shared memory).

I think you could simply add an identification during object initialization with the core.requests value of wsgi_req and check it whenever the object is used (the thing would be race free as threads are disabled in gevent mode). Checking the in_request field too will avoid managing already closed requests.

yueyoum:
Thank you for your advice, I will keep go on work on this.

Issue #980, opened 2015-07-24
uwsgi websocket, gevent error:  [BUG] current_wsgi_req NOT FOUND !!!


yueyoum:
python code:

```python
                                                                                 
def send_job():                                                                  
    gevent.sleep(2)                                                              
    print "start send job"                                                       
    uwsgi.websocket_send("haha")                                                 
                                                                                 
                                                                                 
def application(env, start_response):                                            
    print "before handshake"                                                     
    uwsgi.websocket_handshake()                                                  
    print "after handshake"                                                      
                                                                                 
    uwsgi.websocket_send("<Hi, websocket>")                                      
                                                                                 
    gevent.spawn(send_job)                                                       
                                                                                 
    while True:                                                                  
        x = uwsgi.websocket_recv()                                               
        print "got:", x                                                          
                                                                                 
    return []                                                                    
                
```


uwsgi.ini

```
[uwsgi]
chdir = %d
master = true
processes = 1
enable-threads = true
threads = 1
module = app:application

http-socket = :7000
http-raw-body = true

gevent = 10

```

when a client connected, the error occurred!

log

```
[uWSGI] getting INI configuration from uwsgi_ws.ini
*** Starting uWSGI 2.0.11 (32bit) on [Fri Jul 24 19:37:12 2015] ***
compiled with version: 4.9.2 on 02 July 2015 17:15:33
os: Linux-3.13.0-57-generic #95-Ubuntu SMP Fri Jun 19 09:27:48 UTC 2015
nodename: wang-X230
machine: i686
clock source: unix
pcre jit disabled
detected number of CPU cores: 4
current working directory: /home/wang/learn/uwsgi/raw-socket
detected binary path: /usr/local/bin/uwsgi
chdir() to /home/wang/learn/uwsgi/raw-socket/
your processes number limit is 94605
your memory page size is 4096 bytes
detected max file descriptor number: 100000
- async cores set to 10 - fd table size: 100000
lock engine: pthread robust mutexes
thunder lock: disabled (you can enable it with --thunder-lock)
uwsgi socket 0 bound to TCP address :7000 fd 3
Python version: 2.7.6 (default, Jun 22 2015, 18:03:54)  [GCC 4.8.2]
Python main interpreter initialized at 0x84d6d90
python threads support enabled
your server socket listen backlog is limited to 100 connections
your mercy for graceful operations on workers is 60 seconds
mapped 142800 bytes (139 KB) for 2 cores
*** Operational MODE: threaded ***
[...]

PR #985, opened 2015-07-27
Allow call uwsgi websocket API in other greenlet

yueyoum:
see issure: #980   and pull request: #983 

Now , There are some security check for verify the `wsgi_req` passed from Python code.

check `mywid` and `core.requests` for ensure this is the current `wsgi_req` of the main callable.
check `wsgi_req->fd` for ensure this is a valid connection. (not closed)
check `core.in_request` for ensure this is a valid connection.  (still in request, not closed).

### DIFF
--- a/plugins/python/uwsgi_pymodule.c
+++ b/plugins/python/uwsgi_pymodule.c
@@ -4,6 +4,52 @@ extern struct uwsgi_server uwsgi;
 extern struct uwsgi_python up;
 extern struct uwsgi_plugin python_plugin;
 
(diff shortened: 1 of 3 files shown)


yueyoum:
sorry,  there need more error check for parse the `uwsgi_input` keyword argument.

I will send an other pull request.

unbit:
Ok, but i would change the 'input' name to something like 'request_context'. 'input' in the WSGI/PSGI/Rack gergo refers to the body of the request

PR #987, opened 2015-07-28
lua: Fix malloc allocation size mismatch

xrmx:
argv is of type char** so it should its size should be the size of
n char pointers. argvs instead is of type uint16_t * so its size
should be of an array of uint16_t.

This has been only compile tested.

### DIFF
--- a/plugins/lua/lua_plugin.c
+++ b/plugins/lua/lua_plugin.c
@@ -718,7 +718,7 @@ static int uwsgi_lua_init(){
 	int i;
 	
 	for(i=0;i<uwsgi.numproc;i++) {
-		ulua.state[i] = (lua_State**) uwsgi_malloc( sizeof(lua_State**) * uwsgi.cores );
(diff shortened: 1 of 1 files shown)


Darvame:
really? how did I miss that

xrmx:
@Darvame any chance you can give it a try please? sizeof(**foo) should be the same of sizeof(*foo) so in practice does not matter much. Regarding argvs the size of the array now is bigger.

Darvame:
@xrmx I know, pointer size is 8 or 4, char size is 1
It was accidental mistake, but I did not notice it all time

xrmx:
@Darvame thanks for reviewing! :)

Darvame:
@xrmx thanks to you and your compiler, now looking for more pro memory allocation

xrmx:
Closing this, same changes has been committed here 1e97a5e8d0515312906635cb11ff8b41cc8887ad

PR #1009, opened 2015-08-18
Patch spooler signal as task

yueyoum:
## Background

We heavily use uWSGI  in our products.
There are many jobs managed by uWSGI cron system. 
this jobs are run in spooler.
so the workers can handle requests from our clients efficiently.

## Problem

Some job will loads a lots of data in memory,  In this case, 
the spooler processes will use lots of memory.

Like this:

memory-day1

We want to spooler auto restart like workers, (limit-as ...)
We find an option  `spooler-max-tasks`   in the [doc](http://uwsgi-docs.readthedocs.org/en/latest/Spooler.html)

But, unfortunately, it only works for spool tasks. 

## Solution
This patch introduce a new option `spooler-signal-as-task`.
when set this option, uwsgi will treat signal events (cron, timer) as tasks.
when the `signal events times + tasks times >= spooler-max-tasks`, 
the spool process will restart, to alleviate memory leaks.

#### app.py

```python
# -*- coding: utf-8 -*-

import os
import time
import arrow
from bottle import Bottle

import uwsgidecorators

now = lambda: arrow.now().format("YYYY-MM-DD HH:mm:ss")

@uwsgidecorators.cron(-1, -1, -1, -1, -1, target="spooler")
def mycron(signum):
    print "[PYTHON CRON] {0}. PID {1}".format(now(), os.getpid())
    time.sleep(5)
    print "[PYTHON CRON] done"

@uwsgidecorators.timer(20, target="spooler")
def mytimer(signum):
    print "[PYTHON TIMER] {0}, PID {1}".format(now(), os.getpid())
    time.sleep(5)
    print "[PYTHON TIMER] done"


application = Bottle()

@application.get("/")
def index():
    return "OK"
```
#### uwsgi.ini

```
[uwsgi]
module = app:application
master = true
processes = 1

http = 0.0.0.0:8000
spooler = .
spooler-processes = 4
spooler-max-tasks = 2
spooler-signal-as-task = true
```

#### log

```
*** Operational MODE: single process ***
[uwsgi-signal] signum 0 registered (wid: 0 modifier1: 0 target: spooler)
[uwsgi-signal] signum 1 registered (wid: 0 modifier1: 0 target: spooler)
WSGI app 0 (mountpoint='') ready in 0 seconds on interpreter 0x9d00948 pid: 9746 (default app)
spawned uWSGI master process (pid: 9746)
spawned the uWSGI spooler on dir /home/wang/learn/uwsgi/spool with pid 9747
spawned the uWSGI spooler on dir /home/wang/learn/uwsgi/spool with pid 9748
spawned the uWSGI spooler on dir /home/wang/learn/uwsgi/spool with pid 9749
spawned the uWSGI spooler on dir /home/wang/learn/uwsgi/spool with pid 9750
spawned uWSGI worker 1 (pid: 9751, cores: 1)
spawned uWSGI http 1 (pid: 9752)
[PYTHON CRON] 2015-08-19 00:59:48. PID 9748
[PYTHON CRON] done
[PYTHON TIMER] 2015-08-19 01:00:07, PID 9748
[PYTHON TIMER] done
[spooler /home/wang/learn/uwsgi/spool pid: 9748] maximum number of tasks reached (2) recycling ...
OOOPS the spooler is no more...trying respawn...
spawned the uWSGI spooler on dir /home/wang/learn/uwsgi/spool with pid 9753
[PYTHON TIMER] 2015-08-19 01:00:27, PID 9750
[PYTHON TIMER] done
[PYTHON TIMER] 2015-08-19 01:00:47, PID 9753
[PYTHON CRON] 2015-08-19 01:00:48. PID 9749
[PYTHON TIMER] done
[PYTHON CRON] done
[PYTHON TIMER] 2015-08-19 01:01:07, PID 9750
[PYTHON TIMER] done
[spooler /home/wang/learn/uwsgi/spool pid: 9750] maximum number of tasks reached (2) recycling ...
OOOPS the spooler is no more...trying respawn...
spawned the uWSGI spooler on dir /home/wang/learn/uwsgi/spool with pid 9760
[PYTHON TIMER] 2015-08-19 01:01:27, PID 9753
[PYTHON TIMER] done
[spooler /home/wang/learn/uwsgi/spool pid: 9753] maximum number of tasks reached (2) recycling ...
OOOPS the spooler is no more...trying respawn...
spawned the uWSGI spooler on dir /home/wang/learn/uwsgi/spool with pid 9763
```

### DIFF
--- a/core/signal.c
+++ b/core/signal.c
@@ -467,8 +467,7 @@ cycle:
 	return received_signal;
 }
 
-void uwsgi_receive_signal(struct wsgi_request *wsgi_req, int fd, char *name, int id) {
-
+pid_t uwsgi_receive_signal(struct wsgi_request *wsgi_req, int fd, char *name, int id) {
 	uint8_t uwsgi_signal;
 
 	ssize_t ret = read(fd, &uwsgi_signal, 1);
@@ -487,13 +486,15 @@ void uwsgi_receive_signal(struct wsgi_request *wsgi_req, int fd, char *name, int
 		if (uwsgi_signal_handler(wsgi_req, uwsgi_signal)) {
 			uwsgi_log_verbose("error managing signal %d on %s %d\n", uwsgi_signal, name, id);
 		}
+		return getpid();
 	}
 
-	return;
+	return -1;
 
 destroy:
 	// better to kill the whole worker...
 	uwsgi_log_verbose("uWSGI %s %d screams: UAAAAAAH my master disconnected: I will kill myself!!!\n", name, id);
 	end_me(0);
-
+	// never here
+	return -1;
 }
--- a/core/signal.c
+++ b/core/signal.c
@@ -467,7 +467,7 @@ cycle:
 	return received_signal;
 }
 
-pid_t uwsgi_receive_signal(struct wsgi_request *wsgi_req, int fd, char *name, int id) {
+int uwsgi_receive_signal(struct wsgi_request *wsgi_req, int fd, char *name, int id) {
 	uint8_t uwsgi_signal;
 
 	ssize_t ret = read(fd, &uwsgi_signal, 1);
@@ -486,15 +486,15 @@ pid_t uwsgi_receive_signal(struct wsgi_request *wsgi_req, int fd, char *name, in
 		if (uwsgi_signal_handler(wsgi_req, uwsgi_signal)) {
 			uwsgi_log_verbose("error managing signal %d on %s %d\n", uwsgi_signal, name, id);
 		}
-		return getpid();
+		return 1;
 	}
 
-	return -1;
+	return 0;
 
 destroy:
 	// better to kill the whole worker...
 	uwsgi_log_verbose("uWSGI %s %d screams: UAAAAAAH my master disconnected: I will kill myself!!!\n", name, id);
 	end_me(0);
 	// never here
-	return -1;
+	return 0;
 }
--- a/uwsgi.h
+++ b/uwsgi.h
@@ -2374,6 +2374,7 @@ struct uwsgi_server {
 	int spooler_ordered;
 	int spooler_quiet;
 	int spooler_frequency;
+	int spooler_signal_as_task;
 
 	int snmp;
 	char *snmp_addr;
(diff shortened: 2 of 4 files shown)


core/signal.c
```diff
@@ -467,8 +467,7 @@ int uwsgi_signal_wait(struct wsgi_request *wsgi_req, int signum) {
 	return received_signal;
 }
 
-void uwsgi_receive_signal(struct wsgi_request *wsgi_req, int fd, char *name, int id) {
-
+pid_t uwsgi_receive_signal(struct wsgi_request *wsgi_req, int fd, char *name, int id) {
```
unbit: is this required ? uwsgi_receive_signal is always managed by the process calling it

uwsgi.h
```diff
@@ -2374,6 +2374,7 @@ struct uwsgi_server {
 	int spooler_ordered;
 	int spooler_quiet;
 	int spooler_frequency;
+	int spooler_signal_as_task;
```
unbit: place this at the end of the structure to retain ABI compatibility

core/signal.c
yueyoum: the spooler call `uwsgi_receive_signal` is at `file: core/spooler.c, line: 468`.

when a signal is routed to spooler,  maybe all of the spooler process will be wakeup,
and call the `uwsgi_receive_signal`.
But only one spooler process will success at `ssize_t ret = read(fd, &uwsgi_signal, 1);`
in `file: core/signal.c, line: 474`

So, `uwsgi_receive_signal` should return a sign to indicate whether this spooler process will run the signal handler.


I think this can be simply return 0 for success read ( will run signal handler ), -1 for nothing.

uwsgi.h
yueyoum: OK, i got it

yueyoum:
I make an update, 
see the new codes

yueyoum:
Hi, i made a new update.

There are two big changes:

#### 1, uwsgi_receive_signal 

change `void uwsgi_receive_signal (...)` to `int uwsgi_receive_signal (...)`

When a signal was routed to spooler, maybe all spooler processes will be wake up,
then, they try to read signal from pipe. But only one process will be success read.
and this process will run signal handler.

So, we can check the return value of `uwsgi_receive_signal` to determining whether this process was run a signal event.

If run, we can increase the `uspool->tasks` of this spooler process

#### 2. struct uwsgi_server

add a new field `int spooler_signal_as_task;` at the end of this struct.
and add an option `spooler-signal-as-task` to set this value.

if not set, the spooler's behavior are not changed. 
when set, running signal handler will also increase the `uspool->tasks`.


##### In this way, spooler processes can re-spawn self for signal events.  (not only for tasks)
this can help alleviate memory leaks caused by cron, timers ...

unbit:
Thanks a lot

PR #1016, opened 2015-08-25
Add no-workers option to statsd plugin

mwicat:
This is in some way a reincarnation of #722. Here is our scenario:

We want to gather uwsgi metrics with the following flow:

uwsgi -> statsd -> graphite

Our deployment does not allow us to send metrics generated by all uwsgi workers because in worst case that would result in 2 Data centers x 6 hosts x 40 workers (cores) x 8 metrics (worker.X...) = 3840 extra metrics. This is clearly an overkill, yet I would rather say that it is popular deployment scenario.

Carbon plugin has a nice option carbon-no-workers, yet we cannot push directly to remote graphite server with this plugin because it would hit our performance. In #722 @unbit proposed to implement regular expression filtering, yet I cannot see anyone that would want to do this in C in near future, also is it really needed? I think that merging the basic case for people who just don't have resources for pushing a few thousand metrics every time would be really helpful!

### DIFF
--- a/plugins/stats_pusher_statsd/plugin.c
+++ b/plugins/stats_pusher_statsd/plugin.c
@@ -16,6 +16,14 @@ it exports values exposed by the metric subsystem
 
 extern struct uwsgi_server uwsgi;
 
+struct uwsgi_stats_pusher_statsd {
+    int no_workers;
+} u_stats_pusher_statsd;
+
+static struct uwsgi_option stats_pusher_statsd_options[] = {
+	{"statsd-no-workers", no_argument, 0, "disable generation of single worker metrics", uwsgi_opt_true, &u_stats_pusher_statsd.no_workers, 0}
+};
+
 // configuration of a statsd node
 struct statsd_node {
 	int fd;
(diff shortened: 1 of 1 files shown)


r4fek:
+1

plugins/stats_pusher_statsd/plugin.c
```diff
@@ -88,7 +96,10 @@ static void stats_pusher_statsd(struct uwsgi_stats_pusher_instance *uspi, time_t
 	while(um) {
 		uwsgi_rlock(uwsgi.metrics_lock);
 		// ignore return value
-		if (um->type == UWSGI_METRIC_GAUGE) {
+		if (u_stats_pusher_statsd.no_workers && strstr(um->name, "worker.") != NULL) {
```
unbit: place this check before the lock, and use a "goto next" to the end of the while block (where um is redefined).

Instead of strstr use uwsgi_starts_with(src, src_len, dst, dst_len) that is faster and safer

mwicat:
@unbit: fixed

unbit:
Merged, thanks

PR #1029, opened 2015-09-02
Skip logrotate checking if we are logging to a pipe

kevinxucs:
If we are logging to a pipe file, lseek() will fail and show up as an error in the log. Simply skip the check if the file is a pipe.


xrmx:
Merged, thanks.

Issue #1039, opened 2015-09-15
out of bounds writes in lua plugin


xrmx:
Hi @Darvame, coverity reports a couple of new warnings in lua plugin. The problem here is an off by one using argc loop guard. Please review this inlined patch.
```
diff --git a/plugins/lua/lua_plugin.c b/plugins/lua/lua_plugin.c
index d52d395..a472ba3 100644
--- a/plugins/lua/lua_plugin.c
+++ b/plugins/lua/lua_plugin.c
@@ -1693,7 +1693,7 @@ static int uwsgi_api_queue_push(lua_State *L) {
                list = (char **) uwsgi_malloc(sizeof(char *) * argc);
                slen = (size_t *) uwsgi_malloc(sizeof(size_t) * argc);
                
-               for (i = 1; i <= argc; ++i) {
+               for (i = 1; i < argc; ++i) {
                        
                        if (lua_istable(L, i)) {
                                uwsgi_lua_metatable_tostring(L, i - argc - 1);
@@ -1912,7 +1912,7 @@ static int uwsgi_api_queue_get(lua_State *L) {
                
                list = (uint64_t *) uwsgi_malloc(sizeof(uint64_t) * argc);
                
-               for (i = 1; i <= argc; ++i) {
+               for (i = 1; i < argc; ++i) {
                        list[i] = lua_tonumber(L, i);
                }
                
@@ -1920,7 +1920,7 @@ static int uwsgi_api_queue_get(lua_State *L) {
                
                uwsgi_rlock(uwsgi.queue_lock);
                
-               for (i = 1; i <= argc; ++i) {
+               for (i = 1; i < argc; ++i) {
                        str = uwsgi_queue_get(list[i], &len);
                        
                        if (len) {
```

Darvame:
op, thx, didnt notice, like always do
50bf007069495db90a2b91b0ebbf94f0b3a114ce

xrmx:
This is being handled in #1040, closing.

PR #1040, opened 2015-09-15
lua_plugin: queue segfault

Darvame:
possible segfault with multiple arguments
lua stack iterator is not suitable for c arrays :<


xrmx:
@Darvame could you please add a "Fix #1039" in your commit message? thanks

Darvame:
Nope, sorry, not right now and not from here

PR #1042, opened 2015-09-16
Changing SOL_TCP to IPPROTO_TCP

shawnzvinis:
SOL_TCP not declared in OS X 10.11


xrmx:
@shawnzvinis no need to close pull request, just push -f and it'll get updated automagically. BTW working on branches will make respin stuff less painful. Thanks!

PR #1043, opened 2015-09-20
also remove all master fifo if vacuum is true

multiSnow:
since vacuum option 'try to remove all of the generated file/sockets', the fifo(s) generated by uwsgi should also  be removed.


unbit:
Thanks

PR #1045, opened 2015-09-23
lua_plugin: protected and unprotected __tostring

Darvame:
- protected and unprotected __tostring (protected outside, unprotected internal, avoid double pcall)
- also check userdata for metafields (non simple type)
- mywid and mysid are functions now
- minor api bump


Darvame:
may be merged, if there are no any bugs

Issue #973, opened 2015-07-18
mod_proxy_uwsgi: uds not working, dns lookup failure


BinaryKhaos:
After trying really hard to get mod_proxy_uwsgi working with uds, I am convinced I have hit a bug either in Apache or uwsgi.

The error message is always a variation of: "AH00898: DNS lookup failure for: \xe8:&{\xcd\x0c returned by /" where the actual unicode string varies depending on the configuration.

What is interesting, I can give whatever path to a uds I want (either existent or not) and it does not make the slightest difference when it comes to error messages. The same goes for permissions.

I have tested this w/ uwsgi 2.0.11 and Apache 2.4.12 as well as 2.4.16 -- both make absolutely no difference at all.

As wsgi application I am taking a vanilla Django 1.8.3 installation w/ Python 3.4.3 (same version that uwsgi is running on). Here the (test and not production) configuration for the app:

```
[uwsgi]
uid=someuser
gid=someuser
virtualenv=/home/someuser/.virtualenvs/testenv/
chdir=/home/someuser/public/testsite.test/
module=helloworld.wsgi:application
socket=/tmp/uwsgi-testsite.test.sock
master=True
need-plugin=python34
```

uwsgi is configured in emperor mode -- which works fine. The vassal however never becomes loyal with the uds configuration (and no access is logged on uwsgi). If, on the other hand, a tcp socket is used (socket=127.0.0.1:9999), it works just fine.

Here the relevant Apache configuration for the vhost:

```
<VirtualHost 10.0.0.1:80>
  ServerName testsite.test

  <Location "/">
    ProxyPass unix:/tmp/uwsgi-testsite.test.sock|uwsgi://
#    ProxyPass uwsgi://127.0.0.1:9999/
  </Location>
</VirtualHost>
```

According to the docs, postings and some source code I read, the syntax is just fine. And like I said, what is strange is that I can pass whatever path I like, it never raises a different error message at all. Simply the unicode string varies, nothing more.

If there is anything else I can help with to debug this further, please let me know. I would have gladly jumped right in and tried to hunt this pesky bug down myself but right now I am rather limited in time and unfortunately I have a feeling this could take quite some time for someone not familiar with either the Apache or the uwsgi code base.

xrmx:
@BinaryKhaos do the changes in @niol pull request fixes the issue for you?

xrmx:
This should be fixed in master, please reopen if it's not the case

Issue #912, opened 2015-05-06
mod_proxy_uwsgi


waszi:
I have configured mod_proxy_uwsgi (2.0.6 ubuntu 14.10) and I have problem with 404 errors:

```
<Proxy "unix:/run/uwsgi/app/test.sock|uwsgi://uwsgi-test">
            ProxySet disablereuse=off
</Proxy>
<FilesMatch \.php$>
            SetHandler proxy:uwsgi://uwsgi-test
</FilesMatch>
```
similar configuration with nginx is working fine. I have checked that SCRIPT_NAME/PATH_INFO is different:

apache:
```
{"method": "GET", "uri": "/status.php", "proto": "HTTP/1.1", "status": 404, "referer": "-", "user_agent": "Mozilla/5.0 (Windows) mirall/1.7.1", "remote_addr": "46.250.176.80", "http_host": "178.217.186.153:8080", "pid": 16, "worker_id": 4, "core": 0, "async_switches": 0, "io_errors": 0, "rq_size": 0, "rs_time_ms": 0, "rs_size": 80, "rs_header_size": 71, "rs_header_count": 2, "modifier1": "0", "PATH_INFO": "/www/test/status.php", "SCRIPT_NAME": "/status.php", "SCRIPT_FILENAME": "proxy:uwsgi://uwsgi-test/www/test/status.php" }
```
nginx:
```
{"method": "GET", "uri": "/info.php", "proto": "HTTP/1.1", "status": 200, "referer": "-", "user_agent": "curl/7.37.1", "remote_addr": "91.207.168.255", "http_host": "test.safeonline.pl", "pid": 16, "worker_id": 4, "core": 0, "async_switches": 0, "io_errors": 0, "rq_size": 0, "rs_time_ms": 1, "rs_size": 61839, "rs_header_size": 81, "rs_header_count": 2, "modifier1": "0", "PATH_INFO": "/info.php", "SCRIPT_NAME": "-", "SCRIPT_FILENAME": "-" }
```
my uwsgi config:

```
name = test
uid = 10000
gid = 10000
auto-procname = true
procname-prefix = %(name):
project_dir = /www/%(name)

plugins = 0:php
socket = /run/uwsgi/app/%(name).sock

php-sapi-name = apache
chdir = %(project_dir)
check-static = %(project_dir)
static-skip-ext = .php
static-skip-ext = .inc
static-index = index.html

php-docroot = %(project_dir)
php-allowed-ext = .php
[...]

Issue #890, opened 2015-04-16
Apache2 proxy forwarding to unix sockets with mod_proxy_uwsgi


mostarman:
Hi - I am having several issues trying to set up an Apache2 web server (version 2.4.12) as a proxy forwarding requests to a unix socket using the mod_proxy_uwsgi module.
I am loading the module like this: ProxyPass / unix:/tmp/uwsgi.sock|uwsgi://
So first there are a few iffy lines in the uwsgi_canon function (like char *host and sport never being initialized), but even with that fixed I can never get it to recognize unix: as a web socket - it always tries to connect to host:sport. Am I missing something?

unbit:
The syntax for unix sockets changed during 2.4 development cycle, check if the mod_proxy_fcgi approach works in latest versions: http://httpd.apache.org/docs/2.4/mod/mod_proxy_fcgi.html

niol:
Same here, I am using ```ProxyPass "/r/" "unix:/var/run/uwsgi/foo.socket|uwsgi://localhost/"``` and the patch below which avoids using the unitialized ```char* host``` but it still does not want to connect via unix domain socket (UDS): ```ap_proxy_determine_connection``` does not set ```uds_path``` for some reason. I'll follow up if  I find something.

```diff
--- uwsgi-2.0.7.orig/apache2/mod_proxy_uwsgi.c
+++ uwsgi-2.0.7/apache2/mod_proxy_uwsgi.c
@@ -67,19 +67,14 @@ static int uwsgi_canon(request_rec *r, c
     }
     url += sizeof(UWSGI_SCHEME); /* Keep slashes */

-    // is it a unix socket ?
-    if (strlen(url) == 2) {
-       *sport = 0;
-    }
-    else {
-        err = ap_proxy_canon_netloc(r->pool, &url, NULL, NULL, &host, &port);
-        if (err) {
-            ap_log_rerror(APLOG_MARK, APLOG_ERR, 0, r,
-                      "error parsing URL %s: %s", url, err);
-            return HTTP_BAD_REQUEST;
-        }
-       apr_snprintf(sport, sizeof(sport), ":%u", port);
+    err = ap_proxy_canon_netloc(r->pool, &url, NULL, NULL, &host, &port);
+    if (err) {
+        ap_log_rerror(APLOG_MARK, APLOG_ERR, 0, r,
+                    "error parsing URL %s: %s", url, err);
+        return HTTP_BAD_REQUEST;
     }
[...]

PR #1054, opened 2015-10-02
mod_proxy_uwsgi: fix apache 2.4 integration with unix domain sockets

niol:
There is no way I could make this to work using the existing code and ```uwsgi://``` syntax as said in the docs.

The proposed patch makes this work.

Example usage in apache conf:

    ProxyPass "/foo"  "unix:/var/run/uwsgi/foo.socket|uwsgi://uwsgi-uds-foo/"

Note: I use uwsgi-uds-foo as hostname instead of localhost because if
you have multiple entrie like these in the same apache conf, apache
thinks it is the same backend even though the unix socket path is
different. So it is mandatory to use different hostnames if you have
multiple app sockets.

Some references:
- https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=752783
- https://github.com/unbit/uwsgi/issues/973
- maybe https://github.com/unbit/uwsgi/issues/912
- https://github.com/unbit/uwsgi/issues/890

### DIFF
--- a/apache2/mod_proxy_uwsgi.c
+++ b/apache2/mod_proxy_uwsgi.c
@@ -67,19 +67,17 @@ static int uwsgi_canon(request_rec *r, char *url)
     }
     url += sizeof(UWSGI_SCHEME); /* Keep slashes */
 
-    // is it a unix socket ?
-    if (strlen(url) == 2) {
-	*sport = 0;
+    err = ap_proxy_canon_netloc(r->pool, &url, NULL, NULL, &host, &port);
+    if (err) {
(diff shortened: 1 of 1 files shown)


kontza:
Hi,

It works! It works! :)

I can confirm that this fixed `mod_proxy_uwsgi.c` works as designed. At least on two of my machines:
* Ubuntu 15.04 w/Apache 2.4.10-9ubuntu1.1
* OS X El Capitan w/Apache 2.4.16

xrmx:
Let's merge this. Thanks a lot @nios for the patch and @kontza for testing.

PR #1058, opened 2015-10-03
Add timer interface with millisecond resolution to python plugin

xrmx:
We extend uwsgi timer api to handle a nanosecond value in order to support timer with finer grained resolution than the current second. Then we add interface to use timers with millisecond resolution is added to the python plugin. I've extended the decorators with another decorator mstimer. Alternatively we can add a resolution parameter to the current timer decorator. It's fine for me either way.
Eventually external plugins that used that interface of course need to be updated.
Please note this has NOT ever been compiled on kqueue systems.

### DIFF
--- a/core/event.c
+++ b/core/event.c
@@ -1252,7 +1252,7 @@ static int timerfd_settime(int __ufd, int __flags, __const struct itimerspec *__
 }
 #endif
 
-int event_queue_add_timer(int eq, int *id, int sec) {
+int event_queue_add_timer_hr(int eq, int *id, int sec, long nsec) {
 
 	struct itimerspec it;
 	int tfd = timerfd_create(CLOCK_REALTIME, TFD_CLOEXEC);
@@ -1263,10 +1263,10 @@ int event_queue_add_timer(int eq, int *id, int sec) {
 	}
 
 	it.it_value.tv_sec = sec;
-	it.it_value.tv_nsec = 0;
+	it.it_value.tv_nsec = nsec;
 
 	it.it_interval.tv_sec = sec;
-	it.it_interval.tv_nsec = 0;
+	it.it_interval.tv_nsec = nsec;
 
 	if (timerfd_settime(tfd, 0, &it, NULL)) {
 		uwsgi_error("timerfd_settime()");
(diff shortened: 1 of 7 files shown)


core/event.c
```diff
@@ -1382,7 +1382,7 @@ int event_queue_add_timer(int eq, int *id, int sec) {
 	*id = timer_id;
 	timer_id++;
 
-	EV_SET(&kev, *id, EVFILT_TIMER, EV_ADD, 0, sec * 1000, 0);
+	EV_SET(&kev, *id, EVFILT_TIMER, EV_ADD, 0, sec * 1000 + int(nsec / 1000000), 0);
```
akx: The unit of the 6th parameter isn't immediately clear... Could it somehow be made clearer?

xrmx: Sure, we may use and intermediate variable to make it explicit.

Darvame:
Hi, lua plugin also uses <b>uwsgi_add_timer()</b>
and, in struct uwsgi_timer, order like:
```c
int registered;
uint8_t sig;
\\7 bytes gap?
long nsvalue;
```
is not cool for structure packing (struct memory usage)
but this is not so important

xrmx:
@Darvame i know but we tend to not breaking ABI and add new struct members at the end. Now i don't know if that struct is an internal one or not. We should probably start thinking to separate the public stuff for the internal one in uwsgi.h.

xrmx:
from unbit:
```
i would prefer to not break the old api, what about defining event_queue_add_timer_hr and leaving event_queue_add_timer as is ?
```

xrmx:
Updated pull requests, anyone that can give it a try on a freebsd / mac os x machine appreciated :) cc @unbit

Darvame:
um, 
```bash
core/event.c:1385:39: error: expected expression
        const long timeout_ms = sec * 1000 + int(nsec / 1000000);
```
on my mac with clang

xrmx:
@Darvame thanks for checking. Writing a lot more pytohn than C these days. Pushed a new version and another patch.

Darvame:
```bash
core/event.c:1384:10: error: read-only variable is not assignable
        timer_id++;
```
timer_id must be converted from const before ++?

xrmx:
@Darvame thanks, the last patch was completely non sense, removed.

Darvame:
clang version
```bash
Apple LLVM version 7.0.0 (clang-700.1.76)
Target: x86_64-apple-darwin15.0.0
Thread model: posix
```
python3.4

```bash
git clone https://github.com/xrmx/uwsgi.git -b nsectimer2
cd uwsgi
export CC=clang
python3 uwsgiconfig.py --build
./uwsgi --wsgi-file t/python/timers.py --http :8080 --master
```
no problems so far

xrmx:
@Darvame Thanks for testing!

unbit:
Thanks to all

PR #1067, opened 2015-10-12
lua_plugin: uwsgi.log() multiprocess race

Darvame:
fixing uwsgi.log() multiprocess race (log splitted into parts), doing only one call of uwsgi_log()
removing "false" tabs and spaces from my previous editor

### DIFF
--- a/plugins/python/pyloader.c
+++ b/plugins/python/pyloader.c
@@ -711,7 +711,11 @@ PyObject *uwsgi_paste_loader(void *arg1) {
 		exit(UWSGI_FAILED_APP_CODE);
 	}
 
-	paste_arg = PyTuple_New(1);
+	if (up.paste_name) {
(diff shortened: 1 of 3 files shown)


Darvame:
I mean log corruption, when 2 process are calling log simultaneously:
```lua
uwsgi.log("value1", "-", "value2");
```
Before:
```bash
[uwsgi-lua] value1 *** WAIT! SOME MAY BE SOME IMPORTANT STUFF FROM THREAD 2 ***
 - value2
```
After:
```bash
*** WAIT! SOME MAY BE SOME IMPORTANT STUFF FROM THREAD 2 ***
[uwsgi-lua] value1 - value2
```

xrmx:
Could you please split whitespace cleanup and the race fix in two different commits please? Right now it's impossible to review on github.

Darvame:
reverting last commit was a bad idea

PR #1068, opened 2015-10-12
BUG: argument queue-blocksize is set to wrong variable

pavel-stratoscale:
variable should be uwsgi.queue_blocksize


unbit:
Thanks

PR #1072, opened 2015-10-14
add paths from UWSGI_INCLUDE to CFLAGS

urbaniak:
with that path we just need to specify the include paths once (UWSGI_INCLUDE) instead of twice (UWSGI_INCLUDE and CFLAGS)

### DIFF
--- a/uwsgiconfig.py
+++ b/uwsgiconfig.py
@@ -625,6 +625,9 @@ class uConf(object):
             '-D_FILE_OFFSET_BITS=64'
         ] + os.environ.get("CFLAGS", "").split() + self.get('cflags', '').split()
 
+        for path in self.include_path:
+            self.cflags.append('-I%s' % path)
+
         report['kernel'] = uwsgi_os
 
         if uwsgi_os == 'Linux':


xrmx:
The strange thing is include_path is always set but never used in both master and uwsgi-2.0, so your patch makes sense to me.

xrmx:
@urbaniak does this fix something for you or is it just something that makes sense?

urbaniak:
I was trying to compile uWSGI with SSL (for websocket handshake) and asyncio on OS X and it required those crazy environment variables:

    CFLAGS="-I$VIRTUAL_ENV/include/site/python3.4/ -I/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-migrator/sdk/MacOSX.sdk/usr/include/"
    UWSGI_INCLUDES="/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-migrator/sdk/MacOSX.sdk/usr/include/,$VIRTUAL_ENV/include/site/python3.4/"

With that path - it's much simpler, it requires only `UWSGI_INCLUDES`.

urbaniak:
include_path was used to detect headers (has_include() function), for example here checking for openssl: https://github.com/unbit/uwsgi/blob/abac960e62700117cb96af3cd22e27e04242e096/uwsgiconfig.py#L1222

urbaniak:
and after setting just UWSGI_INCLUDES (without CFLAGS) it detects openssl, tries to build with ssl support but failed at linking stage

xrmx:
Actually this behaviour is documented:
https://uwsgi-docs.readthedocs.org/en/latest/BuildSystem.html#uwsgi-includes

Was wondering if we shouldn't teach uwsgiconfig.py about compiler env variables like CPATH instead so that using the invocation below would work for both us and the compiler
CPATH="/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-migrator/sdk/MacOSX.sdk/usr/include/:$VIRTUAL_ENV/include/site/python3.4/"

PR #1077, opened 2015-10-19
Lua plugin and uWSGI application counter

jan-dx:
Lua plugin doesn't increment application counter while performing initialization. This cause `need-app` option to abort if only Lua applications are loaded/used.

### DIFF
--- a/plugins/lua/lua_plugin.c
+++ b/plugins/lua/lua_plugin.c
@@ -3102,6 +3102,8 @@ static void uwsgi_lua_init_apps() {
 			ulua_log("inited %d lua_State(s) for worker %d", uwsgi.threads, j + 1);
 		}
 	}
+
+	uwsgi_apps_cnt++;
 }
 
 static int uwsgi_lua_mule_msg(char *msg, size_t len) {
--- a/plugins/lua/lua_plugin.c
+++ b/plugins/lua/lua_plugin.c
@@ -3103,7 +3103,9 @@ static void uwsgi_lua_init_apps() {
 		}
 	}
 
-	uwsgi_apps_cnt++;
+	if (ulua.wsapi) {
+		uwsgi_apps_cnt++;
+	}
 }
 
 static int uwsgi_lua_mule_msg(char *msg, size_t len) {


Darvame:
need-app: "exit if no app can be loaded"
in your case if you have anything in config like:
--lua --lua-postload --lua-load (or shell mode)
you will be always recieve true for need-app option.

I think you should check the ulua.wsapi var
if '--lua' file fails to run (runtime exeption, or no wsapi entry point), lua plugin sets 'ulua.wsapi' to 0, to stop responding from 6th modifier1

You need to check this value before increment
```c
if(ulua.wsapi) uwsgi_apps_cnt++;
```

jan-dx:
For example, with following configuration

```
[uwsgi]
set-ph = piwot_root=%D/..
set-ph = piwot_config=%(piwot_root)/etc/%N.lua

plugin = lua
plugin = ugreen

master    = true
processes = 2
async     = 16
ugreen    = true

uid = www
gid = www

fastcgi-socket    = %(piwot_root)/run/piwot.sock
fastcgi-modifier1 = 6

lua = %(piwot_root)/etc/piwot.ws
force-cwd = %(piwot_root)/tmp
thunder-lock = true
need-app = true
```

uWSGI will abort because uwsgi_apps_cnt is not incremented.

Thanks for details about error conditions, and the code definitely requires error checks.

Darvame:
we can track the uwsgi_lua_init_state() success

jan-dx:
I see, then we need to be sure
* we have something to run in worker Lua state,
* what we have in Lua state is WSAPI application.

I'll review the change accordingly.

Darvame:
currently lua_state loads scripts in non stop mode:
any exeption during the load process will not stop it
we can track the inital success, stop it something goes wrong, and show it to  need-up increment

Darvame:
Or, just check out the wsapi app

jan-dx:
In `uwsgi_lua_init_state` function, if there is not WSAPI entry point; code logs message and set `ulua.wsapi` to 0. With current format, I think just checking against `ulua.wsapi` looks enough. Am I correct?

Darvame:
yes, is Lua_State lose wsapi entry point, lua plugin sets ulua.wsapi to 0, and any request to modifier1 6 will be logged as "worker%d[%d]: No WSAPI App. skip."

Darvame:
like i said before:
```c
uwsgi_apps_cnt++;
```
```c
if(ulua.wsapi) uwsgi_apps_cnt++;
```
looks like enough

xrmx:
@jan-dx please fix your branch and force push only the latest commit

jan-dx:
Did as you requested, PR #1078.

xrmx:
@jan-dx next time please just push -f on the same branch / pull request. And please always work on a different branch than master. thanks

PR #1078, opened 2015-10-19
lua_plugin: Increment application count while initializing Lua app.

ghost:
Supersedes #1077


xrmx:
@Darvame can i have your ack for the change please?

Darvame:
Hi, lua_state on worker has 3 entry points:
- rpc handler
- signal handler
- wsapi app (result from --lua file)

In this patch we checking wsapi app entry point and increment application count, if this point exists:
- hiding annoying message "*** no app loaded. going in full dynamic mode ***"
- support --need-app opt

Note: it's looks like --lazy --need-app --master does not work together, but this is not lua-plugin problem

xrmx:
@Darvame thanks, so the patch is fine for you and we should merge it right?

Darvame:
okay

PR #1082, opened 2015-10-21
failed proto_parser_status did not free wsgi_req

Darvame:


Darvame:
or did?

xrmx:
@Darvame speaking of commit styles i tend to use a present tense instead of past tense and describe what i am doing instead of wha ti'm trying to fix. In this case something like "fix wsgi_req leak in proto_parser_status". Anyway your commit message is fine, no need to change that. Thanks.

PR #1085, opened 2015-10-27
Remove leading and trailing quotation marks when processing yaml keys and values

jsivak:
I would like to correct the situation mentioned in these two posts:

http://lists.unbit.it/pipermail/uwsgi/2014-April/007278.html

http://serverfault.com/questions/688317/uwsgi-chdir-no-such-file-or-directory

I believe the simple changes I've made will allow me to use lines like:

chdir: "%d" 

In my .yaml file while still allowing PyYaml to process the file as well.

Thanks

### DIFF
--- a/core/yaml.c
+++ b/core/yaml.c
@@ -16,7 +16,7 @@ void yaml_rstrip(char *line) {
 	off_t i;
 
 	for (i = strlen(line) - 1; i >= 0; i--) {
-		if (line[i] == ' ' || line[i] == '\t') {
+		if (line[i] == ' ' || line[i] == '\t' || line[i] == '"') {
 			line[i] = 0;
 			continue;
 		}
@@ -30,7 +30,7 @@ char *yaml_lstrip(char *line) {
 	char *ptr = line;
 
 	for (i = 0; i < (int) strlen(line); i++) {
-		if (line[i] == ' ' || line[i] == '\t' || line[i] == '\r') {
+		if (line[i] == ' ' || line[i] == '\t' || line[i] == '\r' || line[i] == '"') {
 			ptr++;
 			continue;
 		}


xrmx:
The serverfault post has nothing to do with this, it's an ini file and it's full of strings without quotation except one :) Said that the patch looks trivial, wondering if this may break something else.

jsivak:
Added the serverfault post since it was "close in concept" to what I'm trying to do, which is allow uwsgi's native YAML processing to work while also allowing PyYaml's (more strict) processing work on the same .yaml file.

PyYaml refuses to process:

chdir: %d

and needs

chdir: "%d"

And yes, I agree that I'm not sure if something else will get broken, but I looked through the uwsgi Options docs and didn't find any obvious use of quoted variables/options. Also, the patch should only remove the first and last quotation marks, so if there are embedded quotation marks, they would be preserved.

jsivak:
BTW: I know that uwsgi supports compiling in the LibYAML processor (which does a better job with the "%d" handling), but the version that comes from pypi isn't configured to use the LibYAML library; so we'd have to custom build uwsgi at each new release if we tried to incorporate the LibYAML library.

xrmx:
Il 27/10/2015 21:28, John Sivak ha scritto:
> BTW: I know that uwsgi supports compiling in the LibYAML processor (which does a
> better job with the "%d" handling), but the version that comes from pypi isn't
> configured to use the LibYAML library; so we'd have to custom build uwsgi at
> each new release if we tried to incorporate the LibYAML library.

FYI something like this would work too:
UWSGI_PROFILE=/path/to/yourprofilewithyamlenabled.ini pip install uwsgi


-- 
Riccardo Magliocchetti
@rmistaken

http://menodizero.it

jsivak:
That does work; and I can add it to our setup script; but it would be nice if the native YAML processing in uwsgi was more friendly with PyYaml's needs.

unbit:
the problem are rails-style yaml files that popped up in the past years. Lot of uWSGI users integrated that syntax in their rails/rack apps. I fear this patch will break that behaviour, i leave this open as maybe we are able to support both with special syntax

xrmx:
Then instead of adding more options / syntax we can document better the libyaml possibility and be done with it?

jsivak:
I only work Python frameworks (Pylons, Pyramid).. has anyone tested/tried this patch with a rails application?

jsivak:
Is this change request too low of a priority or possibly to "dangerous" to apply given unbit's concern?:

> the problem are rails-style yaml files that popped up in the past years. Lot of uWSGI users integrated that syntax in their rails/rack apps. I fear this patch will break that behaviour, i leave this open as maybe we are able to support both with special syntax

jsivak:
To close this PR off:
We've gone back to using .ini files rather than .yaml files to configure uwsgi in order to be more consistent with general use seen on the Internet and to allow our dev's to be able to read the standard uWSGI docs without having to "convert" documented .ini style settings into .yaml settings.

Thanks for the PR consideration.

PR #1087, opened 2015-10-28
Eliminate blind excepts

jstasiak:
I don't think any of the blind excepts in this code matters as far as
reliability is concerned but I believe it's worth making the code more
explicit even if it's only for the educational value or in case someone
uses the code as an example.

* Where I was fairly certain what exception is expected I replaced
  "except:" with "except SomeConcreteClass:" (or otherwise simplified
  the code)
* Where I wasn't sure what exaclty is being expected I just put
  "except Exception:" to at least skip the "special" exceptions
* Two remaining cases I replaced with "except BaseException:" which I
  believe makes the code tiny bit more explicit.

### DIFF
--- a/examples/uwsgistatus.py
+++ b/examples/uwsgistatus.py
@@ -26,12 +26,12 @@ def application(env, start_response):
 
     try:
         yield "mode: <b>%s</b><br/>" % uwsgi.mode
-    except:
+    except Exception:
         pass
 
     try:
         yield "pidfile: <b>%s</b><br/>" % uwsgi.pidfile
-    except:
+    except Exception:
         pass
 
     yield "<h2>Hooks</h2>"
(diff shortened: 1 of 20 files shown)


xrmx:
Thanks for the patch! I think it could be improved though. First i don't see how except Exception is better than just except, it is still a catch all, but you have already converted tons of them so we can live with that :) I've made a few comments here and there. Please split the conversion in multiple commits: one for uwsgiconfig, one for tests/ and examples/, one for plugins/, one for uwsgidecorators.py.

jstasiak:
Catching everything (read: ``BaseException``) also intercepts ``KeyboardInterrupt`` and ``SystemExit`` exceptions which I'm willing to say none of those blocks should be silencing. It may be argued that it's either unlikely to get them in those places or it doesn't matter (in the example code for example) but I believe it's important to give a good example.

Please see the patch split into separate commits now and modified according to your notes. Feel free to shuffle the changes however you want if needed.

xrmx:
Merged, thanks!

PR #1091, opened 2015-10-29
Introduce a new api: reload_worker

yueyoum:
with the new api, we not only can reload whole uwsgi server (via `uwsgi.reload()`) ,
but also can reload one worker.

`uwsgi.reload_worker()` with no argument will reload the worker which call the api.
`uwsgi.reload_worker(worker_id)` call with `worker_id` argument will reload the special worker.

when `worker_id` is not valid, ValueError raised.


yueyoum:
@unbit   how about this feature ?

PR #1101, opened 2015-11-03
fix pyloader error message on callable Fix #919

### LINKED ISSUE #919: error message always shows default callable name, very confusing
I'm a new uwsgi user and I got thrown for a loop by the logging error here:
https://github.com/unbit/uwsgi/blob/e45f710694f18052aa00ee5bb866d9aeca76fb80/plugins/python/pyloader.c#L605

My situation was that `callable` was set to `app` in the uwsgi conf but my uwsgi program (a flask app) used `APP`. The log message put my debugging efforts on a false trail by making me think callable was set to its default, i.e. my configuration was wrong.

Please remove the hardcoded string `"application"` from the error message. Even better, show the actual value of the `callable` param.

depaolim:
Print the actual callable instead of a fixed 'application'.

Fix #919

### DIFF
--- a/plugins/python/pyloader.c
+++ b/plugins/python/pyloader.c
@@ -602,12 +602,12 @@ PyObject *uwsgi_file_loader(void *arg1) {
 		Py_DECREF(wsgi_file_dict);
 		Py_DECREF(wsgi_file_module);
                 free(py_filename);
-		uwsgi_log( "unable to find \"application\" callable in file %s\n", filename);
+		uwsgi_log( "unable to find \"%s\" callable in file %s\n", callable, filename);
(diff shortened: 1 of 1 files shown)


xrmx:
@depaolim thanks for the patch! any chance you can get a bit more descriptive in the commit message please? Something like
```
Print the actual callable instead of a fixed 'application'.

Fix #919" 
```
would be enough. While at it you can move the bug reference in the commit title.

depaolim:
hope it is ok, otherwise let me know. Thanks!

xrmx:
@depaolim i think you haven't (force) pushed your updated commit :)

PR #1102, opened 2015-11-03
Improve pyloader error message on missing callable

### LINKED ISSUE #919: error message always shows default callable name, very confusing

depaolim:
Print the actual callable instead of a generic 'application'.

Fix #919

### DIFF
--- a/plugins/python/pyloader.c
+++ b/plugins/python/pyloader.c
@@ -602,12 +602,12 @@ PyObject *uwsgi_file_loader(void *arg1) {
 		Py_DECREF(wsgi_file_dict);
 		Py_DECREF(wsgi_file_module);
                 free(py_filename);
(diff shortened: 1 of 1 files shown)


xrmx:
Merged, thanks!

PR #1106, opened 2015-11-04
Feature:  Worker warehouse

yueyoum:
Add a warehouse (128KB) in worker, for worker store data, 

(e.g. worker can save any thing (string) in it's warehouse. and all workers warehouse data can be obtained convenient. so this makes do statistics simply)
    
#### NEW:
*   `uwsgi.worker_warehouse_set(data)`
    set data(string) in this worker warehouse
    
*   `uwsgi.worker_warehouse_get()`
    get data from this worker warehouse
    
*   `uwsgi.worker_warehouse_clear()`
    clear this worker warehouse
    
*   `uwsgi.workers_warehouse()`
    get all workers warehouse data.
    return as tuple, data ordered by worker id
    
#### CHANGE:
*   `uwsgi.workers()`
    now this api add an optional keyword argument: `warehouse`
    if no argument, or warehouse=False,
    the return value does't change
    
    if warehouse=True,
    the return value also contains `warehouse` in all workers

#### TEST

see `tests/worker_warehouse.py`

I have tested many times to ensure that this APIs are working correctly and no memory leak.

### DIFF
--- a/core/master_utils.c
+++ b/core/master_utils.c
@@ -775,6 +775,8 @@ int uwsgi_respawn_worker(int wid) {
 			memset(uwsgi.workers[uwsgi.mywid].cores[i].buffer, 0, sizeof(struct uwsgi_header));
 		}
 
+		memset(uwsgi.workers[uwsgi.mywid].warehouse, '\0', sizeof(uwsgi.workers[uwsgi.mywid].warehouse));
+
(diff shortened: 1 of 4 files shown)


yueyoum:
DO NOT MERGE, 

I will rewrite this feature!

Issue #1113, opened 2015-11-16
mod_proxy_uwsgi does not work with HTTP/2


fladi:
The Problem seems to be the the call to [`apr_date_checkmask`](https://github.com/unbit/uwsgi/blob/master/apache2/mod_proxy_uwsgi.c#L329) which expects a HTTP version in the form of `HTTP#.#`. This does not support HTTP/2 status lines in the form of `HTTP/#`. Overall, I think that `apr_date_checkmask` should be replaced by `apr_strmatch`.

I'll make a PR that takes HTTP/2 into account with `apr_date_checkmask` to fix this problem and maybe later come up with a version for `apr_strmatch`.

xrmx:
Patch has been merged, closing.

fladi:
Thanks!

PR #1114, opened 2015-11-16
Allow HTTP/2 status format in mod_proxy_uwsgi response.

fladi:
Extend the sanity check for the HTTP version in the status line to accept both
HTTP/#.# and HTTP/# strings. This allows HTTP/2 to work, resolving #1113.


xrmx:
Thanks for the patch! The whitespace is a bit off though, please use tabs for first indent and the spaces to align to the opening parens.

xrmx:
@fladi indentation looks still off, it should be just a tab and 4 spaces each line but you added 2 tabs and on a varying amount of spaces. Also you have to squash patches into one. Please wait to rework before the code changes has been reviewed though.

PR #1115, opened 2015-11-20
Fix inheriting INET address 0.0.0.0

methane:
fixing problem reported in ML


xrmx:
Merged, thanks a lot!

PR #1118, opened 2015-11-26
Add new method to read spooler file from python

Shir0kamii:
Hi,

I added this feature, but as it may benefit others, maybe you could merge it.

I can adjust the modifications if needs be. This code introduce redundancy with the core but I didn't want to change it. Another solution would be to break spooler_manage_task into multiple functions to use them in this feature.

cheers,
Shir0kamii

### DIFF
--- a/core/spooler.c
+++ b/core/spooler.c
@@ -535,6 +535,60 @@ static void spooler_readdir(struct uwsgi_spooler *uspool, char *dir) {
 	}
 }
 
+int uwsgi_spooler_read_header(char *task, int spool_fd, struct uwsgi_header *uh) {
+
+	// check if the file is locked by another process
+	if (uwsgi_fcntl_is_locked(spool_fd)) {
+		uwsgi_protected_close(spool_fd);
+		return -1;
+	}
+
+	// unlink() can destroy the lock !!!
+	if (access(task, R_OK|W_OK)) {
+		uwsgi_protected_close(spool_fd);
+		return -1;
+	}
+
+	ssize_t rlen = uwsgi_protected_read(spool_fd, uh, 4);
+
+	if (rlen != 4) {
+		// it could be here for broken file or just opened one
+		if (rlen < 0)
+			uwsgi_error("spooler_manage_task()/read()");
+		uwsgi_protected_close(spool_fd);
+		return -1;
+	}
+
+#ifdef __BIG_ENDIAN__
+	uh->_pktsize = uwsgi_swap16(uh->_pktsize);
+#endif
+
+	return 0;
+}
+
+int uwsgi_spooler_read_content(int spool_fd, char *spool_buf, char **body, size_t *body_len, struct uwsgi_header *uh, struct stat *sf_lstat) {
+
+	if (uwsgi_protected_read(spool_fd, spool_buf, uh->_pktsize) != uh->_pktsize) {
+		uwsgi_error("spooler_manage_task()/read()");
+		uwsgi_protected_close(spool_fd);
+		return 1;
+	}
+
+	// body available ?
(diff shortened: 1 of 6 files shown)


xrmx:
Thanks for the patch! Your commit message lacks quite a bit of information though. Why are you adding this functionality? what's the use case? what does the current system lacks?

ultrabug:
@xrmx @unbit : @Shir0kamii is a co-worker of mine and I kindly asked him to develop this feature because we have a use case where we have to be able to read a spool file directly from the uWSGI spool directory.

The reason is that there is no "batch" processing possible with the spooler and we needed one while still benefiting from the strong spooling feature of uWSGI.

Our production use case :
- lots of "events" get spooled for persistence
- I need to batch process and send them over the network (say; 5000 batching)
- When my networked batch is ACK'ed on the other side, I can safely remove the 5000 spool files I got

So far I wrote a nasty python hack code to "read & parse" the uwsgi spool files. I remember some times ago asking Roberto for a clean way to do it and he said there was none unless you do it in C... @Shir0kamii is the C guy we didn't have so far ;)

FYI, we'll maintain our own fork of uWSGI for our production servers while this feature gets validated/released.

xrmx:
@ultrabug thanks for the heads up. Haven't looked at spooler code but if can avoid duplicating stuff by extracting smaller helpers of course that's preferred :) From a quick look you are using spaces for indenting while it should be tabs. We are having a sprint next tuesday, maybe you can partecipate and work on this?

Shir0kamii:
@xrmx I improved the code as much as I can

Let me know if there's something you'd like me to change

xrmx:
@Shir0kamii looks a lot better! Will take a thorough look tuesday evening. Thanks

xrmx:
@Shir0kamii can you add a test in t/spooler or tests/ to show how the api is supposed to be used please?

Shir0kamii:
@xrmx I hope it will be enough. Even though I can't participate, I'll be on #uwsgi (IRC) during the sprint, you can ask me anything if you need

xrmx:
@Shir0kamii thanks for the update, haven't had time yesterday to review your stuff. Could you please explain a bit whay the reader and consumer are not racing against each other please? I thought you did want to batch jobs and consume them somewhere else so the consumer function surprises me.

Shir0kamii:
@xrmx I thought of an use case where you spool failing jobs to restart them later. You might want to have informations about these failing jobs (reader) while trying to restart them (consumer).

The real case we have at work is that, because spooler processes are slow, we want to use mules instead. For this purpose, we have to read the files and it'll be cleaner if there's a uwsgi built-in function.

xrmx:
@Shir0kamii mmm so you want the persistence if the task fail but you don't want the coarse grained spooler frequency. Wouldn't it better to store stuff in a cache and then signal mules instead? I mean while the ability to inspect spooler jobs from python may be handy and thus can merge the example code is not really something one want to implement it its application :)

Shir0kamii:
@xrmx We use the spooler for the persistence indeed, so using a cache wouldn't be a good solution for our use case. I agree that the example code may be unadapted in real use case, but I think it's a good example to show the API. With an example imitating the spooler behaviour like we do, people could wonder why, so what would you suggest ?

xrmx:
@Shir0kamii merged this in master, thanks!

xrmx:
@Shir0kamii i think that removing the consumer and adding --spooler-external to uwsgi configuration example may give some hints. Does it make sense to you?

Shir0kamii:
@xrmx Yes, I think it does. Do you want me to make these modifications even though you merged it ?

xrmx:
@Shir0kamii please open a new pull request. Usually the first line line of the example file contains the uwsgi configuration needed for running it. Thanks!

PR #1117, opened 2015-11-25
Avoiding #ifdefs that break statements.

flaviommedeiros:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

PR #1121, opened 2015-12-01
core/utils: cleanup uwsgi_resolve_ip ifdefery

xrmx:
Use a macro to check that ip address is of AF_INET / AF_INET6
families.

See https://github.com/unbit/uwsgi/pull/1117


xrmx:
Merged.

PR #1131, opened 2015-12-17
Adding del_cron function to cron interface

blablacio:
I've got a use case, where a cron schedule is controlled through Django admin interface. I run a timer which monitors for changes in the setting along these lines:

```python
from uwsgi import add_cron, del_cron           
from uwsgidecorators import timer, signal

@signal(88)
def hello(args):
  print 'hello'

@timer(5)
def monitor_setting(args):
  signum = 88
  cached = cache.get('cron_setting', None)
  new_val = getattr(config, 'cron_setting')
  func_args = str(new_val).split(',')
  func_args = map(str.strip, func_args)
  func_args = map(int, func_args)
    
  if cached is not None:
    if cached != new_val:
      del_cron(signum)
      add_cron(signum, *func_args)
      cache.set('cron_setting', new_val, timeout=None)
  else:
    add_cron(signum, *func_args)
    cache.set('cron_setting', new_val, timeout=None)
```

So, I needed the del_cron part in order to de-register a signal and re-register it with a new schedule with add_cron. Not sure if something like
```c
int uwsgi_signal_update_cron(uint8_t sig, int minute, int hour, int day, int month, int week)
```
would actually make more sense for directly updating the schedule, instead of removing it and re-adding it to ushared->cron.


xrmx:
Uhm, if we are going to add the ability to remove a cron at runtime i think it would be make sense to let people list all the registered cron jobs so one does not have to do the accounting himself. What do you think? cc @unbit

blablacio:
@xrmx @unbit I guess it makes sense in cases when you're directly decorating with ```@cron```, because a signal is automatically created and the user doesn't know what signal number is.

Just pushed some code that adds ```list_cron``` function to Python plugin. Haven't thoroughly tested, but seems to be working on my end.

blablacio:
@xrmx I made the changes you suggested anyway, seem to be working fine. Let me know if that looks good to you.

blablacio:
@xrmx Would you be interested to get traction on this one? Getting it into an official release would be nice...

s265925:
what happened to this change? I need to delete/list :)

PR #1132, opened 2015-12-18
Add logic option if-hostname-match

Shir0kamii:
Hi,

I added a new logic option to uwsgi called if-hostname-match that take a regex and evaluate to true if the hostname match the regex.

In my team, we have multiple servers running uwsgi and setting several if-hostname condition blocks is rather long, so it would be great if we could just use a regex.

What do you think about this ?

### DIFF
--- a/core/config.c
+++ b/core/config.c
@@ -1,4 +1,5 @@
 #include <uwsgi.h>
+#include <regex.h>
 
 /*
 
@@ -312,6 +313,40 @@ int uwsgi_logic_opt_if_not_hostname(char *key, char *value) {
         }
 
         return 0;
+}
+
+int uwsgi_logic_opt_if_hostname_match(char *key, char *value) {
+
+	regex_t preg;
+
+	if (regcomp(&preg, uwsgi.logic_opt_data, REG_NOSUB)) {
+		uwsgi_error("regcomp");
+		exit(1);
+	}
+
+	if (!regexec(&preg, uwsgi.hostname, 0, NULL, 0)) {
+		add_exported_option(key, uwsgi_substitute(value, "%(_)", uwsgi.logic_opt_data), 0);
+		return 1;
+	}
+
+	return 0;
+}
+
+int uwsgi_logic_opt_if_not_hostname_match(char *key, char *value) {
+
+	regex_t preg;
+
+	if (regcomp(&preg, uwsgi.logic_opt_data, REG_NOSUB)) {
+		uwsgi_error("regcomp");
+		exit(1);
+	}
+
+	if (regexec(&preg, uwsgi.hostname, 0, NULL, 0)) {
+		add_exported_option(key, uwsgi_substitute(value, "%(_)", uwsgi.logic_opt_data), 0);
+		return 1;
+	}
+
+	return 0;
 }
 
 int uwsgi_count_options(struct uwsgi_option *uopt) {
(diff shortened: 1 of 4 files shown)


Lujeni:
+1 :cactus: 
@Shir0kamii can you add a configuration example of this feature pls?

Shir0kamii:
Of course, here is a simple example of configuration using if-hostname-match : 

```
…
threads = 10

if-hostname-match = prod
    threads = 20
end-if = true
…
```

If you have several servers used for production containing "prod" in their hostname, you can use these configuration lines to use more threads on production servers.

btall:
:+1:

ramnes:
Definitely :+1:

xrmx:
@Shir0kamii thanks for the patch! Could you please tell to your coworkers that +1 are even more annoying than usual if they came from people of the same company? :)

xrmx:
From a quick look the patch is fine, we may want in future to try to reuse some code for similar matches though. After this go in it would be nice to update uwsgi-docs with this and the spooler api addition.

xrmx:
Well in this case, why one cannot be the negation of the other one? Like:
```
int uwsgi_logic_opt_if_not_hostname_match(char *key, char *value) {
   return !uwsgi_logic_opt_if_hostname_match(key, value);
}
```

ramnes:
Alright, fair enough. Next time I'll make a fake account so I can +1 great features without being caught!

Shir0kamii:
Of course I will update the documentation as soon as I can.

I'll try to clean up the code the next time I work on it. For now, spaces and tabulations are mixed in the codebase, should I replace spaces with tabulation or the inverse ? Also, do you prefer a specific standard for the code ?

For example, the goto directives. I found some in the code (note that it's inconsistent with most of the codebase) and it's usually considered bad practice to use them. So should I change them into loop when I encounter one or is there a good reason for them to be there ?

unbit:
@Shir0kamii goto's in c are not a bad practice (it is a common misconception), you will find tons of them in the linux kernel, nginx, and friends :)

Darvame:
They should add "like" feature for pull requests

Shir0kamii:
@unbit I didn't know, my bad.

unbit:
@xrmx the feature is interesting, my only fear is in the use of the regex header that is not ubiquitous. I would prefer the usage of uWSGI pcre api. And yes, generalizing the feature (to allow regexps in other if-based opts) would be cool but would require re-writing some of the internal api. @Shir0kamii would you like to try refactor it using pcre api ?

```c
uwsgi_regexp_build(char *regexp, pcre ** pattern, pcre_extra ** pattern_extra);
uwsgi_regexp_match(pcre * pattern, pcre_extra * pattern_extra, char *subject, int length);
```

Shir0kamii:
@unbit I tried what you asked, please tell me if it seems good to you.

Also I added a function that should be useful if you want to implement other options using regexp. It should reduce the code duplication.

unbit:
@Shir0kamii nice work. Even the api addition seems a handy thing. I will merge it and backport to 2.0 too. Thanks

Shir0kamii:
@unbit Thank you ! Could you please backport the spooler feature at the same time ?

unbit:
@Shir0kamii which one ?

Shir0kamii:
@unbit I'm talking about #1118

PR #1138, opened 2015-12-30
async proto_parser_status fail backport

Darvame:
looks like this is realy solves my problem with deadlock and 100% cpu load, when I 'attack' my server with 'ab' util


unbit:
thanks

PR #1152, opened 2016-01-12
Luap weird spawn

Darvame:
spawning lua_States with old method (my previous method sucks)
Init steps:
--lua-load is .preinit_apps
--lua-postload is .postinit_apps
--lua is .init_apps
--lua-postfork is .post_fork

Also add async_id() and fd() for websocket_handlers (is_alive() now returns bool)


Darvame:
Sec, not ok yet

Darvame:
looks fine for me right now

PR #1153, opened 2016-01-12
fix emperor SIGURG feature (cleanup blacklist)

Shir0kamii:
Hi,

@unbit told me about a SIGURG feature for emperor to clean its blacklist but it didn't work. Here is the fix I writed.

If i understand clearly, the previous code was taking the vassal list which does not contains blacklisted vassals. Now I simply take the blacklist and free all of its items.

I will also make a PR for the documentation.

### DIFF
--- a/core/emperor.c
+++ b/core/emperor.c
@@ -1941,11 +1941,12 @@ static void emperor_wakeup(int sn) {
 }
 
 static void emperor_cleanup(int signum) {
-	uwsgi_log_verbose("[uwsgi-emperor] cleaning up blacklist ...\n");
-	struct uwsgi_instance *ui_current = ui;
-	while (ui_current->ui_next) {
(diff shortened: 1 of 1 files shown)


xrmx:
A commit message would be appreciated. Explaining why this fix things would be event better :)

Shir0kamii:
I will do it right now, since I should have done it since the beginning !

Shir0kamii:
@xrmx Any chance that you check this PR soon, to know if I have to make change to it ?

unbit:
sorry for the late, i will review it during the week end

unbit:
So, the bug is related to the fact that blacklisted instances are not listed in the healthy instances, so the cleanup procedure will never find a blacklisted record.

Your patch does the right thing, but you missed to reset emperor_blacklist as NULL

Shir0kamii:
@unbit Sorry about forgetting that. It's done now.

PR #1165, opened 2016-01-26
uwsgi_cache_magic_get_callback

Darvame:
let's plugin to decide how to copy and alloc memory

### DIFF
--- a/core/cache.c
+++ b/core/cache.c
@@ -26,6 +26,7 @@ extern struct uwsgi_server uwsgi;
 
 static void cache_full(struct uwsgi_cache *uc) {
 	uint64_t i;
+	int force_clear;
 
 	if (!uc->ignore_full) {
         	if (uc->purge_lru)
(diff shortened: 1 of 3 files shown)


Darvame:
just avoid some unnecessary memory copy while doing cache_get stuff
Normaly:
- calling uwsgi_cache_magic_get()
- func places lock
- finds something
- copy and allocate memory
- lifting lock
- gives new chunk to plugin
- plugin copies it again!
- plugin frees given memory chunk

What callback does:
- calling uwsgi_cache_magic_get_callback()
- func places lock
- finds something
- calls callback handler, where plugin decides how to place this thing and gives the status code
- lifting lock
- forwards handler status code and exits from from uwsgi_cache_magic_get_callback()

unbit:
So, if i understand correctly, this is an optimization allowing plugins to override the memory model of the cache system (that is written for easy porting between languages). Right ?

Darvame:
We just 'ask' the plugin, while standing between the read(or write) locks, where exactly to copy data from current key, when getting it

PR #1169, opened 2016-01-27
too many file descriptors!!!!

Darvame:
I think we should clean this before exit

### DIFF
--- a/core/cache.c
+++ b/core/cache.c
@@ -1735,6 +1735,7 @@ char *uwsgi_cache_magic_get(char *key, uint16_t keylen, uint64_t *vallen, uint64
 		}
 
 		// now the magic, we dereference the internal buffer and return it to the caller
+		close(fd);
 		char *value = ub->buf;
 		ub->buf = NULL;
 		uwsgi_buffer_destroy(ub);
@@ -1811,6 +1812,8 @@ int uwsgi_cache_magic_exists(char *key, uint16_t keylen, char *cache) {
 			return 0;
                 }
 
+		close(fd);
+		uwsgi_buffer_destroy(ub);
 		return 1;
         }
 
@@ -1884,6 +1887,7 @@ int uwsgi_cache_magic_set(char *key, uint16_t keylen, char *value, uint64_t vall
                         return -1;
                 }
 
+		close(fd);
 		uwsgi_buffer_destroy(ub);
 		return 0;
 
@@ -1956,6 +1960,8 @@ int uwsgi_cache_magic_del(char *key, uint16_t keylen, char *cache) {
                         return -1;
                 }
 
+				close(fd);
+				uwsgi_buffer_destroy(ub);
                 return 0;
         }
 
@@ -2029,6 +2035,8 @@ int uwsgi_cache_magic_clear(char *cache) {
                         return -1;
                 }
 
+				close(fd);
+				uwsgi_buffer_destroy(ub);
                 return 0;
         }
 


Darvame:
**clear_on_full=1** - drop cache when got cache if full error
**sweep_on_full=N** - sweep cache when got cache if full error (N - is min sweep interval in seconds, must be more 1), no full cache drop if fail to free items
**clear_on_full=1,sweep_on_full=N** - BOTH  sweep cache when got cache if full error or drop it in case of fail (0 items freed)
**sweep_on_init=1** - sweep cache after cache's locks init (log messge about inited cache), usefull to remove expierd itmes from file

Darvame:
must be ok, when restoring from file or node

Darvame:
sec, we can update next sweep time on cache_fix()
and run sweep_on_init() only if next time lower than now()

Darvame:
restoring next_scan value on cache_fix()
run sweep on init if have active sweeper thread or sweep_on_init=1 and next_scan is expired

goir:
Thanks you so much for fixing this!
What does @unbit think ? Can we get this in for the next release ?

xrmx:
@goir have you tried it and this solves your issues?

Darvame:
@xrmx this solves issues:
- fd leak after succes remote cache call
- uc->n_items value after cache_fix() (n_items used in metric)
- restoring sweeper (next_scan was always 0 after restoring)

All 3 are critical anyway

goir:
@xrmx sorry for the late response. Yes it does fix the problem. When loading a "corrupted" cache file it gets correctly cleaned and all expired items are removed. Adding new ones is working perfectly.

xrmx:
@Darvame have applied the first two patches to both uwsgi-2.0 and master after adding a sensible commit message. Now trying to understand your last patch, also splitting the addition of the sweep on init stuff from the cache_full changes.
Plus some cache related changes are in uwsgi-2.0 but not in master just to make it harder :)

Darvame:
uwsgi_cache_sweep_on_init() and cache_sweeper_free_items()
almost look the same, but got a priblem with the uwsgi.current_time, it is 0 on init

xrmx:
@Darvame I've splitted your changes there https://github.com/xrmx/uwsgi/tree/darvamesweep . Now need to forward port to master some old cache changes to match uwsgi-2.0 and then i can merge your fixes.

xrmx:
@Darvame i've fscked the author, i'll fix that of course.

Darvame:
@xrmx u wot?

xrmx:
This is the branch for master:
https://github.com/xrmx/uwsgi/tree/cachefixesmaster

@unbit do you think we can avoid the sweep_on_init flag and just do the sweep on init? i fail to see why we don't want to reuse cache_sweeper_free_items, the only difference i see is the parameters passed to the del function. AFAIU the different use of the current time for checking the expiration should not make any difference.

Darvame:
sweep_on_init is an addition to other flags like sweep_on_full and etc, to run sweep_on_init without a sweeper thread. You can remove it and use cache_sweeper_free_items() instead, but _!uc->next_scan || uc->next_scan > (uint64_t)uwsgi.current_time_ is always true on init stage, and cache_sweeper_free_items() will always exit coz uwsgi.current_time is 0

xrmx:
@Darvame and having uwsgi.current_time being 0 isn't a bug? Or can't we workaround it calling uwsgi_now() if it's 0?

Darvame:
@xrmx  init uwsgi.current_time earlier, or use uwsgi_now() if uwsgi.current_time is 0

PR #1173, opened 2016-02-03
Fix --catch-exceptions causing a segfault in Python 3.5.

johnnyg:
This bug arose due to the fact that the traceback module was overhauled
in python 3.5: https://hg.python.org/cpython/rev/73afda5a4e4c
The crash was due to PyTuple_GetItem returning NULL as a traceback is
no longer a list of tuples but rather a list of objects (which support indexing).


unbit:
The patch looks good, i need only to check it for python 2.4 (we still support it)

johnnyg:
Yeah I wasn't sure what the earliest version of Python was that you supported.
As far as I can't tell it from the code/docs, the sequence API existed in 2.4 (https://docs.python.org/release/2.4.4/api/sequence-structs.html) so this patch *should* work in 2.4 but obviously actually testing is the best way to check.

xrmx:
Merged in #2672, thanks!

PR #1175, opened 2016-02-04
Don't increment emperor_throttle_level during emperor warmup.

kevinxucs:
During startup phase, if there are many vassals, emperor_throttle_level will be increased to a very high value thus causing the first few on-demand vassals to have high throttle time.


unbit:
Thanks (and sorry for the late)

PR #1185, opened 2016-02-19
Fix inet_ntop buffer size

orivej:
Buffer size specified in a call to `inet_ntop` did not reflect the size of the buffer, `cs->client_address` (46 bytes in case of AF_INET6).  IPv6 addresses whose string representation was longer than the specified size (28) were misrepresented as `0.0.0.0`.

No-AF_INET6 client_address buffer size also seems off, although I could not test the change, and I doubt that your Travis setup will test it.


Roguelazer:
correctly representing IPv6 addresses will also require the changes I put in #1203, I believe.

orivej:
Thanks. My change fixes http plugin (running with `--http`) and yours fixes the core `--http-socket` mode. It seems our changes are independent.

Issue #1191, opened 2016-03-02
Add an spooler-create-directory option


ultrabug:
Right now, if the spooler directory is not present, the spooler process dies horribly.

When deploying applications, they should be autonomous in creating their own spool directory if it does not exist (so we don't need a sysadmin to mkdir it..).

I'd like to add a `spooler-create-directory` option (defaulting to false to remain backward compatible) to allow the spooler process to create it if needed (and permitted ofc).

We will submit a PR for this unless you guys are against the proposal.

Ping @Shir0kamii, TODO plz thanks !

xrmx:
Can't the tooling you have for deploying uwsgi create the spooler dir too? What if the mkdir for whatever reason fails? Instead of adding an option i'd very much add the mkdir. If it fails we continue to die horribly otherwise you are up and running. We have enough options i think.

ultrabug:
@xrmx well when the goal of the whole stack is to give freedom to the developers, no. They code and deploy their own uWSGI INI and this should be enough to ship their project with a stack agnostic continuous delivery tool.

IMHO it's okay that packagers create a basic spool directory root folder (I do create /var/spool/uwsgi on the uwsgi package for Gentoo Linux) and it's okay that sysadmins adjust their rights to match the CD tool deployment user but this should be all.

I'm all for not adding a new option, then maybe allow us to propose that uWSGI will try to create the directory it needs and still die horribly like as of today if that fails. I'd better see it tried and failed than failed without trying...

After all, **uWSGI already does create the sub directories when you use the `priority` feature**, that's not asking much I guess !

unbit:
Hi @ultrabug i think you can already use the hooking features.

Check this template (its a django one used to generate our customer's vassals)

https://github.com/unbit/uwsgi.it/blob/master/uwsgi_it_api/uwsgi_it_api/templates/vassal.ini

it automatically creates directories when required

ultrabug:
A bit overkill and forget-prone @unbit but okay, doable thanks :) wanna close this then ?

xrmx:
@ultrabug sorry if i wasn't clear, it's fine for me to do the mkdir by default but i think it's really overkill to add an option to do so.

ultrabug:
ok @xrmx cool.

@Shir0kamii go plz :heart:

xrmx:
Pull request has been merged, closing.

PR #1200, opened 2016-03-08
Create the spooler directory if it doesn't exist

Shir0kamii:
Hi !

As discussed in #1191, it would be a good idea to let uwsgi create the spooler if it can rather than die without trying.

Here is the PR proposing this modification. With it, uwsgi dies only if there is no directory and if it can't create it either.

Cheers


Shir0kamii:
@xrmx Alright, sorry for this one ! It should be good now

Shir0kamii:
@xrmx Sorry for the time I took, but it should be ok now

xrmx:
Applied, thanks

PR #1203, opened 2016-03-11
correctly handle IPv6 sources in proto/http

Roguelazer:
I'd like to be able to bind HTTP sockets to IPv6 addresses and get a `REMOTE_ADDR` other than 0.0.0.0. As far as I can tell, this patch will implement that.

Not sure about the change to `uwsgi.h`; I think it's more correct (especially for any platform where `UNIX_PATH_MAX` isn't Linux's/Solaris's 108 bytes), but I would understand if folk wanted to avoid changing this structure and keep relying on the implementation detail of `sockaddr_in` and `sockaddr_in6` fitting in a `sockaddr_un`.

### DIFF
--- a/proto/http.c
+++ b/proto/http.c
@@ -822,17 +822,17 @@ void uwsgi_proto_http11_close(struct wsgi_request *wsgi_req) {
 }
 
 int uwsgi_proto_http11_accept(struct wsgi_request *wsgi_req, int fd) {
-        if (wsgi_req->socket->retry[wsgi_req->async_id]) {
-                wsgi_req->fd = wsgi_req->socket->fd_threads[wsgi_req->async_id];
+	if (wsgi_req->socket->retry[wsgi_req->async_id]) {
+		wsgi_req->fd = wsgi_req->socket->fd_threads[wsgi_req->async_id];
 		wsgi_req->c_len = sizeof(struct sockaddr_un);
 		int ret = getsockname(wsgi_req->fd, (struct sockaddr *) &wsgi_req->client_addr, (socklen_t *) &wsgi_req->c_len);
-                if (ret < 0)
+		if (ret < 0)
 			goto error;
-                ret = uwsgi_wait_read_req(wsgi_req);
-                if (ret <= 0)
+		ret = uwsgi_wait_read_req(wsgi_req);
+		if (ret <= 0)
 			goto error;
-                return wsgi_req->socket->fd_threads[wsgi_req->async_id];
-        }
+		return wsgi_req->socket->fd_threads[wsgi_req->async_id];
+	}
 	return uwsgi_proto_base_accept(wsgi_req, fd);
 
 error:
--- a/proto/http.c
+++ b/proto/http.c
@@ -337,14 +337,12 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
 					/* check if it's an IPv6-mapped-IPv4 address and, if so,
 					 * represent it as an IPv4 address
 					 *
-					 * Note: This macro isn't in any version of the POSIX
-					 * spec that I've found, but it's implemented on BSD,
-					 * Linux, and OS X.
+					 * these IPv6 macros are defined in POSIX.1-2001.
 					 */
 					if (IN6_IS_ADDR_V4MAPPED(&http_sin->sin6_addr)) {
 						/* just grab the last 4 bytes and pretend they're
-						 * IPv4. Linux has a convenience wrapped to get the
-						 * componenets word-wise, but OS X/BSD do not.
+						 * IPv4. None of the word/half-word convenience
+						 * functions are in POSIX, so just stick to .s6_addr
 						 */
 						uint32_t in4_addr = ((uint32_t*)http_sin->sin6_addr.s6_addr)[3];
 						if (inet_ntop(AF_INET, (void*)&in4_addr, ip, INET_ADDRSTRLEN)) {
--- a/proto/http.c
+++ b/proto/http.c
@@ -332,7 +332,7 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
 		switch (http_sa->sa_family) {
 			case AF_INET6:
 				{
-					memset(ip, 0, INET6_ADDRSTRLEN+1);
+					memset(ip, 0, sizeof(ip));
 					struct sockaddr_in6* http_sin = (struct sockaddr_in6*)http_sa;
 					/* check if it's an IPv6-mapped-IPv4 address and, if so,
 					 * represent it as an IPv4 address
@@ -367,7 +367,7 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
 			default:
 				{
 					struct sockaddr_in* http_sin = (struct sockaddr_in*)http_sa;
-					memset(ip, 0, INET_ADDRSTRLEN+1);
+					memset(ip, 0, sizeof(ip));
 					if (inet_ntop(AF_INET, (void *) &http_sin->sin_addr, ip, INET_ADDRSTRLEN)) {
 						wsgi_req->len += proto_base_add_uwsgi_var(wsgi_req, "REMOTE_ADDR", 11, ip, strlen(ip));
 					}
--- a/proto/http.c
+++ b/proto/http.c
@@ -208,7 +208,7 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
 	char *base = ptr;
 	char *query_string = NULL;
 	char ip[INET6_ADDRSTRLEN+1];
-	struct sockaddr *http_sa = (struct sockaddr *) &wsgi_req->c_addr;
+	struct sockaddr *http_sa = (struct sockaddr *) &wsgi_req->client_addr;
 	char *proxy_src = NULL;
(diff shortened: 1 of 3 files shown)


proto/http.c
```diff
@@ -301,14 +301,35 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
...
-			wsgi_req->len += proto_base_add_uwsgi_var(wsgi_req, "REMOTE_ADDR", 11, ip, strlen(ip));
-		}
-		else {
-			uwsgi_error("inet_ntop()");
-			return -1;
+		// TODO log something useful for AF_UNIX sockets
+		switch (http_sa->sa_family) {
+			case AF_INET6:
+				{
+					memset(ip, 0, INET6_ADDRSTRLEN+1);
```
eklitzke: you should use `sizeof(ip)` in the last field, it's less fragile if the size of `ip` changes in the future

proto/http.c
```diff
@@ -301,14 +301,35 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
...
+						uwsgi_error("inet_ntop()");
+						return -1;
+					}
+				}
+				break;
+			case AF_INET:
+			default:
+				{
+					struct sockaddr_in* http_sin = (struct sockaddr_in*)http_sa;
+					memset(ip, 0, INET_ADDRSTRLEN+1);
```
eklitzke: same comment applies here

eklitzke:
I looked at it and it looks OK to me (although I'm not qualified to merge your commit). My only comment is that since the structure size is known in advance you should be able to use `sizeof(ip)` which makes the code more resilient to changes in the structure size (and more readable IMO).

Roguelazer:
Incidentally, it would be awesome to get this in a hypothetical 2.0.13; let me know if you would like another pull request submitted against the `uwsgi-2.0` branch as well.

Roguelazer:
@xrmx Anything else you'd like changed in this review?

xrmx:
@Roguelazer yeah, we need to handle solaris and cygwin, @thwarted suggestion looks the way to go.

Roguelazer:
Given that the macro is in POSIX, it should just work as written on Solaris and Cygwin. Someone with access to those platforms will have to check, of course.

Andy's suggestion actually *won't* work because the macro definition he referenced is only valid on Linux -- not even on BSD.

-- 
James Brown, currently mobile

xrmx:
@Roguelazer cool, thanks for the clarification.

Roguelazer:
@xrmx I see that I just missed 2.0.13; any changes you'd like in order to get this into 2.0.14?

Roguelazer:
Is there anything I can do to move this forward? It's been rather a while...

proto/http.c
```diff
@@ -301,14 +301,61 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
...
+					if (IN6_IS_ADDR_V4MAPPED(&http_sin->sin6_addr)) {
+						/* just grab the last 4 bytes and pretend they're
+						 * IPv4. None of the word/half-word convenience
+						 * functions are in POSIX, so just stick to .s6_addr
+						 */
+						union {
+							unsigned char s6[4];
+							uint32_t s4;
+						} addr_parts;
+						addr_parts.s6[0] = http_sin->sin6_addr.s6_addr[12];
```
xrmx: memset maybe?

proto/http.c
```diff
@@ -301,14 +301,61 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
...
+							return -1;
+						}
+					} else {
+						if (inet_ntop(AF_INET6, (void *) &http_sin->sin6_addr, ip, INET6_ADDRSTRLEN)) {
+							wsgi_req->len += proto_base_add_uwsgi_var(wsgi_req, "REMOTE_ADDR", 11, ip, strlen(ip));
+						} else {
+							uwsgi_error("inet_ntop()");
+							return -1;
+						}
+					}
```
xrmx: there's a space before the bracket

proto/http.c
```diff
@@ -301,14 +301,61 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
...
-		memset(ip, 0, INET_ADDRSTRLEN+1);
-		if (inet_ntop(AF_INET, (void *) &http_sin->sin_addr.s_addr, ip, INET_ADDRSTRLEN)) {
-			wsgi_req->len += proto_base_add_uwsgi_var(wsgi_req, "REMOTE_ADDR", 11, ip, strlen(ip));
-		}
-		else {
-			uwsgi_error("inet_ntop()");
-			return -1;
+		// TODO log something useful for AF_UNIX sockets
+		switch (http_sa->sa_family) {
+			case AF_INET6:
```
xrmx: We can save two levels if you align cases to the switch and remove the brackets.

uwsgi.h
```diff
@@ -1356,8 +1356,11 @@ struct wsgi_request {
 	char *appid;
 	uint16_t appid_len;
 
-	//this is big enough to contain sockaddr_in
-	struct sockaddr_un c_addr;
+	union address {
```
xrmx: Changes at the end of struct please for ABI compatibility, so in this case move the union at the end and switch the c_addr callers to the new field i suppose.

Roguelazer: Just to confirm, you want me to leave the existing type-punned `c_addr` in place and just not use it? I want to make sure I make the right change.

Roguelazer:
I'm going to do one push to this branch to rebase it on master, then I'll add some more commits to fix your comments above.

proto/http.c
Roguelazer: Removing the brackets changes the scoping semantics (it means that I need to effectively hoist the variables defined in these sections to the top level). I can do that, but I'm not sure it's the desired goal.

As far as alignment, I was trying to match the existing `case/switch` usage in this file (e.g., `uwsgi_is_full_http` on line 750). If your preference is to align `case` with `switch`, I'll do so.

proto/http.c
```diff
@@ -329,7 +329,7 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
 	}
 	else {
 		// TODO log something useful for AF_UNIX sockets
-		switch (http_sa->sa_family) {
+		switch(http_sa->sa_family) {
 			case AF_INET6:
```
xrmx: i meant unindent this

proto/http.c
```diff
@@ -329,7 +329,7 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
...
 				{
```
xrmx: and the removal of this

proto/http.c
```diff
@@ -329,7 +329,7 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
...
 					memset(ip, 0, sizeof(ip));
```
xrmx: so that from there is two tabs less

xrmx:
Il 31/01/2017 19:21, James Brown ha scritto:
> *@Roguelazer* commented on this pull request.
>
> --------------------------------------------------------------------------------
>
> In proto/http.c <https://github.com/unbit/uwsgi/pull/1203>:
>
>> @@ -301,14 +301,61 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
>  		}
>  	}
>  	else {
> -		// TODO add ipv6 support
> -		memset(ip, 0, INET_ADDRSTRLEN+1);
> -		if (inet_ntop(AF_INET, (void *) &http_sin->sin_addr.s_addr, ip, INET_ADDRSTRLEN)) {
> -			wsgi_req->len += proto_base_add_uwsgi_var(wsgi_req, "REMOTE_ADDR", 11, ip, strlen(ip));
> -		}
> -		else {
> -			uwsgi_error("inet_ntop()");
> -			return -1;
> +		// TODO log something useful for AF_UNIX sockets
> +		switch (http_sa->sa_family) {
> +			case AF_INET6:
>
> Removing the brackets changes the scoping semantics (it means that I need to
> effectively hoist the variables defined in these sections to the top level). I
> can do that, but I'm not sure it's the desired goal.
>
> As far as alignment, I was trying to match the existing |case/switch| usage in
> this file (e.g., |uwsgi_is_full_http| on line 750). If your preference is to
> align |case| with |switch|, I'll do so.

Sorry i haven't seen this on github and reviewed again before reading your comments.

If you need the scope, then fine but please puts brackets after the case label 
on the same line or just on new line but same column. We are wasting a lot of 
horizontal space for no good reason.

xrmx:
Il 31/01/2017 19:17, James Brown ha scritto:
> *@Roguelazer* commented on this pull request.
>
> --------------------------------------------------------------------------------
>
> In uwsgi.h <https://github.com/unbit/uwsgi/pull/1203>:
>
>> @@ -1356,8 +1356,11 @@ struct wsgi_request {
>  	char *appid;
>  	uint16_t appid_len;
>
> -	//this is big enough to contain sockaddr_in
> -	struct sockaddr_un c_addr;
> +	union address {
>
> Just to confirm, you want me to leave the existing type-punned |c_addr| in place
> and just not use it? I want to make sure I make the right change.

Yeah we have to keep the size and aligment the same

unbit:
Sorry for the late :) Thanks to all

PR #1207, opened 2016-03-14
Added python classifiers for version 2 and 3.

### LINKED ISSUE #1206: Add classifiers for supported python versions to setup.py
It would be neat if you'd add classifiers to your setup.py so tools like caniusepython3.com can check if uwsgi is actually supported for python3.

jaapz:
Fixes #1206


xrmx:
Merged, thanks! Explicit is better than implicit i suppose :)

jaapz:
Cool, thanks!

PR #1217, opened 2016-03-25
Add variable plugin_build_dir to uwsgiconfig

jirutka:
Package managers for Alpine Linux (APK) and Gentoo (Portage) builds packages in different directory than where are eventually installed. Therefore we need to set different `plugin_dir` for build and runtime.

Gentoo ebuild for uWSGI solved this problem using simple workaround that involves patching `uwsgiconfig.py` (see [here](https://github.com/gentoo/gentoo/blob/master/www-servers/uwsgi/uwsgi-2.0.11.2-r1.ebuild#L149)). This patch do the same thing; it adds build variable `plugin_build_dir`, but with fallback to `plugin_dir` for backward compatibility.

/cc @ncopa, https://github.com/alpinelinux/aports/pull/38

### DIFF
--- a/uwsgiconfig.py
+++ b/uwsgiconfig.py
@@ -1424,7 +1424,7 @@ def build_plugin(path, uc, cflags, ldflags, libs, name=None):
         pass
 
     if uc:
-        plugin_dest = uc.get('plugin_dir') + '/' + name + '_plugin'
+        plugin_dest = uc.get('plugin_build_dir', uc.get('plugin_dir')) + '/' + name + '_plugin'
     else:
         plugin_dest = name + '_plugin'
 


xrmx:
The patch looks fine, any chance you can add a patch to uwsgi-docs to document the behaviour? Thanks

jirutka:
Thanks for quick response!

> …any chance you can add a patch to uwsgi-docs…

@xrmx I don’t know where to add it, the `plugin_dir` variable is not mentioned anywhere in the docs.

xrmx:
@jirutka http://uwsgi-docs.readthedocs.org/en/latest/BuildSystem.html may be the right place to mention it

Issue #1218, opened 2016-03-27
Distribute wheels to pypi


mattrobenolt:
Granted, we were always able to generate wheels for OSX and Windows, it's not possible to generate wheels for Linux as well.

This provides the benefits of a faster install and doesn't depend on compilers existing on the target machine to install/run uwsgi.

gevent was the first package to publish "manylinux" wheels: https://pypi.python.org/pypi/gevent and there's a project to facilitate generation: https://github.com/pypa/manylinux

I'm not personally super familiar with this, or what it'd take for uwsgi to support, but I'm willing to help. :)

xrmx:
Be our guest then! I don't think anyone will oppose to a pull request :) I see some issues though:
- centos 5 is not a good baseline for us since IIRC we use some kernel features not available there and we want to build with that features
- we need to find a configuration that covers most use cases
- we then need to sort out which configurations willl need a --no-binary switch when installing, unless we can still keep the source installation default of course.

mattrobenolt:
@xrmx GH-1219 is a start. It doesn't do anything fancy, but it fixes generation of platform specific whl files as a first step. So with this, in theory, it'd be possible now to fire up a bunch of virtualenvs to compile against different versions of python and generate all the wheels.

ushuz:
https://github.com/joerick/cibuildwheel can make building manylinux1 wheel on Travis CI really easy.

---

> centos 5 is not a good baseline for us since IIRC we use some kernel features not available there and we want to build with that features

Docker uses the kernel on host machine, instead of the kernel that's available for CentOS 5 distro. So it may not be a problem, but I'm not entirely sure about this, just a thought.

ipmb:
We've been working on getting wheels published to PyPI (including manylinux) under the `pyuwsgi` package. https://pypi.org/project/pyuwsgi/

There are currently some pre-releases up there if you'd like to test it out.

```
pip install --pre pyuwsgi
pyuwsgi --help
```

If you notice any problems with the packaging, please file an issue at https://github.com/lincolnloop/pyuwsgi-wheels. 👍

mattrobenolt:
@ipmb You're doing the Lord's work. 🙏

ipmb:
I think this one is safe to close now. pyuwsgi 2.0.18 has wheels on PyPI.

mattrobenolt:
But uwsgi itself is not. This was at least my original intent.

ipmb:
[...]

PR #1219, opened 2016-03-27
Compile properly platform tagged wheel

mattrobenolt:
Refs GH-1218

Prior to this, running `setup.py bdist_wheel` generated a completely untagged platform whl file.

Note: this was tested against OS X and does not solve Linux wheels yet.

With this, it's possible to now build wheels for different OSX platforms in virtualenvs of different python versions.

In my case, I ended up with a whl as such: `uWSGI-2.1_dev_c0c77cf-cp27-cp27m-macosx_10_11_x86_64.whl`, which means it only works with cpython2.7, on macosx 10.11.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,12 @@ from setuptools.command.install import install
 from setuptools.command.install_lib import install_lib
 from setuptools.command.build_ext import build_ext
 
+try:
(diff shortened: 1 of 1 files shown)


xrmx:
Looks good, thanks!

PR #1228, opened 2016-04-06
Complex improving plugin/rados

funny-falcon:
- librados have ability to read part of object from a middle. HTTP has 'Range: bytes=' header.
  Lets join these abilities.
- rados plugin used very small buffersizes. Given that requests with librados are not pipelined, it leads to
  extremely poor performance.
  Lets increase default buffer size and make it configurable.
- fix PUT for objects larger than buffer size and add ability to PUT zero sized objects
- fix PUT large objects into Erasure coded pools
- small fix for handling mount point. It seems that it were not working with mountpoint different from '/'.

### DIFF
--- a/uwsgi.h
+++ b/uwsgi.h
@@ -1522,9 +1522,9 @@ struct wsgi_request {
 	// when set, do not send warnings about bad behaviours
 	int post_warning;
 
-	enum uwsgi_range range_parsed;
-	int64_t range_from;
-	int64_t range_to;
+	// deprecated fields: size_t is 32bit on 32bit platform
+	size_t __range_from;
+	size_t __range_to;
 
 	// current socket mapped to request
 	struct uwsgi_socket *socket;
@@ -1618,6 +1618,12 @@ struct wsgi_request {
 
 	// uWSGI 2.1
 	uint64_t len;
+
+	// 64bit range, deprecates size_t __range_from, __range_to
+	enum uwsgi_range range_parsed;
+	int64_t range_from;
+	int64_t range_to;
+
 };
 
 
--- a/uwsgi.h
+++ b/uwsgi.h
@@ -1350,6 +1350,13 @@ struct uwsgi_transformation {
 	struct uwsgi_transformation *next;
 };
 
+enum uwsgi_range {
+	UWSGI_RANGE_NOT_PARSED,
+	UWSGI_RANGE_PARSED,
+	UWSGI_RANGE_VALID,
+	UWSGI_RANGE_INVALID,
+};
+
 struct wsgi_request {
 	int fd;
 	struct uwsgi_header *uh;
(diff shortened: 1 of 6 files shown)


funny-falcon:
Should I make pull request for documentation now, or if/when this pull request will be merged?

funny-falcon:
pushed with force to fix a leak.

uwsgi.h
```diff
@@ -1515,8 +1522,9 @@ struct wsgi_request {
 	// when set, do not send warnings about bad behaviours
 	int post_warning;
 
-	size_t range_from;
-	size_t range_to;
+	enum uwsgi_range range_parsed;
+	int64_t range_from;
```
unbit: please move new struct fields at the end of the struct to retain ABI compatibility

funny-falcon: what about changing type of `range_from` ?

funny-falcon: In other words: do we care that 32bit platform will not able to handle files more than 4GB if `range_from` and `range_to` remains to be `size_t` ?

unbit: the code is compiled with '-D_LARGEFILE_SOURCE', '-D_FILE_OFFSET_BITS=64' so size_t should be always 64bit

funny-falcon: No, `size_t` is only for arrays in memory, so it is always 32bit on 32bit platform.
`-D_FILE_OFFSET_BITS=64` will set `off_t` type as 64bit integer.

unbit: so what about setting both to off_t ? (as technically they are effectively offsets)

funny-falcon: On one side, I really care a little, cause our company doesn't use 32bit platform.

On the other side, this offsets are read from user input, and there is application/plugin decision how to treat them.
For example `plugin/rados` could be used to fetch part of a huge file, and `rados_read` has signature:
```
int rados_read(rados_ioctx_t io, const char * oid, char * buf, size_t len, uint64_t off)
```
So, it uses 64bit offsets, and there is no reason to limit user.

unbit: well i think nowadays being limited to 32bit filesize is not a good thing. Please rename range_from to __range_from and range_to to __range_to and add all the required (64bit) fields to the end of the structure. This will not break binary plugins and will work on 64bit numbers

funny-falcon: done

funny-falcon:
If you wish, I may unite commits `fix API` into `improve handling http byte range`

funny-falcon:
Couple of additional changes:
- fix PUT for objects larger than buffer size and add ability to PUT zero sized objects
- fix PUT large objects into Erasure coded pools

unbit:
Thanks a lot, feel free to add documentation on uwsgi-docs and your credits in the CONTRIBUTORS file and on top of the rados plugin sources

PR #1229, opened 2016-04-06
Extend stats-pusher to support more address types

worr:
The stats_pusher_socket and stats_pusher_statsd plugins did not
correctly support any other address types other than ipv4 addresses.

### DIFF
--- a/core/socket.c
+++ b/core/socket.c
@@ -1952,6 +1952,56 @@ int uwsgi_socket_passcred(int fd) {
 #endif
 }
 
+int uwsgi_socket_from_addr(union uwsgi_sockaddr *addr, socklen_t *len, char *addrtxt, int sock_type) {
+	int fd = -1;
+	int type;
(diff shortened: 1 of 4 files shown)


sixninetynine:
+1 to this change

unbit:
please refactor it to have the following code path:

* check for : (for tcp)
* undef #ifdef AF_INET6 check for square brackets
* else: fallback to unix socket

this is how other parts of the code work.

Why you disable non_blocking mode for unix sockets ?

worr:
I've updated the patch to do a few things.

1. Implement the logic for determining the type of address that you requested
2. Move this logic into `core/socket.c` since multiple plugins can make use of it
3. Used non-blocking mode for unix sockets
4. In cases where the socket buffer was too small, and sending would require blocking, drop messages. I'm not sure this is great, but ensures that that codepath doesn't block

Let me know what you think. Thanks!

unbit:
Thanks, really good patch

PR #1232, opened 2016-04-09
luap: meta tostring cleanup

Darvame:
- uwsgi_lua_tolstring() to run the lua_tolsting() first, and, in case of fail, try a metatable.__tostring() (fallback)
- try to use stack instead of malloc() first in some functions
- and protecting the headers parsing in the uwsgi_lua_request()


Darvame:
@unbit 
Sorry for bothering with that plugin
I think this one should be merged to fix _uwsgi.rpc()_ and tolstring() mess
no time right now to continue, but this patch looks fine and complited for me

unbit:
Sorry for the late, thanks

Issue #1235, opened 2016-04-17
Version number in C compiler string causing error


justsostephen:
On line 33 of `uwsgiconfig.py`, `os.environ.get('CC', sysconfig.get_config_var('CC'))` is used to detect the installed C compiler. On OS X 10.11.3 with Xcode 7.2.1 the `CC` environment variable isn't set by default, so `sysconfig.get_config_var('CC')` is used. On my system, this returns `gcc-4.2`, which is stored as the constant `GCC`.

Consequently, when `gcc_version = spcall("%s -dumpversion" % GCC)` runs on line 682, Bash reports that the command `gcc-4.2` cannot be found. Seemingly taking this into account, `llvm-` is prepended onto the `GCC` constant, resulting in the string `llvm-gcc-4.2`. On line 688, `gcc_version = spcall("%s -dumpversion" % GCC)` runs again using the modified value, however Bash once again reports that the command cannot be found.

Neither `gcc-4.2` or `llvm-gcc-4.2` are valid commands. Removing `-4.2` makes them both valid. Unless I'm missing something (and that could well be the case!), revising `uwsgiconfig.py` to remove the version number from the string returned by `sysconfig.get_config_var('CC')` would fix this issue.

xrmx:
thanks for the reports, i think the less painful way to handle this would be to special case os x and use clang as default compiler. All supported mac os x versions have clang as default compiler and gcc is just a symlink.

xrmx:
Does this work for you?
```
diff --git a/uwsgiconfig.py b/uwsgiconfig.py
index 617d6b3..d64a775 100644
--- a/uwsgiconfig.py
+++ b/uwsgiconfig.py
@@ -30,9 +30,12 @@ except ImportError:
 
 PY3 = sys.version_info[0] == 3
 
-GCC = os.environ.get('CC', sysconfig.get_config_var('CC'))
-if not GCC:
-    GCC = 'gcc'
+if uwsgi_os == 'Darwin':
+    GCC = os.environ.get('CC', 'clang')
+else:
+    GCC = os.environ.get('CC', sysconfig.get_config_var('CC'))
+    if not GCC:
+        GCC = 'gcc'
 
 def get_preprocessor():
     if 'clang' in GCC:
```

justsostephen:
Yes, that works! Does `sysconfig.get_config_var('CC')` return a string that doesn't contain a version number on other operating systems?

xrmx:
[...]

PR #1236, opened 2016-04-17
'uwsgiconfig.py' now removes detected C compiler version number.

justsostephen:
My proposed fix for issue #1235, where the detected C compiler string can contain a version number causing errors in subsequent version detection code.

### DIFF
--- a/uwsgiconfig.py
+++ b/uwsgiconfig.py
@@ -30,8 +30,10 @@ except ImportError:
 
 PY3 = sys.version_info[0] == 3
 
-GCC = os.environ.get('CC', sysconfig.get_config_var('CC'))
-if not GCC:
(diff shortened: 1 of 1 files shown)


xrmx:
This would break every other sane operating system though. I think the general consensus is to use CC=clang when building uwsgi on mac os x.

justsostephen:
Thanks for your feedback! Apologies, I don't currently have another OS available to test on. Doesn't removing the version number simply make the string equivalent to that returned by `os.environ.get('CC')`, assuming the `CC` environment variable is set?

xrmx:
If build with ``CC=gcc-6`` i want to really build with ``gcc-6`` and not ``gcc`` as your code will do. ``gcc`` may be a different version, not the one i want.

justsostephen:
I see, thanks for filling me in!

Issue #1233, opened 2016-04-11
polling fails on uwsgi_simple_wait_read_hook


pavel-stratoscale:
I got this error from the log:

1456612568.000000 Feb 27 22:36:08 uwsgi DEBUG    uwsgi_simple_wait_read_hook()/poll(): Interrupted system call [core/reader.c line 21] (/usr/bin/uwsgi:0)
1456612568.000000 Feb 27 22:36:08 uwsgi DEBUG    [uwsgi-body-read] Error reading 3276800 bytes. Content-Length: 16858218496 consumed: 1942880256 left: 14915338240 message: Interrupted system call (/usr/bin/uwsgi:0)
1456612568.000000 Feb 27 22:36:08 uwsgi DEBUG    1456612568.600(41080uWSGIWorker4Core0):ERROR:Failed to write to device: error during read(4194304) on wsgi.input

From looking at the code here: https://github.com/unbit/uwsgi/blob/master/core/reader.c#L20
I think that the code should check if the error is EINTR and retry polling and not fail the request

What do you think?

thatsafunnyname:
Thanks for uWSGI.
I was wondering why the change has not been merged into -2.0 branch and made available in a 2.0.X release?  Is it because it is a change in behaviour?  Maybe it could be an opt-in option for a 2.0.X or 2.X.Y release?
  I am asking because I wish to use the fix but would prefer to avoid:
  .) Using the master branch
  .) Using 3.X.X
  .) Maintaining a patch for the fix against 2.0.X releases
Is it just a case of needing to request it be included in the next release?  I noticed "last call" posts in the mailing list: http://lists.unbit.it/pipermail/uwsgi/2018-February/008860.html
Thanks.

thatsafunnyname:
Backported from master branch into -2.0 in https://github.com/unbit/uwsgi/commit/b5ece04d38e6c4ad5b0107cee779d3dbc5907097 , thanks.

PR #1238, opened 2016-04-18
BUG: When a process receives a signal, poll is interrupted and returns with errno=EINTR

pavel-stratoscale:
Adding a retry on EINTR


pavel-stratoscale:
related issue:
https://github.com/unbit/uwsgi/issues/1233

unbit:
Thanks a lot

PR #1245, opened 2016-05-04
Enable keepalive also for 1 second

eyal-stratoscale:
The current condition ignore http-keepalive == 1.
Since it is set in seconds, the timeout should be set also for a value of  1.

### DIFF
--- a/plugins/http/http.c
+++ b/plugins/http/http.c
@@ -1009,7 +1009,7 @@ ssize_t hr_instance_read(struct corerouter_peer *peer) {
 			hr->can_gzip = 0;
 			hr->has_gzip = 0;
 #endif
-			if (uhttp.keepalive > 1) {
+			if (uhttp.keepalive > 0) {
 				http_set_timeout(peer->session->main_peer, uhttp.keepalive);
 			}
 		}


fuxpavel:
+1

intgr:
Note that the argument to `http-keepalive` is optional. When using the command line switch `--http-keepalive` without a value, `uwsgi_opt_set_int` sets the option to 1, meaning simply "enabled". It actually used to be a boolean before 61ee03a99eeb6365e7d7eecc2b32d6e1ab646de2.

Not sure if it's worth breaking that? If yes, the argument should be marked as required. And the docs should be updated to describe what the value does (there's no word about it currently).

The current behavior is already confusing and surprising. I spent a few hours trying to figure out why `--http-keepalive` from command line worked, but `http-keepalive=true` in config didn't, because I assumed it was a boolean switch.

There are plenty of other people who got it wrong too: https://github.com/search?utf8=%E2%9C%93&q=http-keepalive&type=Code&ref=searchresults

karmux:
Merge this and #1637?

Current behaviour is super confusing and can take many hours of time to figure this out.

Issue #1248, opened 2016-05-10
stats_pusher_statsd uses incorrect types


thinkski:
Some metrics, such as `worker.n.requests`, are counters, meaning they are "generally-growing up number[s] (like the number of requests)." However, the term counter means something else in StatsD terminology. In StatsD, a counter is a bucket that is incremented by the reported value each time it is reported within a flush interval, then when it is flushed, the bucket is zeroed.

The default StatsD flush interval is every 10 seconds. The default uwsgi stats push frequency is every 3 seconds. So say that the number of requests per second is 100 are we are starting from 0. After 1 second, 100 requests have been processed, after 2 seconds 200 requests have been processed, after 3 seconds 300 requests have been processed. 300 is reported to StatsD. Then at 6 seconds 600 is reported to StatsD. Then at 9 seconds 900 is reported to StatsD. StatsD meanwhile, seeing this as a counter, aggregates 300 + 600 + 900 and reports 1800 to the backend. This is not only incorrect, but every 30 seconds, 4 datapoints will fit into the 10 seconds interval. This _cannot_ be resolved by matching the reporting frequency to the flush interval as it introduces a race -- sometimes 1 sample will be included in the flush interval, sometimes 2, and again, those 2 samples will sum.

The solution is to ignore the uwsgi metric type and use gauges for all metrics when reporting to StatsD.

PR #1249, opened 2016-05-10
Adds all-gauges option to stats_pusher_statsd

thinkski:
Adds option to statsd pusher plugin to report all metrics as gauges,
regardless of what uwsgi designates them as. This works around a
nomenclature difference where 'counters' in uwsgi are unlike 'counters'
in StatsD.

Fixes #1284

### DIFF
--- a/plugins/stats_pusher_statsd/plugin.c
+++ b/plugins/stats_pusher_statsd/plugin.c
@@ -18,10 +18,12 @@ extern struct uwsgi_server uwsgi;
 
 struct uwsgi_stats_pusher_statsd {
     int no_workers;
+    int all_gauges;
 } u_stats_pusher_statsd;
 
 static struct uwsgi_option stats_pusher_statsd_options[] = {
-	{"statsd-no-workers", no_argument, 0, "disable generation of single worker metrics", uwsgi_opt_true, &u_stats_pusher_statsd.no_workers, 0}
+	{"statsd-no-workers", no_argument, 0, "disable generation of single worker metrics", uwsgi_opt_true, &u_stats_pusher_statsd.no_workers, 0},
+	{"statsd-all-gauges", no_argument, 0, "push all metrics to statsd as gauges", uwsgi_opt_true, &u_stats_pusher_statsd.all_gauges, 0}
 };
 
 // configuration of a statsd node
@@ -92,7 +94,7 @@ static void stats_pusher_statsd(struct uwsgi_stats_pusher_instance *uspi, time_t
 		}
 		uwsgi_rlock(uwsgi.metrics_lock);
 		// ignore return value
-		if (um->type == UWSGI_METRIC_GAUGE) {
+		if (u_stats_pusher_statsd.all_gauges || um->type == UWSGI_METRIC_GAUGE) {
 			statsd_send_metric(ub, uspi, um->name, um->name_len, *um->value, "|g");
 		}
 		else {


unbit:
Thanks and sorry for the late

sinahmr:
I'm facing the problem you've mentioned here [#1248](https://github.com/unbit/uwsgi/issues/1248).
I've installed uwsgi with the command below:
`UWSGI_EMBED_PLUGINS=stats_pusher_statsd pip install --ignore-installed --no-cache-dir uwsgi`
And I'm running it like this:
`uwsgi ... --stats-push=statsd:IP:PORT,PREFIX`

Would you please help me fix the problem? @thinkski

thinkski:
@sinahmr Try with `--statsd-all-gauges`.

sinahmr:
Do you mean something like this?
`uwsgi ... --stats-push=statsd:IP:PORT,PREFIX --statsd-all-gauges`
It causes this error:
`uwsgi: unrecognized option '--statsd-all-gauges'
getopt_long() error`
@thinkski

thinkski:
@sinahmr Looks like there's a bug in the plugin -- the option structure is missing the `UWSGI_END_OF_OPTIONS` terminator. I suspect this only worked when the memory immediately following the struct happened to be zeroed. I suspect you are on OSX?

I have submitted #1579 to address the issue, although it will take a bit to propagate to PyPI. In the meantime, you can compile from source:

    UWSGI_EMBED_PLUGINS=stats_pusher_statsd python setup.py install

This should make the option available.

sinahmr:
I removed the old uwsgi and installed it via the second command you mentioned above.
And then run it via this command:
`uwsgi ... --stats-push=statsd:IP:PORT,PREFIX --statsd-all-gauges`
(I also changed the prefix, so I could find out if any data point is received by statsd.)
This time no error occurred, but the problem is, the configured prefix is not getting visible in Grafana dashboard, and it seems that no data point is set for that prefix. When I omit `--statsd-all-gauges` option, data points are received and shown in Grafana dashboad.
(I'm using Debian 8.8)
Thank you so much for your time. @thinkski

thinkski:
That is strange. I would recommend running a simple netcat client -- statsd is a simple text-based protocol:

    nc -u -l 8125

You should see your metrics showing up as gauges.

sinahmr:
I'm sorry, that was my fault, thanks for your help. I didn't know that gauges will be saved in stats.gauges.PREFIX and was searching somewhere else. So the `--statsd-all-gauges` option is now working.
Is it correct that even by using the option `--statsd-all-gauges`, buckets are not zeroed and the diagram (for number of requests) will still be ascending?
The picture below shows `worker.n.requests`. I thought by using that option, it would show number of requests in every 10-second bucket and after every 10 seconds, bucket would get empty.
image
I apologize this thread is getting long and I do thank you so much for helping.

thinkski:
Nope, the option doesn't affect zeroing, what you're seeing looks correct correct. If you're looking for a requests per second number, applying a derivative using the graphing backend (I used Librato which supported this) could be useful.

sinahmr:
I thank you so much for all answers. @thinkski

Jonnymcc:
When will a new release come out with this added feature?

thinkski:
@Jonnymcc Fair point. Only case against that which comes to mind is that existing users may somehow be working around it (e.g. expect graphs to look a certain way, or have set alerting thresholds to account for the erroneous values), which could cause alerts, say, to be missed. But it does seem like it would be generally better to fix the issue.

Jonnymcc:
Why does this not appear in version 2.0.17.1?

PR #1254, opened 2016-05-14
fixed #701

### LINKED ISSUE #701: gevent-monkey-patch happens too late
It seems to be happening directly after the wsgi entry point is loaded, however by this time a bunch of stuff has already been imported; in my specific case I am running a Pyramid app and pyramid.threadlocal, which imports threading.local,  has already been imported. The symptom for me is that I have some logging filters that use get_current_request(), which uses thread locals, to get some stuff from the request object like post data and urls. With a few concurrent requests these get mixed up.

I can fix it fairly simply in my case by calling gevent.monkey.patch_all() first thing in my wsgi.py file but if possible it would be really good for the --gevent-monkey-patch option to do this earlier on. At the least it would be worthwhile mentioning in the docs.

jianbin-wei:
cherry-picked --gevent-early-monkey-patch fix.


unbit:
Thanks and sorry for the late

PR #1255, opened 2016-05-15
Fix OpenBSD version check

pkorovin:
The original os version check does not work properly when obsd_major > 5 and obsd_minor = 0.
OpenBSD-current is at version 6.0 now, and uwsgi no longer builds.


xrmx:
Thanks!

PR #1260, opened 2016-05-23
plugin/rados: fix deprecation warning/error

funny-falcon:
- `rados_ioctx_pool_requires_alignment` is marked as deprecated in last
  versions of librados2. It causes warning which is converted to error.
- detect presence of `rados_ioctx_pool_requires_alignment2` and
  use it if present.


xrmx:
Looks good to me, thanks.

PR #1261, opened 2016-05-24
Fix pong timeout check

adrianbg:
last_pong may be much lower than last_ping from the previous pong

### DIFF
--- a/core/websockets.c
+++ b/core/websockets.c
@@ -62,7 +62,7 @@ static int uwsgi_websockets_check_pingpong(struct wsgi_request *wsgi_req) {
 	}
 	// pong not received ?
 	if (wsgi_req->websocket_last_pong < wsgi_req->websocket_last_ping) {
-		if (wsgi_req->websocket_last_ping - wsgi_req->websocket_last_pong > uwsgi.websockets_pong_tolerance) {
+		if (now - wsgi_req->websocket_last_ping > uwsgi.websockets_pong_tolerance) {
                                 uwsgi_log("[uwsgi-websocket] \"%.*s %.*s\" (%.*s) no PONG received in %d seconds !!!\n", REQ_DATA, uwsgi.websockets_pong_tolerance);
 				return -1;
 		}


xiejieyi-bear:
@unbit 
this commit fix a bug.
can you merge it?

when I use these code to receive msg
`message=True
while message:
    message = uwsgi.websocket_recv_nb()
`
has a Bug:
(1)uwsgi send ping to browser,
(2)at the same time uwsgi receive a bussiness message 
(3)'while true' loop will read until buf is empty
(4)the expression  'wsgi_req->websocket_last_ping - wsgi_req->websocket_last_pong ' will equal 30
it bigger than 3，will close the connection
(5)I modify the code ,delete 'return -1' ,uwsgi will receive pong  in next while loop.

image

ajs124:
Same here, this (seems to) fix a bug for me where websockets were regularly disconnecting.

I haven't done extensive testing yet, but with this merged it seems at least *much better* than before, if not completely fixed.

maxim-s-barabash:
Is there a solution to this problem?

ajs124:
Yes, this *is* the solution, not the problem. The *problem* is that whoever maintains this project doesn't seem to care about this issue.

There hasn't even been a comment by @unbit on this PR…

rdeioris:
@ajs124 thanks for your patch but there is no need to be harsch. Unfortunately not having automated tests makes merging way more complex/slow and frankly speaking i completely overlooked this patch. I always ask in the mailing-list if there is some pr to include before each release, but this one never popped up. I will check it in the next few days.

adrianbg:
@rdeioris it's my PR, which I wrote almost two years ago. I had no idea that there was a mailing list or that it would be the correct place to mention a PR. @unbit was also tagged almost a year ago.

rdeioris:
@adrianbg no problem. it is in the queue now :) thanks

unbit:
@xrmx safe for backport

unbit:
Thanks to all for the endless patience :)

adrianbg:
Haha np. Thanks for merging

amks1:
@unbit 
I've been seeing a lot of these messages in my uwsgi logs in the latest version (2.0.20). 

```
[uwsgi-websocket] "GET /socket.io/?EIO=4&transport=websocket&sid=CMefXSdp9F6AR3X-AAEU" (172.28.0.7) no PONG received in 3 seconds !!!
[uwsgi-websocket] "GET /socket.io/?EIO=4&transport=websocket&sid=MXiq2W8p38SN6g1DAAEY" (172.28.0.7) no PONG received in 3 seconds !!!
[uwsgi-websocket] "GET /socket.io/?EIO=4&transport=websocket&sid=D4c8Vtpjqd-i0TUKAAEg" (172.28.0.7) no PONG received in 3 seconds !!!
```
etc...

When a lot of these messages show up, CPU goes up to 100% and stays there till I restart the server (although I am not a 100% sure these are connected).
Any idea what this could be due to, and how to bugfix it?

Issue #1289, opened 2016-06-16
Worker that handles a signal is marked as non-busy


shoham-stratoscale:
When checking if a worker is busy, there's a check if it's currently handling a signal.
But the check is wrong and checks "mywid" instead of the worker id in question

https://github.com/unbit/uwsgi/blob/fb3628d7295772615004a03cdbc67de702b81c2a/core/master_checks.c#L369

shoham-stratoscale:
Fix was merged

PR #1290, opened 2016-06-16
If worker is handling signal it's busy

shoham-stratoscale:
When checking if worker is busy, check that it's not handling a signal.

Until now, the function checked that mywid is not handling
signal instead of the worker in question.


shoham-stratoscale:
Solves #1289

unbit:
Nice catch, @xrmx feel free to backport to stable

Issue #1288, opened 2016-06-16
When cheaping, master worker may cheap busy workers


shoham-stratoscale:
When cheaper algorithm says it's ok to cheap workers, the master kills the oldest worker. When doing so, it doesn't check if the worker is busy, and it may kill a worker while it's handling a request.

https://github.com/unbit/uwsgi/blob/fb3628d7295772615004a03cdbc67de702b81c2a/core/master_utils.c#L193

shoham-stratoscale:
Fix was merged

abarghou:
Hi,
Is this fix released? If not, is there any estimate for when it would be?
Thanks

unbit:
They are already merged in uwsgi-2.0, next stable release should be out during the week

PR #1291, opened 2016-06-16
Only kill worker if it's not busy

shoham-stratoscale:
We had a bug where in busyness, if the load is too low, uwsgi would kill
a worker, without checking if it's busy - simply kills the oldest worker.

I've added a check that the worker is not busy handling
a request before killing it.

Solves #1288


unbit:
Thanks again, @xrmx this is good for backport too

PR #1297, opened 2016-06-21
add pex builder for uwsgi

eddwardo:
It is not straightforward how to integrate pex (distribution library) with uwsgi. I'm creating this PR but it needs a review from someone with more experience.
Usage:

* build uwsgi with: ` python uwsgiconfig.py --build pex`
* Build your pex file `pex -r requirements.txt . -o app.pex`
* Run uwsgi with `./pex_uwsgi --env UWSGI_PEX=/tmp/app.pex --module 'app.wsgi'`

- [ ] Where to put pexbootstrap.py?
- [ ] How to fail if build happens on non-linux os? Running pex on MacOS will fail silently.

### DIFF
--- a/buildconf/pex.ini
+++ b/buildconf/pex.ini
@@ -1,0 +1,6 @@
+[uwsgi]
+main_plugin = python
+inherit = base
+bin_name = pex_uwsgi
+embed_files = pexbootstrap.py
+embed_config = uwsgipex.ini
--- a/pexbootstrap.py
+++ b/pexbootstrap.py
@@ -1,0 +1,23 @@
+import os
+import sys
+import pex.pex_bootstrapper
+
+def activate_pex():
+    entry_point = os.environ.get('UWSGI_PEX')
+    if not entry_point:
+        sys.stderr.write('couldnt determine pex from UWSGI_PEX environment variable, bailing!\n')
+        sys.exit(1)
+
+    sys.stderr.write('entry_point=%s\n' % entry_point)
+
+    sys.path[0] = os.path.abspath(sys.path[0])
+    sys.path.insert(0, entry_point)
+    sys.path.insert(0, os.path.abspath(os.path.join(entry_point, '.bootstrap')))
+
+    pex.pex_bootstrapper.bootstrap_pex_env(entry_point)
+
+    sys.stderr.write('sys.path=%s\n\n' % sys.path)
+
+    return entry_point
+
+activate_pex()
--- a/uwsgipex.ini
+++ b/uwsgipex.ini
@@ -1,0 +1,2 @@
+[uwsgi]
+import = sym://pexbootstrap_py


xrmx:
Haven't heard of pex since now:
- what is pexbootstrap.py?
- what will fail silently on mac os x?

eddwardo:
Pex files are single-file python virtual environments (dependencies + developers code) - https://pex.readthedocs.io/en/stable/whatispex.html#tl-dr 

In order to run them - pex needs to find the package (*.pex) and load it to memory. Usually pex files are executables with defined entry point, but in this case uwsgi doesn't execute the package but loads it to memory (`UWSGI_PEX` env variable).

eddwardo:
Maybe @kwlzn will provide a better insight.

xrmx:
@eddwardo Thanks for the explanation!

kwlzn:
yup - to put it another way, a `pex` is like a self-contained, zipped virtualenv with a default executable entrypoint. we use them at Twitter to simplify deployment of "hermetically sealed" python applications.

you can "bootstrap" a pex in a runtime path (as seen in the bootstrapping example here) to avoid executing its normal entrypoint and gain full access to imports of all of the modules within the pex including transitive thirdparty dependencies. so the idea of pex + uwsgi would be that uwsgi could bootstrap its import environment against a given pex to essentially load up the equivalent of a virtualenv-sourced-environment and then run any of the WSGI application entrypoints embedded in the pex.

however, it seems to me like first class support for pex in uWSGI probably shouldn't depend on environment variable passing - but rather something akin to the `--virtualenv` flag. then you'd handle the pex bootstrapping in uWSGI itself similar to how the virtualenv support is implemented rather than injecting an embedded bootstrapper in the build.

sixninetynine:
Hey @eddwardo @kwlzn @xrmx 

I've actually found a much more natural way of running uwsgi inside of pex is by leveraging the UWSGI_AS_LIB environment variable at build time.

Simply export that environment variable pointing to some location in your project, say:

`UWSGI_AS_LIB=src/project_name/submodule/runner/uwsgi.so`

Then, in the submodule, add this snippet:

```
import sys
import ctypes


def uwsgi_run(uwsgi_args):
    # load the uwsgi library in the global namespace
    uwsgi = ctypes.CDLL('./libuwsgi.so',mode=ctypes.RTLD_GLOBAL)

    uwsgi_args.insert(0, sys.argv[0])
    uwsgi_args.insert(1, '--binary-path')
    uwsgi_args.insert(2, sys.argv[0])

    # build command line args
    argv = (ctypes.c_char_p * (len(uwsgi_args)+1))()
    for pos,arg in enumerate(uwsgi_args):
        argv[pos] = arg
    # inform the uwsgi engine, the passed environ is not safe to overwrite
    envs = (ctypes.c_char_p * 1)()
    # enter into uWSGI !!!
    uwsgi.uwsgi_init(len(uwsgi_args), argv, envs)
```

Then invoke it as the entry point to your app:

```
from 
def run_server(args):
    uwsgi_run(args)
```

This bootstraps uwsgi's runtime after initializing the python runtime inside of pex, so your environment is ready to use and available to your wsgi application. It's very clean and self contained in a single pex file. I've used this paradigm in production for a few months now with no issues, whereas bootstrapping into a pex running under uwsgi has had some problems (such as segfaults when trying to load some plugins, etc).

-- Loren

sixninetynine:
just want to also point out - my example above solves the OSX compatibility problem as well

PR #1300, opened 2016-06-24
python: add uwsgi.accepting() for chain-reload + worker-override combo

enkore:
Previously it was not possible to use `touch-chain-reload` and `python-worker-override` at the same time since the worker-override script was not able to set the necessary `accepting` flag on the worker structure.

This patch adds a small `uwsgi` module API, `accepting([bool])` that does this.


unbit:
Cool use of the double exclamation mark :)

Would you like to post a snippet on the uwsgi-docs repos about the use of your function ?

enkore:
Thanks, sure! :)

PR #1304, opened 2016-06-29
Update logging.c

Nikolo:
Add pid to uwsgi_log\* message

### DIFF
--- a/core/logging.c
+++ b/core/logging.c
@@ -53,7 +53,10 @@ void uwsgi_log(const char *fmt, ...) {
 			rlen = 24 + 3;
 		}
 	}
-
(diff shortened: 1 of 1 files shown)


unbit:
Sorry, we do not change behaviours unless strictly required (and only after a long deprecation phase).

What about extending the variables supported by log encoders ?

http://uwsgi-docs.readthedocs.io/en/latest/LogEncoders.html

a ${pid} could be absolutely useful

unbit:
Just to be more clear, i mean this: http://uwsgi-docs.readthedocs.io/en/latest/LogEncoders.html#core-encoders

PR #1308, opened 2016-07-01
plugin/cgi: added cgi-safe option

nnnn20430:
Check if file is under safe path, useful when using symlinks.


unbit:
Thanks. @xrmx good for backporting

xrmx:
Backported, thanks!

Issue #1309, opened 2016-07-01
readline support not working under python3 (with --honour-stdin)


asottile:
Here's my sample setup:

## test.wsgi
```python
import pdb; pdb.set_trace()
```

## Set up a virtualenv
```
virtualenv venv34 -ppython3.4  # I've also tried python3.5
venv34/bin/pip install uwsgi  # At the time of writing 2.0.13.1
virtualenv venv27 -ppython2.7
venv27/bin/pip install uwsgi
```

## Under python2.7

I type `a`, press `[enter]`, press `[up arrow]` (a couple times).  You'll notice that readline is functioning (it gives me the `a` that I previously typed).

```
$ venv27/bin/uwsgi --wsgi-file test.wsgi --http ':9001' --honour-stdin
*** Starting uWSGI 2.0.13.1 (64bit) on [Fri Jul  1 13:30:36 2016] ***
compiled with version: 4.8.4 on 01 July 2016 13:10:01
os: Linux-3.16.0-76-generic #98~14.04.1-Ubuntu SMP Fri Jun 24 17:04:54 UTC 2016
nodename: work
machine: x86_64
clock source: unix
pcre jit disabled
detected number of CPU cores: 6
current working directory: /tmp
detected binary path: /tmp/venv/bin/uwsgi
*** WARNING: you are running uWSGI without its master process manager ***
your processes number limit is 31843
your memory page size is 4096 bytes
detected max file descriptor number: 1024
lock engine: pthread robust mutexes
thunder lock: disabled (you can enable it with --thunder-lock)
uWSGI http bound on :9001 fd 4
spawned uWSGI http 1 (pid: 5512)
uwsgi socket 0 bound to TCP address 127.0.0.1:41406 (port auto-assigned) fd 3
Python version: 2.7.6 (default, Jun 22 2015, 18:01:27)  [GCC 4.8.2]
*** Python threads support is disabled. You can enable it with --enable-threads ***
Python main interpreter initialized at 0xcd25d0
your server socket listen backlog is limited to 100 connections
your mercy for graceful operations on workers is 60 seconds
mapped 72768 bytes (71 KB) for 1 cores
*** Operational MODE: single process ***
--Return--
> /tmp/test.wsgi(1)<module>()->None
-> import pdb; pdb.set_trace()
(Pdb) a
(Pdb) a
```

## Under python3

Same procedure as above, however readline is not working!

```
[...]

PR #1310, opened 2016-07-01
Fix readline support in python3 when working interactively

asottile:
See #1309 for description.

This is probably not a correct solution, but figured I'd open this for discussion.

### DIFF
--- a/plugins/python/python_plugin.c
+++ b/plugins/python/python_plugin.c
@@ -556,16 +556,24 @@ void init_uwsgi_vars() {
 	pysys_dict = PyModule_GetDict(pysys);
 
 #ifdef PYTHREE
-	// fix stdout and stderr
+	// In python3, stdout / stderr was changed to be buffered (a bug according
+	// to many):
+	// - https://bugs.python.org/issue13597
+	// - https://bugs.python.org/issue13601
+	// We'll fix this by manually restore the unbuffered behaviour.
+	// In the case of a tty, this fix breaks readline support in interactive
+	// debuggers so we'll only do this in the non-tty case.
+	if (!Py_FdIsInteractive(stdin, NULL)) {
 #ifdef HAS_NO_ERRORS_IN_PyFile_FromFd
-	PyObject *new_stdprint = PyFile_FromFd(2, NULL, "w", _IOLBF, NULL, NULL, 0);
+		PyObject *new_stdprint = PyFile_FromFd(2, NULL, "w", _IOLBF, NULL, NULL, 0);
 #else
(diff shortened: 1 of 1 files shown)


unbit:
Unfortunately this break logging in python3

asottile:
Would it be acceptable to disable this when stdin is a tty?

Could you elaborate on "breaks logging"? I have a feeling you're referring to the (bug?) in python3 that stderr is buffered.

asottile:
I've clarified the comment and updated my commit to only run the reassignment when it is a non-tty.  This fixes my usecase and seems to be the least invasive -- let me know what you think :D

asottile:
ping: 1 week, this is blocking some development internally (we're considering a private fork)

rdeioris:
Unfortunately every decision we make in this topic will end in mess for someone else (and honestly i agree that the current python3 behaviour is more near to a bug than a feature). I continue to see funny uses/abuses of stdout/stderr so i would be more inclined to add a python3 plugin specific option to allow the user/admin to choose the behaviour that fit best to him/her.  It is a fast thing to do (you already covered 90% of the problem), but if you are in super-hurry i strongly suggest the 'private fork way'. @xrmx any opinion on the topic ?

xrmx:
Adding this as a specific option for the python plugin looks good to me

asottile:
To be honest, I was pretty surprised that uwsgi attempts to "solve" this.  As bad as this "feature" of python3 is, it should really be up to the application to decide whether it wants that behaviour or not.  And uwsgi's monkeypatching of `sys` to change the streams (and put stdout to stderr?) was certainly surprising!  If `uwsgi`'s python plugin aims to be a simple wrapper around the python interpreter, I certainly expect it to act the same as `python` :)  (which is why I initially suggested this patch: https://github.com/unbit/uwsgi/issues/1309#issuecomment-230048415)

unbit:
Unfortunately the python3 choice, broke advanced loggers (syslog mainly), this is why that part is here. I think adding a dedicated option for python3 is the best thing to do. Expect an official patch (with your credits) soon

asottile:
Any timeline on when this or something like this will reach mainline?

unbit:
By the end of this week :) thanks again

unbit:
@xrmx safe for backporting too.

@asottile After a second review of your patch it looks like it does not need a specific option to enable it, as your proposed approach is the saner one. Thanks a lot, and sorry for the late

asottile:
\o/ What's uwsgi's release cycle like (wondering how long we want to plan to maintain a fork)

PR #1314, opened 2016-07-12
plugins/coroae: graceful reload

aleksey-mashanov:
* add ability to wait for all coroutines to finish their jobs before exiting worker
* fix segfault on stop introduced by changes in plugins/psgi in 2.0.9
* fix memory leaks caused by incorrect usage of perl's refcounting and mortality

Graceful reload was ported from `gevent` plugin, so logic is the same.
This patch (backported to 2.0, see `uwsgi-2.0-coroae-graceful` branch) have been deployed on highload production cluster for a couple of weeks. Memory leakage for our code decreased from 1400MB/day to 100MB/day.
No problems were detected except for absence of `worker_reload_mercy` in some cases, but I decided to leave it as is to keep its behavior compatible with `gevent`.

### DIFF
--- a/plugins/coroae/coroae.c
+++ b/plugins/coroae/coroae.c
@@ -1,8 +1,15 @@
 #include "../psgi/psgi.h"
 #include "CoroAPI.h"
 
+struct uwsgi_coroae {
+	SV *condvar;
+	AV *watchers;
+	int destroy;
+};
+
 extern struct uwsgi_server uwsgi;
 extern struct uwsgi_perl uperl;
+struct uwsgi_coroae ucoroae;
 
 MGVTBL uwsgi_coroae_vtbl = { 0,  0,  0,  0, 0 };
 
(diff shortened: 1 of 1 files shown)


xrmx:
applied with this on top:
```
diff --git a/plugins/coroae/coroae.c b/plugins/coroae/coroae.c
index 8255b9b..ec66253 100644
--- a/plugins/coroae/coroae.c
+++ b/plugins/coroae/coroae.c
@@ -457,11 +457,11 @@ XS(XS_coroae_int_sighandler) {
        coroae_int();
 }
 
-static void coroae_gbcw(const char *reason) {
+static void coroae_gbcw(void) {
        if (ucoroae.destroy) return;
        ucoroae.destroy = 1;
 
-       uwsgi_log("...The work of process %d is done (reason: %s). Seeya!\n", getpid(), (reason != NULL ? reason : "no reason given"));
+       uwsgi_log("...The work of process %d is done. Seeya!\n", getpid());
 
        uwsgi_time_bomb(uwsgi.worker_reload_mercy, 0);
```

Issue #1316, opened 2016-07-19
per-request time in stats


bplotnick:
Currently, I don't believe there is a way to know how long a request has been running without harakiri (which may be a bit too heavy-handed for some). It may be useful to know how long the _current_request_ has been running for (I think avg_rt is close, but not good enough for in-flight requests). This could be done as simply as a per-core timestamp for when the most recent request came through.

For reference, Apache has this statistic in mod_status, called "SS"

PR #1317, opened 2016-07-20
add wsgi_request info to stats output

bplotnick:
As I mentioned in #1316, we don't have much in the way of per-request information in the stats server. Some of this information is available in the logs and some might be available in the "vars" field, but it is helpful to get this from stats (as in the apache server_status).

### DIFF
--- a/core/master_utils.c
+++ b/core/master_utils.c
@@ -1272,6 +1272,20 @@ struct uwsgi_stats *uwsgi_master_generate_stats() {
 			if (uwsgi_stats_list_close(us))
 				goto end;
 
+			if (uwsgi_stats_comma(us)) goto end;
+
+			if (uwsgi_stats_key(us, "req_info"))
+				goto end;
+
+			uwsgi_stats_symbol(us, '\n');
+
+			if (uwsgi_stats_object_open(us))
+				goto end;
+
+			if (uwsgi_stats_dump_request(us, uc)) goto end;
+
+			if (uwsgi_stats_object_close(us))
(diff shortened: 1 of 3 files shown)


unbit:
Hi, thanks for the patch, the "request_start" info is good, the other fields are not required as you already can get them from the "vars" list. Would you like to re-push the patch with only the request_time field ?

bplotnick:
That's a good point! They [must](https://www.python.org/dev/peps/pep-0333/#environ-variables) be "vars" if they are not null.  I've updated the branch to remove the excess. Thanks!

drolando:
@unbit Hey, do you have any more comments on this patch?
We'd like to get this merged soon since it's blocking our migration to uWSGI.

core/master_utils.c
```diff
@@ -1272,6 +1272,20 @@ struct uwsgi_stats *uwsgi_master_generate_stats() {
 			if (uwsgi_stats_list_close(us))
 				goto end;
 
+			if (uwsgi_stats_comma(us)) goto end;
+
+			if (uwsgi_stats_key(us, "req_info"))
+				goto end;
+
+			uwsgi_stats_symbol(us, '\n');
```
unbit: with this newline ?

bplotnick: Without the newline you get this:

```
"req_info":					{
```

with it you get this:

```
"req_info":
{
```

unbit:
@xrmx it is safe for backport, but take in account i have removed the newline to allow minification

PR #1324, opened 2016-07-24
router: add fcgiauth action

ftarnell:
```
This adds 'fcgiauth' as a router action, to perform authentication
against a FastCGI authorizer.  Usage:

	uwsgi --http :8080 --route-run fcgiauth:127.0.0.1:8000 --route-run return:200

As per the FastCGI specification, we send the request headers to the
authorizer; if it returns 200, we ignore the authorizer resonse and
process the request as normal, otherwise we send the authorizer response
(probably a 401/403) back to the client.
```

This is a very rough implementation; it works, but it's not ready for merging yet.  I wanted to open the PR to make sure this looks like a reasonable way to implement it.

The main missing feature is that `Variable-FOO: bar` headers from the responder should add `FOO: bar` headers to the request; this is required for us to support Shibboleth (the reason I wrote this in the first place).  However, basic username/password auth should be working:

```
91!klamath:~/tbx/uwsgi>curl -H'Authorization: test' -si http://localhost:8080/
HTTP/1.1 200 OK
Content-Type: text/plain
Content-Length: 2

OK
92!klamath:~/tbx/uwsgi>curl -si http://localhost:8080/
HTTP/1.1 401 Unauthorized
WWW-Authenticate: basic realm="foo"

93!klamath:~/tbx/uwsgi>
```

I've attached [a trivial Perl authorizer implementation for testing](https://github.com/unbit/uwsgi/files/380394/auth.pl.txt)

### DIFF
--- a/plugins/router_fcgi/router_fcgi.c
+++ b/plugins/router_fcgi/router_fcgi.c
@@ -1,0 +1,268 @@
+#include <uwsgi.h>
+
+#ifdef UWSGI_ROUTING
+
+extern struct uwsgi_server uwsgi;
+
+struct fcgi_record {
+	unsigned char version;
+	unsigned char type;
+	unsigned char requestIdB1;
+	unsigned char requestIdB0;
+	unsigned char contentLengthB1;
+	unsigned char contentLengthB0;
+	unsigned char paddingLength;
+	unsigned char reserved;
+	char *data;
+} FCGI_Record;
+
+static int fcgi_send(struct wsgi_request *wsgi_req, char *addr, struct uwsgi_buffer *ub, int timeout) {
+	int fd = uwsgi_connect(addr, 0, 1);
+	if (fd < 0)
+		return -1;
+
+	int ret = uwsgi.wait_write_hook(fd, timeout);
+	if (ret <= 0)
+		goto error;
+
+	if (uwsgi_write_true_nb(fd, ub->buf, ub->pos, timeout))
+		goto error;
+
+	return fd;
+
+error:
+	close(fd);
+	return -1;
+}
+
+static int uwsgi_routing_func_fcgi(struct wsgi_request *wsgi_req, struct uwsgi_route *ur) {
+	struct uwsgi_buffer *ub = NULL, *headers = NULL;
+	int ret = UWSGI_ROUTE_BREAK;
+	int inbody = 0;
+
+	// mark a route request
+        wsgi_req->via = UWSGI_VIA_ROUTE;
+
+	char **subject = (char **) (((char *)(wsgi_req))+ur->subject);
+        uint16_t *subject_len = (uint16_t *)  (((char *)(wsgi_req))+ur->subject_len);
+
+	struct uwsgi_buffer *ub_addr = uwsgi_routing_translate(wsgi_req, ur, *subject, *subject_len, ur->data, ur->data_len);
+	if (!ub_addr) return UWSGI_ROUTE_BREAK;
+
+	// convert the wsgi_request to an fcgi request
+	ub = uwsgi_to_fastcgi(wsgi_req, ur->custom ? FCGI_AUTHORIZER : FCGI_RESPONDER);
+
+	if (!ub) {
+		uwsgi_log("unable to generate fcgi request for %s\n", ub_addr->buf);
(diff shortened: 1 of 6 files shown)


ftarnell:
Okay, now both `fcgi:` and `fcgiauth:` are supported; the former does a normal fcgi request, without the authorizer's special handling for a 200 response.

I've tested the streaming buffer code with `buf` sizes of 999, 1023, 1024, 1025, 8191, 8192, 8193 and 65535, so I'm fairly certain it's correct now.

xrmx:
@ftarnell thanks for the patches! could you please squash all your commits into one when you are done with the changes please? Also what about adding the test script to the pull request and place it somewhere in t/routing with a comment on how to run it?

plugins/router_fcgi/router_fcgi.c
```diff
@@ -72,39 +71,89 @@ static int uwsgi_routing_func_fcgiauth(struct wsgi_request *wsgi_req, struct uws
...
+			rptr = NULL;
+		} else { 
+			if ((sizeof(buf) - (ptr - buf) - left) > 0)
+				rlen = read(fd, ptr + left, sizeof(buf) - (ptr - buf) - left);
+		}
+#endif
+		/* Amount left in buffer is not a full record header, so we
+		 * need to fudge the next read to append to the current buffer. */
+		if ((sizeof(buf) - (ptr - buf) - left) < 8) {
+//printf("fudging\n");
```
xrmx: Do we really need these commented prints?

plugins/router_fcgi/router_fcgi.c
```diff
@@ -72,39 +71,89 @@ static int uwsgi_routing_func_fcgiauth(struct wsgi_request *wsgi_req, struct uws
...
 
 	for (;;) {
                 int r = uwsgi_waitfd(fd, uwsgi.socket_timeout);
                 if (r <= 0) goto end;
 
-		size_t rlen = read(fd, ptr, sizeof(buf) - (ptr - buf));
-		if (rlen <= 0)
-			break;
+		ssize_t rlen = 0;
+#if 0
```
xrmx: Need it or not? :)

ftarnell:
Okay, rebased :)

And I've removed the debug code and `#if 0`.  (I've also realised that `POST` bodies aren't implemented yet, I'll add that today.)

plugins/router_fcgi/router_fcgi.c
```diff
@@ -0,0 +1,268 @@
...
+static int uwsgi_routing_func_fcgi(struct wsgi_request *wsgi_req, struct uwsgi_route *ur) {
+	struct uwsgi_buffer *ub = NULL, *headers = NULL;
+	int ret = UWSGI_ROUTE_BREAK;
+	int inbody = 0;
+
+	// mark a route request
+        wsgi_req->via = UWSGI_VIA_ROUTE;
+
+	char **subject = (char **) (((char *)(wsgi_req))+ur->subject);
+        uint16_t *subject_len = (uint16_t *)  (((char *)(wsgi_req))+ur->subject_len);
```
xrmx: spaces

plugins/router_fcgi/router_fcgi.c
```diff
@@ -0,0 +1,268 @@
...
+	struct uwsgi_buffer *ub_addr = uwsgi_routing_translate(wsgi_req, ur, *subject, *subject_len, ur->data, ur->data_len);
+	if (!ub_addr) return UWSGI_ROUTE_BREAK;
+
+	// convert the wsgi_request to an fcgi request
+	ub = uwsgi_to_fastcgi(wsgi_req, ur->custom ? FCGI_AUTHORIZER : FCGI_RESPONDER);
+
+	if (!ub) {
+		uwsgi_log("unable to generate fcgi request for %s\n", ub_addr->buf);
+		uwsgi_buffer_destroy(ub_addr);
+                return UWSGI_ROUTE_BREAK;
```
xrmx: spaces too

unbit:
@xrmx i am about to re-run an indent round on the whole source tree so i will fix spacing issues by myself. @ftarnell thanks a lot, would you like to make a bunch of docs for this plugin ? (make pull requests for the uwsgi-docs repository)

Issue #915, opened 2015-05-10
Mules are not being restarted on touch-chain-reload 


offline:

ramnes:
I can confirm this.

We were hit by that issue in production yesterday, and sadly discovered that touch-chain-reload doesn't reload mules nor spoolers.

This is a serious flaw in the touch-chain-reload feature, which is an awesome feature but that we can't use anymore in our current use-case.

Ping @ultrabug @Shir0kamii

unbit:
Please make a pull-request but ensure it is governed by an option, as we do not change established behaviours

ultrabug:
Sorry @unbit but are you saying this is the intended behaviour ?

unbit:
Yes, chain reloading is intended only for workers, but i have nothing against adding the recycling of mules and spoolers in the same action

PR #1327, opened 2016-07-26
Reload mules and spoolers

Shir0kamii:
Hello,

This PR relates to issue #915.

We didn't really need the mules and spooler to be chain reloaded, so I thought a general option would be better for other users of uwsgi.

I added two new options to enable the reloading of the mules and the spoolers upon the modification of a file. It's independent from the reloading of the workers.

What do you think ? Is it good for you ?

### DIFF
--- a/core/master.c
+++ b/core/master.c
@@ -181,6 +181,15 @@ void uwsgi_master_check_mercy() {
 			}
 		}
 	}
+
+
+	struct uwsgi_spooler *us;
+	for (us = uwsgi.spoolers; us; us = us->next) {
+		if (us->pid > 0 && us->cursed_at && uwsgi_now() > us->no_mercy_at) {
+				uwsgi_log_verbose("spooler %d (pid: %d) is taking too much time to die...NO MERCY !!!\n", i + 1, us->pid);
+				kill(us->pid, SIGKILL);
+		}
+	}
 }
 
 
@@ -867,6 +876,12 @@ int master_loop(char **argv, char **environ) {
 					uwsgi_reload_mules();
 					continue;
 				}
+				touched = uwsgi_check_touches(uwsgi.touch_spoolers_reload);
+				if (touched) {
+					uwsgi_log_verbose("*** %s has been touched... spoolers reload !!! ***\n", touched);
+					uwsgi_reload_spoolers();
+					continue;
+				}
 				touched = uwsgi_check_touches(uwsgi.touch_chain_reload);
 				if (touched) {
 					if (uwsgi.status.chain_reloading == 0) {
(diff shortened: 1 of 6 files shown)


xrmx:
@Shir0kamii tabs please :)

Shir0kamii:
@xrmx Sorry, that's good now!

core/master.c
```diff
@@ -861,6 +870,18 @@ int master_loop(char **argv, char **environ) {
...
+				touched = uwsgi_check_touches(uwsgi.touch_mules_reload);
+				if (touched) {
+					uwsgi_log_verbose("*** %s has been touched... mules reload !!! ***\n", touched);
+					uwsgi_reload_mules();
+					continue;
+				}
+				touched = uwsgi_check_touches(uwsgi.touch_spoolers_reload);
+				if (touched) {
+					uwsgi_log_verbose("*** %s has been touched... spoolers reload !!! ***\n", touched);
+					uwsgi_reload_spoolers(SIGHUP);
```
unbit: why the SIGHUP arg here ?

unbit:
The patch looks really good, i know i am asking a bit too much, but do you think is it possible to add a test for it like i did it here ?

https://github.com/unbit/uwsgi/commit/aeb2fc4f883da7d5b8d5c3fb88bcd74f50ce987b

it is pretty hard to test process management, but maybe it is enough to get infos from the stats server to check if the pid of the mules/spoolers change at the right time.

Shir0kamii:
Sorry, the SIGHUP was a mistake.

I will try to test these two features and come back to you soon!

Shir0kamii:
@unbit I did my best, tell me what you think of these tests

I didn't use the uwsgi stats server since it doesn't send any data about the mules (though it does for the spooler). Instead, I test the list of subprocesses of uwsgi before and after reloading.

the sleeps pains me but if didn't put them here, uwsgi wouldn't have time to reload and the test fails.

If you have a better idea to test this feature, just tell me!

unbit:
the approach is good, maybe i would enable the --auto-procname feature to recognize spooler and mules processes by the name

Shir0kamii:
I didn't know about that feature until now, it seems great !

I see it renames the processes of uwsgi, but I don't understand how you would use it in the context of the tests. Could you explain your thoughts please ?

xrmx:
Il 27/07/2016 19:31, Alexandre Bonnetain ha scritto:
> I didn't know about that feature until now, it seems great !
>
> I see it renames the processes of uwsgi, but I don't understand how you would
> use it in the context of the tests. Could you explain your thoughts please ?

So you can grep for the spooler / mules and see if that process has changed pid 
instead of checking the uwsgi one.

Shir0kamii:
@xrmx @unbit It seems I have learnt a lot about process management while making these tests, thank you for your guidance!

They should now be just like you want, tell me if something still bother you

uwsgi.h
```diff
@@ -2361,6 +2364,8 @@ struct uwsgi_server {
 
 	struct uwsgi_string_list *touch_reload;
 	struct uwsgi_string_list *touch_chain_reload;
+	struct uwsgi_string_list *touch_mules_reload;
```
unbit: please move new structure fields at the and, to maintain ABI compatibility

Shir0kamii:
@unbit This is done! Is there anything left before merging ?

unbit:
@xrmx safe for backport too

ultrabug:
Great news thanks @Shir0kamii and @unbit @xrmx !

I hope this gets tagged soon so we can have a real package for this :+1:

xrmx:
Backported to uwsgi-2.0 as 24fd91e1000251fa44f0cfd630bcf3c867336264

ultrabug:
Thanks, any planned 2.0.14 release plz ?

PR #1292, opened 2016-06-16
Support error override

smallredbean:
1. Suport ProxyErrorOverride + ErrorDocument #1282 
2. backend_broke bug fix #1286

PR #1330, opened 2016-08-03
Guard error_override against different APIs (particularly 2.2 to 2.4)

smallredbean:
The previous code in #1292 cannot compile with Apache httpd 2.4 API. The definition of error_override is moved from proxy_server_conf to proxy_dir_conf in commit 9eb0360e041928ca79a799c4d1cfa0a5377a6787.


xrmx:
Thanks!

PR #1333, opened 2016-08-06
php-constant option for the php plugin

gdamjan:
allows to define global constants, the same as if you `define` them in php

### DIFF
--- a/plugins/php/php_plugin.c
+++ b/plugins/php/php_plugin.c
@@ -17,6 +17,7 @@ struct uwsgi_php {
 	struct uwsgi_regexp_list *app_bypass;
 #endif
 	struct uwsgi_string_list *vars;
+	struct uwsgi_string_list *constants;
 	char *docroot;
 	char *app;
 	char *app_qs;
@@ -67,6 +68,7 @@ struct uwsgi_option uwsgi_php_options[] = {
         {"php-app-bypass", required_argument, 0, "if the regexp matches the uri the --php-app is bypassed", uwsgi_opt_add_regexp_list, &uphp.app_bypass, 0},
 #endif
         {"php-var", required_argument, 0, "add/overwrite a CGI variable at each request", uwsgi_opt_add_string_list, &uphp.vars, 0},
+        {"php-constant", required_argument, 0, "define a php constant for each request", uwsgi_opt_add_string_list, &uphp.constants, 0},
         {"php-dump-config", no_argument, 0, "dump php config (if modified via --php-set or append options)", uwsgi_opt_true, &uphp.dump_config, 0},
         {"php-exec-before", required_argument, 0, "run specified php code before the requested script", uwsgi_opt_add_string_list, &uphp.exec_before, 0},
         {"php-exec-begin", required_argument, 0, "run specified php code before the requested script", uwsgi_opt_add_string_list, &uphp.exec_before, 0},
@@ -223,8 +225,17 @@ static void sapi_uwsgi_register_variables(zval *track_vars_array TSRMLS_DC)
 		}
 		usl = usl->next;
 	}
-
-
+	usl = uphp.constants;
+	while(usl) {
+		char *equal = strchr(usl->value, '=');
+		if (equal) {
+			size_t name_len = equal-usl->value;
+			char *name = estrndup(usl->value, name_len);
+			char *strval = equal+1;
+			zend_register_string_constant(name, name_len, strval, CONST_CS, 0);
+		}
+		usl = usl->next;
+	}
 }
 
 static sapi_module_struct uwsgi_sapi_module;


gdamjan:
maybe backport this to 2.0.14 too?

@unbit @xrmx

unbit:
@xrmx safe for me

gdamjan:
huh, I've found two issues on debian jessie (`libphp5-embed  5.6.27+dfsg-0+deb8u1`) that didn't appear on my Arch system with 7.x (at the time of this PR).

1. strval needs to be `estrdup`-ed before calling `zend_register_string_constant` - which is ok in 7.0 too.
2. the call to `zend_register_string_constant` for some reason wants `name_len + 1` instead of just `name_len` - I have no idea how that makes sense, and it can't work like that on 7.0

gdamjan:
one more thing,
in 5.6 on debian, Zend/zend_constants.h has this macro:
`#define REGISTER_STRING_CONSTANT(name, str, flags)  zend_register_string_constant((name), sizeof(name), (str), (flags), module_number    TSRMLS_CC)`

in 7.0 on arch,
`#define REGISTER_STRING_CONSTANT(name, str, flags)  zend_register_string_constant((name), sizeof(name)-1, (str), (flags), module_number)`

there is that `sizeof(name) -1` in 7.0. I should better use the macro itself, but then I would need that `module_number` which I don't know what is.

xrmx:
@gdamjan care to cook a patch? regarding the second issue just try REGISTER_STRING_CONSTANT I've tried to dig a bit using the github search but haven't found who is supposed to define it. If you don't want to dig into php source code to sort out what's module_number we can do some ifdefery (UWSGI_PHP7).

gdamjan:
the `module_number` is added to a function defined with some macros. it's all awful, that's how far I've come.

gdamjan:
I can move the constants to `uwsgi_php_minit` actually, I can use the `REGISTER_STRING_CONSTANT` macro there, since the function is defined through the `PHP_MINIT_FUNCTION` macro.

Alas, the `REGISTER_STRING_CONSTANT` uses `sizeof` in way that only works with literal strings, and not with variables. In other words I can't use it.

TIL
```
sizeof("SOME STRING") -> 11
// but
sizeof(name) -> 8
```

It seems to me that an "ifdefery" will be the only way forward.

gdamjan:
preliminary patch at https://github.com/gdamjan/uwsgi/commit/2761a85f9bae4148e7aa7fa3762c5498c8cfb1af

@iron-udjin can you check too?
also, if `--php-constant = A_CONSTANT=testing123` and 
```
if ( defined("A_CONSTANT")) {
    $a = A_CONSTANT;
    echo "Test: " . $a . "\n";
}
```
works

xrmx:
@gdamjan your patch looks good to me, confirmation that it works too would be appreciated :)

gdamjan:
tested it, and works on debian 8 (`libphp5-embed 5.6.27+dfsg-0+deb8u1`) and Arch (`php 7.0.13-2`).
I want to test it on ubuntu xenial too, and maybe dotdeb php packages.
freebsd/macOS is out of my capabilities.

gdamjan:
Ubuntu Xenial with `PHP 7.0.8-0ubuntu0.16.04.3` works too

gdamjan:
finally made it work with php56 on FreeBSD 11. compiles and the php-constant option both work.

preparing php from ports itself for uwsgi is a pain though.

I'm gonna open a new pull request.

xrmx:
@gdamjan thanks! btw could you please check if using usl->value instead of duplicating it as a name works too? It looks like the zend function already dup the string.

gdamjan:
I did try without the dup, it crashed badly.

the change:
```                        size_t name_len = equal - usl->value;
-                       char *name = estrndup(usl->value, name_len);
+                       char *name = usl->value;
+                       equal = NULL;
                        char *strval = equal + 1;
```

the crash:
```
!!! uWSGI process 2978 got Segmentation Fault !!!
*** backtrace of 2978 ***
./uwsgi(uwsgi_backtrace+0x2c) [0x46ad6c]
./uwsgi(uwsgi_segfault+0x21) [0x46b141]
/usr/lib/libc.so.6(+0x330b0) [0x7fecabf3e0b0]
/usr/lib/libc.so.6(strlen+0x26) [0x7fecabf8b446]
/usr/lib/libphp7.so(zend_register_string_constant+0x1f) [0x7fecac5f543f]
./uwsgi(zm_startup_uwsgi_php_minit+0x50) [0x47fa20]
/usr/lib/libphp7.so(zend_startup_module_ex+0x12e) [0x7fecac609bde]
/usr/lib/libphp7.so(+0x360c8c) [0x7fecac609c8c]
/usr/lib/libphp7.so(zend_hash_apply+0x82) [0x7fecac617a52]
/usr/lib/libphp7.so(zend_startup_modules+0x2a) [0x7fecac609f3a]
/usr/lib/libphp7.so(php_module_startup+0x713) [0x7fecac5a4543]
./uwsgi() [0x47f9b8]
./uwsgi() [0x48039d]
./uwsgi(uwsgi_start+0x586) [0x46c376]
./uwsgi(uwsgi_setup+0x2616) [0x46faf6]
./uwsgi(main+0x9) [0x41af09]
/usr/lib/libc.so.6(__libc_start_main+0xf1) [0x7fecabf2b291]
./uwsgi(_start+0x2a) [0x41af3a]
*** end of backtrace ***
```

gdamjan:
ahahaa, just as I posted this I've seen my error.
ok works without the strdup too

PR #1341, opened 2016-08-16
plugin/rados: add ability to configure rados namespace

funny-falcon:
Let user configure rados pool namespace.
If librados is too old, then non-empty namespace is error.


funny-falcon:
Alternative way could be to allow pass namespace in url. It is easy to do for threaded environment, but harder for async (cause one ioctx used).

unbit:
Thanks a lot

PR #1342, opened 2016-08-19
stdio logger

gdamjan:
logs to uwsgi.original_log_fd

example:

     ./uwsgi --master -s :0 --logger 'stdio stdio:' --log-route 'stdio ^'

I'm open for a better name instead of `stdio`

### DIFF
--- a/plugins/logfile/logfile.c
+++ b/plugins/logfile/logfile.c
@@ -1,4 +1,5 @@
 #include <uwsgi.h>
+extern struct uwsgi_server uwsgi;
 
 struct logfile_data {
 	char *logfile;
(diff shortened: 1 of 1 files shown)


gossel-j:
This will be very useful to log to both syslog/file/etc and terminal.

unbit:
what about simply calling it original_fd ?

gdamjan:
seems a bit too long, also it only makes sense if you know what the code does

stdio seemed most close to stderr or out

unbit:
@xrmx safe for backport

thijsterlouw:
Is this supposed to work in daemon mode? Because I would like to log from a docker container to stdout, while running in daemon mode. This seems useful, but doesn't appear to do anything even though I run 2.0.14 ( http://uwsgi-docs.readthedocs.io/en/latest/Changelog-2.0.14.html )

gdamjan:
daemon mode and stdio logging dont make sense together, why would you even use daemon mode in docker???!

thijsterlouw:
because we want to run a few processes together. In this case consul-template is running uwsgi. But indeed, we could run it slightly differently and not use daemonize. I understand it makes little sense, but when logging via the master it could be done. Actually logging in daemonized mode to /dev/stdout almost works, except that logging.c depends on seek (for logrotate, which causes error and seemingly cannot be disabled)

Issue #1343, opened 2016-08-19
What is the right way to return no answer and silently close connection in PSGI protocol?


jef-sure:
I want to jailbreak out of PSGI and use `$env->{'psgix.io'}` directly, but I can't find the right way to do it. Twiggy uses [] as for such an action, but uWSGI says "invalid PSGI status code". There's no official documents for such a jailbreak but Twiggy was written by author of the PSGI specs. 

In your websocket examples you don't return from handler at all but I think it's not portable. For exmaple, Twiggy doesn't call handlers `async`-ed and they can't be blocked forever.

PR #1344, opened 2016-08-19
Work around some inconvenience in PSGI specs for out of PSGI protocols

jef-sure:
This patch should close issue #1343


jef-sure:
av_top_index() was introduced only in perl 5.18. So, I replaced it with old av_len()

unbit:
@xrmx safe for backport

unbit:
@jef-sure thanks a lot

xrmx:
*cough* squash next time please *cough*

jef-sure:
I think it's better not to close $env->{'psgix.io'} on PSGI cancel. Unfortunately, it takes time to go through unfamiliar uwsgi sources. That's why one more patch.

Issue #1339, opened 2016-08-15
Infinite loop when emperor dies and vassal is trying to go down


shoham-stratoscale:
When emperor goes down, the vassals should see that the communication with it is lost, and to go down by themselves. This is done by worker-id 1.
https://github.com/unbit/uwsgi/blob/master/core/uwsgi.c#L3661
But, when worker-id 1 calls gracefully_kill_them_all(), to kill all the workers, it kills itself first!
So after killing it self, it's vassal process will bring it back up, and it will kill itself again, and the going down process will be stuck in an infinite loop.

Solutions might be that the check if the emperor is down will be done by the vassal's master process, that will kill the workers, or that worker-id 1 will not kill itself. I'm pro the first solution

PR #1346, opened 2016-08-22
Don't kill yourself when killing all

shoham-stratoscale:
gracefully_kill_them_all() may be called on worker id 1.
When that happens, the worker may try to kill itself by mistake.

Added a check for the worker not to kill itself.

Solves #1339


unbit:
You mean when no master is in place ?

shoham-stratoscale:
This only happens when using emperor mode.
When using emperor, the wid 1 is the one responsible of killing all the workers, not the master

PR #1347, opened 2016-08-22
don't close $env->{'psgix.io'} on "PSGI cancel"

jef-sure:
let application do it


jef-sure:
With this patch I don't have to `dup()` socket before returning `[]`

unbit:
in such a case if the dev forget to close the fd we will have a leak. It is how psgix.io is expected to work ?

jef-sure:
I can't see any alternative. It's not so difficult to close the socket in DESTROY method of request, so this usually not a problem. When application takes care of the socket and protocol, it must control when it closes the socket.

jef-sure:
When I have to `dup()` the socket -- I have the same "socket leak" problem.

PR #1348, opened 2016-08-23
Add reload-on-{pss,uss}

bplotnick:
This pull request adds `reload-on-pss` and `reload-on-uss` options for linux only. It parses `/proc/self/smaps` to determine these values. See [this](https://lwn.net/Articles/230975/) article for explanation of these values.

I have not yet added support for other operating systems, since my development environment is linux only right now. If we want to add more OS support, [psutil](https://github.com/giampaolo/psutil) is a good reference for multiple OS implementations.

### DIFF
--- a/core/logging.c
+++ b/core/logging.c
@@ -864,6 +864,27 @@ void get_memusage(uint64_t * rss, uint64_t * vsz) {
 
 }
 
+//TODO: add non-linux uss and pss metrics
+void get_smaps_memusage(uint64_t * uss, uint64_t * pss) {
+#ifdef __linux__
+  FILE *file = fopen("/proc/self/smaps", "r");
+
+  char line [BUFSIZ];
+  while (fgets(line, sizeof line, file))
+    {
+      char substr[32];
+      int n;
+      if (sscanf(line, "%31[^:]: %d", substr, &n) == 2)
+        {
+	  if (strcmp(substr, "Private_Clean") == 0)  { *uss += n * 1024; }
+	  else if (strcmp(substr, "Private_Dirty") == 0)  { *uss += n * 1024; }
+	  else if (strcmp(substr, "Pss") == 0)            { *pss += n * 1024; }
+        }
+    }
+  fclose(file);
+#endif
+}
+
 void uwsgi_register_logger(char *name, ssize_t(*func) (struct uwsgi_logger *, char *, size_t)) {
 
 	struct uwsgi_logger *ul = uwsgi.loggers, *old_ul;
--- a/core/logging.c
+++ b/core/logging.c
@@ -866,21 +866,23 @@ void get_memusage(uint64_t * rss, uint64_t * vsz) {
 
 #ifdef __linux__
 void get_memusage_extra(uint64_t * uss, uint64_t * pss) {
-  FILE *file = fopen("/proc/self/smaps", "r");
+	FILE *file = fopen("/proc/self/smaps", "r");
 
-  char line [BUFSIZ];
-  while (fgets(line, sizeof line, file))
-    {
-      char substr[32];
-      int n;
-      if (sscanf(line, "%31[^:]: %d", substr, &n) == 2)
-        {
-	  if (strcmp(substr, "Private_Clean") == 0)  { *uss += n * 1024; }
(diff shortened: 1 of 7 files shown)


bplotnick:
@xrmx @unbit Any feedback on this? We've been running it internally for a few weeks now and it seems to work fine.

xrmx:
@bplotnick new fields at the end of the struct please to keep ABI compatibility. Once you are done please squash your commit into one

bplotnick:
I just did that for uwsgi.h. Does this need to happen for `uwsgi_base_options`? Where is ABI compatibility a concern here?

Also, you should be able to squash the commit in github when you merge (there should be an option for it).

core/logging.c
```diff
@@ -864,6 +864,26 @@ void get_memusage(uint64_t * rss, uint64_t * vsz) {
 
 }
 
+#ifdef __linux__
+void get_memusage_extra(uint64_t * uss, uint64_t * pss) {
+  FILE *file = fopen("/proc/self/smaps", "r");
+
+  char line [BUFSIZ];
+  while (fgets(line, sizeof line, file))
+    {
```
xrmx: Please use K&R for indentation and brackets, also tabs please

unbit:
Hi, thanks for your work. I think we can add uss/pss to all the operating system (as fields of the structure). While it will work only on linux we will be prepared for future patches for the other operating systems. Can you add them even to the stats server ?

bplotnick:
@xrmx Sorry about the formatting! My editor was misconfigured, so hopefully it's good now

@unbit I got rid of the `#ifdef` in the struct and added uss/pss to the stats server. I've only tested this on linux, though.

unbit:
One last thing, can you govern the printing of uss in the log (when --memory-report) with a --memory-report-extra option ? This is for not causing issues with users log parsers

bplotnick:
@unbit I'd like to have it so that `--memory-report-extra` implies `--memory-report` but i can't figure out a way to set one option as a result of another option being set.

One alternative option would be to add something like `UWSGI_OPT_MEMORY_EXTRA`. Another option would be to have `uwsgi.logging_options.memory_report` be an integer instead of a boolean and have `--memory-report` take an optional argument, which when `2` would include `uss` and `pss`. The other option is to just check for `uwsgi.logging_options.memory_report_extra` everywhere we check for `uwsgi.logging_options.memory_report`, which is pretty ugly 😕 .

Do you have a preference?

unbit:
@bplotnick i will add a flag to impliy memory reporting, so you can work on it. Give me a couple of days

shivawu:
@bplotnick This is great! Thanks for working on this. Have a little concern, have you considered the performance impact brought by parsing smaps?

bplotnick:
Hey @shivawu. Sorry for the delay on the microbenchmark. Using [this](https://gist.github.com/bplotnick/87f18e199f3b3f7fb1bd4941f5d7213c), I found that the `get_memusage_extra` function takes about 160 microseconds. That's on a 2.3GHz Sandy Bridge processor.

drolando:
@unbit @bplotnick @xrmx Is there any update on this? Are we still blocked on the memory-report extra option?

unbit:
Hi, the only problem i see is here: https://github.com/unbit/uwsgi/pull/1348/files#diff-18630480decdbb0c741c4ab30fe2e857R720

we should avoid changing default logging format.

Please add memory-report-extra option so users will be free to have it in logs or not

Thanks a lot

bplotnick:
hey @unbit I actually addressed that way back in my comment on Oct 5. I wanted to have `--memory-report-extra` imply `--memory-report`, but i couldn't figure out how to do it. I think if I make `uwsgi.logging_options.memory_report` be an integer instead of a boolean, I can have `--memory-report-extra` set that to something like 2, which implies `memory-report` as well as the extra bits. Does this sound good?

If you could point me in the right direction, i'd love to add that so we can finally get this merged!

bplotnick:
I rebased (I hope that's ok :worried:, the only difference from before is 773e8c3) and changed `memory-report` to take an optional argument. That is, if it is `2`, it will give you the uss and pss report.

It is not optimal because you have to specify it with an equals sign like `--memory-report=2`, but I think it works for what we're trying to do.

unbit:
Seems good to me :)

PR #1355, opened 2016-09-01
Misc coverity fixes

xrmx:
@unixwitch could you please check the routing patches?
@worr could you please check the socket patch?


worr:
@xrmx: socket patch lgtm. Thanks!

PR #1371, opened 2016-09-29
graceful shutdown resets pending responses resulting in 502 errors

rectalogic:
We front uwsgi with nginx, and shutdown uwsgi by sending "q" to the master fifo. This causes uwsgi to finish processing any requests and then exit, which closes the sockets that nginx is reading.

If nginx has not yet finished reading a response, it gets an RST from uwsgi and fails the request with a 502 and logs the ECONNRESET (104):
```
2016/09/29 22:11:00 [error] 32643#0: *79 recv() failed (104: Connection reset by peer) while reading response header from upstream, client: 127.0.0.1, server: , request: "GET / HTTP/1.1", upstream: "uwsgi://unix:/tmp/uwsgi.sock:", host: "localhost:9091"
127.0.0.1 - - [29/Sep/2016:22:11:00 +0000] "GET / HTTP/1.1" 502 173 "-" "curl/7.35.0"
```

So when gracefully shutting down uwsgi, nginx is returning HTTP 502 errors to clients.

The socket needs to be shutdown() before closing so that we go through a normal `FIN -> ACK/FIN -> ACK -> closed` roundtrip and then can safely close the socket.

> close() with any pending readable data could lead to an immediate reset being sent

https://blog.netherlabs.nl/articles/2009/01/18/the-ultimate-so_linger-page-or-why-is-my-tcp-not-reliable

This PR attempts to fix this.

You can reproduce the issue using nginx and uwsgi with the config files and scripts in this gist
https://gist.github.com/rectalogic/0f7603be7d4c8e8520598d8b9e42c3e1#file-readme-md

Clone the gist, symlink nginx and uwsgi binaries into that directory and run `nginx.sh` and `uwsgi.sh` in separate terminals, and then `runquit.sh` and watch nginx stderr for the `recv() failed (104: Connection reset by peer)` error.

### DIFF
--- a/core/loop.c
+++ b/core/loop.c
@@ -60,6 +60,8 @@ void *uwsgi_get_loop(char *name) {
 
 void simple_loop() {
 	uwsgi_loop_cores_run(simple_loop_run);
+	if (uwsgi.workers[uwsgi.mywid].close_sockets)
+		uwsgi_close_all_sockets();
 }
 
 void uwsgi_loop_cores_run(void *(*func) (void *)) {
--- a/core/socket.c
+++ b/core/socket.c
@@ -1259,8 +1259,10 @@ void uwsgi_close_all_sockets() {
         struct uwsgi_socket *uwsgi_sock = uwsgi.sockets;
 
         while (uwsgi_sock) {
-                if (uwsgi_sock->bound)
+                if (uwsgi_sock->bound) {
+                        shutdown(uwsgi_sock->fd, SHUT_RDWR);
                         close(uwsgi_sock->fd);
+                }
                 uwsgi_sock = uwsgi_sock->next;
         }
 }
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -1288,7 +1288,8 @@ void gracefully_kill(int signum) {
 		struct wsgi_request *wsgi_req = current_wsgi_req();
 		wait_for_threads();
 		if (!uwsgi.workers[uwsgi.mywid].cores[wsgi_req->async_id].in_request) {
-			uwsgi_close_all_sockets();
+			if (uwsgi.workers[uwsgi.mywid].close_sockets)
+				uwsgi_close_all_sockets();
 			exit(UWSGI_RELOAD_CODE);
 		}
 		return;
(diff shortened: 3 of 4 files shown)


unbit:
Hi, thanks for the patch. Unfortunately in its current incarnation it breaks graceful reload (where sockets must not be closed).

I suggest you to call uwsgi_close_all_sockets in simple_loop (core/loop.c) only if graceful stop has been asked. This would probably require adding a new field (close_sockets ?) in the uwsgi_instance structure

rectalogic:
Hmm, but the worker just exits from `gracefully_kill`, so we never return control to `simple_loop` to be able to call `uwsgi_close_all_sockets`.

Also I'm not clear how the master would communicate the `uwsgi_instance` field to the worker, the workers copy doesn't share memory with the master. e.g. when I gracefully quit the master sets `uwsgi.status.gracefully_destroying` but this is still unset in the worker.

unbit:
uwsgi.workers holds a shared memory array. There you can access per-worker data.

From a worker uwsgi.workers[mywid] will return the proper structure. Here you can put a mark (from the master) that the worker can read anc interpret as 'close all sockets'

gracefully_kill() calls exit() only if the worker is doing nothing, in such a case you have to add the check here. Otherwise in the loop (that runs only if manage_next_request is non-zero)

unbit:
Really good work, thanks.

rectalogic:
So this fix (or the modified one in #1409) didn't make it into release 2.0.15? Any idea when this will be released? We've been waiting 6 months for this fix.

rectalogic:
So this fix (or #1409) is not in release 2.0.16 either? We've been waiting over a year now for this fix. What is the release plan for this fix @unbit ?

unbit:
Sorry, completely get lost in the queue of pull request.s A new release will be issued during the week-end. i will include it for sure

unbit:
@rectalogic the patch landed in LTS tree. Take into account that lacking a proper test suite any logic change in LTS will be governed by an option. In this casa the option is `--shutdown-socket`. In uWSGI 2.1/3.0 will be the default behaviour

PR #1388, opened 2016-10-19
Make sure wsgi.errors is line-buffered on Python 3

### LINKED ISSUE #1222: Please make wsgi.errors line-buffered or unbuffered
When stderr goes to the console, frameworks like bottle successfully log exceptions to stderr.  When stderr gets redirected via --daemonize, --logger or similar, it ends up block buffered and messages get lots or misplaced.

I couldn't find any statement in PEP 333 or PEP 3333 about what type of buffering should apply to wsgi.errors, but IMO it would be much more polite if uwsgi never used block buffering.

mgedmin:
Should fix #1222.

This patch hasn't been tested in any way.

### DIFF
--- a/plugins/python/uwsgi_python.h
+++ b/plugins/python/uwsgi_python.h
@@ -72,7 +72,7 @@ PyAPI_FUNC(PyObject *) PyMarshal_ReadObjectFromString(char *, Py_ssize_t);
 #define	PyString_Size		PyBytes_Size
 #define	PyString_Concat		PyBytes_Concat
 #define	PyString_AsString	(char *) PyBytes_AsString
-#define PyFile_FromFile(A,B,C,D) PyFile_FromFd(fileno((A)), (B), (C), -1, NULL, NULL, NULL, 0)
+#define PyFile_FromFile(A,B,C,D) PyFile_FromFd(fileno((A)), (B), (C), 1, NULL, NULL, NULL, 0)
 #define uwsgi_py_dict_get(a, b) PyDict_GetItem(a, PyBytes_FromString(b));
 #define uwsgi_py_dict_del(a, b) PyDict_DelItem(a, PyBytes_FromString(b));
 


vytas7:
@xrmx The proposed patch works just fine in my testing, using CPython 3.12.9, uWSGI `master` (at the time of this writing, 0bcd3838) + this patch, gcc version 11.4.0.

Using the below `test.py` + [Falcon 4.0.2](https://pypi.org/project/falcon/4.0.2/) (which in turn has no other dependencies):
```python
import falcon


class ErrorResource:
    def on_get(self, req, resp):
        resp.media = {'message': 'Testing uWSGI errors'}
        req.log_error('Something is not quite right with the error logging.')

    def on_put(self, req, resp):
        resp.text = str(0 / 0)


application = app = falcon.App()
app.add_route('/error', ErrorResource())
```
...which is run as:
```
uwsgi --http :8000 --logto /tmp/uwsgi.log --wsgi-file test.py
```

Performing `GET` and `PUT` requests to http://localhost:8000/error then makes the respective error messages show up in `/tmp/uwsgi.log` immediately as expected.

vytas7:
@xrmx it is true, as an Emacs user I can confirm that you can never know: :sweat_smile: 
image
(https://xkcd.com/1172/)

Seriously speaking though, almost all other WSGI server out there do wrangle the flushing part for the user.
E.g.:
* Gunicorn: https://github.com/benoitc/gunicorn/blob/master/gunicorn/http/wsgi.py#L66
* Waitress: https://github.com/Pylons/waitress/blob/ed0149beb3c91def3150ac3cbd57df250a241a9c/src/waitress/task.py#L549 (uses `sys.stderr` which is unbuffered)
* Meinheld: https://github.com/mopemope/meinheld/blob/6ddb22b52783d22ba15c6cdfbcbc9ed05b5c0c46/meinheld/server/http_request_parser.c#L897 (also simply obtains `sys.stderr)

Issue #1384, opened 2016-10-14
Bad regression in 2.0.9 due to my pull request #772 in high-memory psgi environments


avar:
Back in #772 I submitted code to fix the psgi implementation so that we tear down the interpreter on exit. This is good if you want your DESTROY hooks run and any other at-exit actions carried out by perl itself. This was backported to 2.0.9 in 9d4f4eb.

However this is much more expensive than I thought, something I hadn't anticipated and am only just figuring out now since I'm deploying 2.0.14 to production (we'd been running 2.0.3 with various patches for a while).

We have an application that has a large preload size, since we take advantage of copy-on-write to load a lot of code/data that we then use in the children, and this is really cheap due to *nix fork() semantics.

I have an application like that where the master and any given child is 4GB RES size (4.5GB VSZ), with 50 workers on a 48 core box.

If I psgi.harakiri.commit on the first request and make around 2 requests/sec I consistently have 5-10 workers pegged at 100% CPU. The reason is that their perl interpreter is being freed, which results in a lot of refcounts being updated, which means the kernel has to frantically run around copying memory pages before it can destroy the process.

Running "service uwsgi stop" then completely hoses the box. Before it has ~20GB of used memory with ~130GB on the box. Running "stop" causes all the workers to exit at once and free their interpreters, which causes all the memory for each one to be completely copied.

This causes 100% CPU on all cores, and the memory use rapidly rises to 100% before the OOM killer kicks in.

Now, there's basically two ways to deal with this:

 * Randomly sleep for X seconds in the atexit handler. In my case randomly sleeping for 0..60s does the trick.
 * Call POSIX::_exit(0) as the very last thing I do in the atexit handler. This skips the whole `destroyperl` codepath and brings us back to pre-2.0.9 semantics: https://github.com/unbit/uwsgi/blob/master/plugins/psgi/psgi_plugin.c#L857 although we might be skipping some important internal uWSGI code too, but we're exiting anyway, so it can't be that important

I'm doing the latter as a workaround for this, but I think it makes sense to have some option or general facility for this, but since the atexit hooks are cross-language I thought I'd open a discussion here. What do you think the facility should be? A startup-time option? A flag passed/returned to the atexit function? Calling some uwsgi::* function to set this in the child?

[...]

PR #1392, opened 2016-10-20
core: Add a --skip-atexit-teardown option to skip perl/python teardown

avar:
This is a general option meant to be used for any number of embedded
languages, but which is currently just used by python & perl.

As detailed at length in my issue #1384[1] the full destruction I
added for perl in 1dcdb72[2] can be *very* expensive when a child
dies, and in particular when all children die at once when the server
goes down. This was first introduced in uWSGI 2.0.9.

This is not a perl-specific issue as the issue notes, but rather a
generic issue with embedding any number of programming language
interpreters & compilers. Fully tearing them down can be very
expensive, but just calling C's exit function is cheap, and often
people running uWSGI at scale are very willing to make that trade-off.

This issue also impacts python embedding since uWSGI 1.1 (f4f4d5a) as
noted in a follow-up comment of mine to that issue.

So to fix this add a --skip-atexit-teardown option to allow skipping
DESTROY during perl destruction, __del__ during python destruction
etc. This can then be used for other languages in the future as we add
uwsgi_*_atexit handlers for them.

This emulates the hack of calling POSIX::_exit() in perl or
os._exit(0) in python at the very end of the atexit handler, but
allows uWSGI to do any cleanup it wanted to do on its own after
calling atexit.

1. https://github.com/unbit/uwsgi/issues/1384
2. https://github.com/unbit/uwsgi/commit/1dcdb72

### DIFF
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -95,6 +95,7 @@ static struct uwsgi_option uwsgi_base_options[] = {
 
 	{"skip-zero", no_argument, 0, "skip check of file descriptor 0", uwsgi_opt_true, &uwsgi.skip_zero, 0},
 	{"skip-atexit", no_argument, 0, "skip atexit hooks (ignored by the master)", uwsgi_opt_true, &uwsgi.skip_atexit, 0},
+	{"skip-atexit-teardown", no_argument, 0, "skip atexit teardown (ignored by the master)", uwsgi_opt_true, &uwsgi.skip_atexit_teardown, 0},
 
 	{"set", required_argument, 'S', "set a placeholder or an option", uwsgi_opt_set_placeholder, NULL, UWSGI_OPT_IMMEDIATE},
 	{"set-placeholder", required_argument, 0, "set a placeholder", uwsgi_opt_set_placeholder, (void *) 1, UWSGI_OPT_IMMEDIATE},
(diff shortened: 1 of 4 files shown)


uwsgi.h
```diff
@@ -2539,6 +2539,7 @@ struct uwsgi_server {
 	int no_orphans;
 	int skip_zero;
 	int skip_atexit;
+	int skip_atexit_teardown;
```
xrmx: new fields at the end of the struct please

core/uwsgi.c
```diff
@@ -95,6 +95,7 @@ static struct uwsgi_option uwsgi_base_options[] = {
 
 	{"skip-zero", no_argument, 0, "skip check of file descriptor 0", uwsgi_opt_true, &uwsgi.skip_zero, 0},
 	{"skip-atexit", no_argument, 0, "skip atexit hooks (ignored by the master)", uwsgi_opt_true, &uwsgi.skip_atexit, 0},
+	{"skip-atexit-teardown", no_argument, 0, "skip atexit teardown (ignored by the master)", uwsgi_opt_true, &uwsgi.skip_atexit_teardown, 0},
```
xrmx: shouldn't we make default the safer option instead?

unbit:
@xrmx i agree with @avar, the current behaviour is the most expected (albeit not-safe in a bunch of scenarios) and is a well-established practice for developers. The patch is not invasive, so i think currently is the best approach

unbit:
@xrmx safe for backport too

xrmx:
Backported and pushed to uwsgi-2.0. Thanks!

PR #1403, opened 2016-11-11
[enh] mule: direct patch dispatch to certain plugin

zevlg:
with 0187247899d7ec303264c3daf601b7dd802643e4 `python_plugin` now extremely limits how patch for mules can be specified, i.e. it forces to be dispatched by `python_plugin` if mule patch contains colon

This change introduces `<PLUGIN>(<PATCH>)` syntax for mule patches, for things like:

    mule= python(module:callable)
    mule= customplugin(another:string:containing:colon)

is possible

### DIFF
--- a/core/mule.c
+++ b/core/mule.c
@@ -89,12 +89,34 @@ void uwsgi_mule(int id) {
 }
 
 void uwsgi_mule_run() {
-	int id = uwsgi.muleid;
+        char* patch;
 	int i;
-	if (uwsgi.mules[id - 1].patch) {
+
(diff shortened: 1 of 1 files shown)


core/mule.c
```diff
@@ -89,12 +89,34 @@ void uwsgi_mule(int id) {
 }
 
 void uwsgi_mule_run() {
-	int id = uwsgi.muleid;
+        char* patch;
```
xrmx: tabs please

zevlg: yeah, done.  There was already spaces in `uwsgi_mule_run`, so I thought there is a tendency to use spaces

PR #1410, opened 2016-11-29
core/mount.c:uwsgi_mount: support "defaults" for options

unixwitch:
This change allows `mount()` options to be set to 0 by specifying `defaults` (like `/etc/fstab`), and fstype to `NULL`, to support bind mounts.


unbit:
Thanks

PR #1413, opened 2016-12-01
fix offload-threads with honour-range

unixwitch:
```
when using --offload-threads with static files, the range request
offset was not passed to the offload thread, meaning uwsgi would
respond with 206 Partial Content, but return the entire file.
```

I've tested this in our staging environment and it fixes the particular problem we were having: serving video requires range requests in most browsers, but we use offload-threads, and it caused playback to hang when the browser made the first range request.

It hasn't had a lot of testing, and there may still be other places it's broken, but I don't think it's introduced any new bugs.

### DIFF
--- a/core/offload.c
+++ b/core/offload.c
@@ -674,11 +674,12 @@ void uwsgi_offload_engines_register_all() {
 	uwsgi.offload_engine_pipe = uwsgi_offload_register_engine("pipe", u_offload_pipe_prepare, u_offload_pipe_do);
 }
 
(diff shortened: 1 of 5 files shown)


xrmx:
I think you missed to update the  uwsgi_offload_request_sendfile_do call in plugins/python/uwsgi_pymodule.c no?

unbit:
@xrmx the offload api in python is currently broken (it is commented) in master. I do not know if it will be ever implemented now that we have a routing api doing the same thing

chaosk:
@unbit Would it be possible to backport this to 2.0?

xrmx:
@chaosk cherry-picked in uwsgi-2.0

PR #1415, opened 2016-12-06
fix uwsgi_request_body_readline without new lines found 

### LINKED ISSUE #1412: flaw in readline implementation causing it to return excess data
Here is - directly copied from https://bugs.debian.org/846362 (so please credit Juho Heikkinen <juho.heikkinen@f-secure.com> for the info but blame me for any typos):

The language-independent readline implementation in uwsgi's 
core/reader.c appears to contain a flaw causing it to append extra bytes 
from the readline buffer to the final line returned, if the request body 
is not terminated by a newline. Reproducing this seems to require that 
the request body is at least 4KB in size.

The issue originally appeared in context of running a Python application 
with the uswgi Python plugin, but investigation led me to believe that 
it's in the uwsgi core itself. The issue is also present in upstream. 
The fix seems to be to replace near the end of 
uwsgi_request_body_readline():

  *rlen = wsgi_req->post_readline_size - wsgi_req->post_readline_pos;

with:

  *rlen = wsgi_req->post_readline_watermark - wsgi_req->post_readline_pos;

To reproduce:

  * Install uwsgi, uwsgi-plugin-python3, python3-werkzeug
  * Run the attached repro-app.py with: uwsgi_python34 --socket 
0.0.0.0:8080 --protocol=http -w repro-app
  * Generate a multiline request body >4KB with: ( for x in $(seq -w 
1000); do echo $x; done; echo -n "final" ) >data.txt
  * Send request to uwsgi: curl http://localhost:8080/ --data-binary 
@data.txt >data2.txt

Expected:

data.txt and data2.txt are identical

Actual:

data2.txt contains data in data.txt, followed by several lines of extra 
data from the readline buffer.

While this reproduction only causes extra data from the readline buffer 
to be returned, we have also seen cases where the returned extra bytes 
seem to be other unrelated data from the process heap. This has probably 
been caused by consume_body_for_readline() having realloced the buffer 
just before.

xrmx:
The language-independent readline implementation in uwsgi's
core/reader.c appears to contain a flaw causing it to append extra bytes
from the readline buffer to the final line returned, if the request body
is not terminated by a newline. Reproducing this seems to require that
the request body is at least 4KB in size.

While this reproduction only causes extra data from the readline buffer
to be returned, we have also seen cases where the returned extra bytes
seem to be other unrelated data from the process heap. This has probably
been caused by consume_body_for_readline() having realloced the buffer
just before.

Fixes #1412

### DIFF
--- a/core/reader.c
+++ b/core/reader.c
@@ -283,7 +283,7 @@ char *uwsgi_request_body_readline(struct wsgi_request *wsgi_req, ssize_t hint, s
                 }
 
 	// no line found, let's return all
-        *rlen = wsgi_req->post_readline_size - wsgi_req->post_readline_pos;
+        *rlen = wsgi_req->post_readline_watermark - wsgi_req->post_readline_pos;
         char *buf = wsgi_req->post_readline_buf + wsgi_req->post_readline_pos;
 	wsgi_req->post_readline_pos = 0;
 	return buf;
--- a/t/core/readline/app.py
+++ b/t/core/readline/app.py
@@ -1,0 +1,16 @@
+#!/usr/bin/python3
+# uwsgi --plugin python,http --http 0.0.0.0:8000 -w app
+
+from werkzeug.wrappers import Request, Response
+
+
+def application(env, start_response):
+    request = Request(env)
+    lines = b""
+    while True:
+        line = request.stream.readline()
+        if not line:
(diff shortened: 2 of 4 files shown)


xrmx:
@unbit please review

PR #1418, opened 2016-12-07
fix several issues with zend_register_string_constant

gdamjan:
* move the call in the PHP_MINIT_FUNCTION(uwsgi_php_minit) function and
make the constants persistent. they don't change
* there seems to be some difference between php5 and php7 how it wants
the `name_len` in the call. php5 wants to include the NULL terminator
too. (super weird)
* introduce module_number too - for whatever reason it is

### DIFF
--- a/plugins/php/php_plugin.c
+++ b/plugins/php/php_plugin.c
@@ -225,17 +225,6 @@ static void sapi_uwsgi_register_variables(zval *track_vars_array TSRMLS_DC)
 		}
 		usl = usl->next;
 	}
-	usl = uphp.constants;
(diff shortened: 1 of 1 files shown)


gdamjan:
this was tested on:
* Debian 8 with `libphp5-embed 5.6.27+dfsg-0+deb8u1` 
* ArchLinux with `php 7.0.13-2`
* Ubuntu 16.04 with `libphp7.0-embed 7.0.8-0ubuntu0.16.0`
* FreeBSD 11 with `php56`

xrmx:
Merged, thanks!

PR #1421, opened 2016-12-07
Add set_logvar and get_logvar methods in PSGI plugin

Shevviakov:
Implementation of set_logvar() and get_logvar() methods for PSGI plugin. These methods allow users to create and read user-defined log variables from psgi app.

The solution is just rewritten version of python implementation (/plugins/python/uwsgi_pymodule.c)


unbit:

PR #1437, opened 2017-01-10
fix compilation with php 7.1

### LINKED ISSUE #1427: Backward incompatible changes in php 7.1 (sigh)
https://github.com/php/php-src/commit/2809a676b55cebd4c6f8a8895ffe286db0704e87#diff-c6995589f3b41ca3c9b4c17c8327f7f9

so compiling uwsgi + php fails with:

```
plugins/php/php_plugin.c:600:2: error: initialization from incompatible pointer type [-Werror=incompatible-pointer-types]
  sapi_uwsgi_log_message,         /* Log message */
  ^~~~~~~~~~~~~~~~~~~~~~
plugins/php/php_plugin.c:600:2: note: (near initialization for ‘uwsgi_sapi_module.log_message’)
```

they added a `int syslog_type_int` argument to the log function. dunno how they expect this API to be used through versions.

gdamjan:
php 7.1 changed the signature of the log_message function in
sapi_module_struct. Let's assume it will stay like this for all versions >= 7.1

fixes #1427 

applies to the uwsgi-2.0 branch too

### DIFF
--- a/plugins/php/php_plugin.c
+++ b/plugins/php/php_plugin.c
@@ -568,8 +568,11 @@ static int php_uwsgi_startup(sapi_module_struct *sapi_module)
 	}
 }
 
+#if ((PHP_MAJOR_VERSION >= 7) && (PHP_MINOR_VERSION >= 1))
+static void sapi_uwsgi_log_message(char *message, int syslog_type_int) {
+#else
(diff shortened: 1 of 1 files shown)


unbit:
@xrmx safe for backport

xrmx:
backported, thanks @gdamjan. You may want to add php to the plugin smoke tested on travis since you care about it.

PR #1448, opened 2017-01-17
fix broken response when no_threads_wait option.

afiram:
in `docker alpine` environment,  with `--max-requests` option make hung up worker, and end up worker killed by uwsgi_master_check_mercy() .

```
Thu Dec 15 22:38:00 2016 - worker 1 (pid: 1186) is taking too much time to die...NO MERCY !!!
```

may be `pthread_cancel/pthrad_join` not work properly in (docker alpine) `musl libc`.

next, I try `--no-threads-wait` option, but this make response content corruption sometimes.

so, I tried use uwsgi_worker_is_busy() instead of pthread_join(). 

```
# run uwsgi with --no-threads-wait
$ make
$ ./uwsgi --http=0.0.0.0:8000 --static-check=. --master --threads=2 --max-requests=1 \
    --min-worker-lifetime=1 --worker-reload-mercy=10 --reload-mercy=10 --no-threads-wait
```

it seems to work fine.

@unbit please review.

### DIFF
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -1242,7 +1242,7 @@ void wait_for_threads() {
 			if (! is_busy){
 				break;
 			}
-			sleep(0.1);
+			sleep(0.001);
 		}
 		return;
 	}
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -1234,7 +1234,18 @@ void wait_for_threads() {
 	int i, ret;
 
 	// on some platform thread cancellation is REALLY flaky
-	if (uwsgi.no_threads_wait) return;
+	if (uwsgi.no_threads_wait){
+		int is_busy;
+		for(;;){
+			is_busy = uwsgi_worker_is_busy(uwsgi.mywid);
+			uwsgi_log("wait_for_threads worker_is_busy? %d mywid:%d mypid:%d\n", is_busy, uwsgi.mywid, uwsgi.mypid);
+			if (! is_busy){
+				break;
+			}
+			sleep(0.1);
+		}
+		return;
+	}
 
 	int sudden_death = 0;
 
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -1238,7 +1238,6 @@ void wait_for_threads() {
 		int is_busy;
 		for(;;){
 			is_busy = uwsgi_worker_is_busy(uwsgi.mywid);
-			uwsgi_log("wait_for_threads worker_is_busy? %d mywid:%d mypid:%d\n", is_busy, uwsgi.mywid, uwsgi.mypid);
 			if (! is_busy){
 				break;
 			}
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -1241,7 +1241,7 @@ void wait_for_threads() {
 			if (! is_busy){
 				break;
 			}
-			sleep(0.001);
+			sleep(1);
 		}
 		return;
 	}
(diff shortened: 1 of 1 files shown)


xrmx:
Any chance this can be discussed / fixed in musl?

afiram:
this is version info.

```
- docker for mac 1.12.6 (14937)
- docker-image-tag: python:3.5.2-alpine (inherit FROM alpine-release:3.4.6)
- musl-libc (x86_64) 1.1.14
- python:3.5.2
```

after create PR, I digging more.  then, I noticed hunging point is pthread_join() for [`initial thread`]( 
https://github.com/unbit/uwsgi/blob/f262899bc5eb890007ba157d8b1da1e836049b74/core/uwsgi.c#L1274).

so, I check function in small world but this never happened.
if they will occur in more complicated situation (mixed in fork, lock, signals and more),
discuss in musl is a little hard to me. I think core developer theam is more suitable, sorry.

anyway, point of this patch is fix to "no_threads_wait option breaking graceful reload".
this maybe help not only musl but some other platforms.

[comment from uwsgi.c](
https://github.com/unbit/uwsgi/blob/f262899bc5eb890007ba157d8b1da1e836049b74/core/uwsgi.c#L1236) say
```
    // on some platform thread cancellation is REALLY flaky
    if (uwsgi.no_threads_wait) return;
```

Thank you for your reply.

<details>
 <summary> short pthread_join() code </summary> 

```
cat p.c
#include <pthread.h>
#include <signal.h>
#include <stdio.h>
#include <string.h>
#include <unistd.h>

void * thread_func(void *arg)
{
    printf("Hello! I`m thread_func!\n");
    for(;;){
            sleep(1);
    }
    printf("I`m thread_func!\nGood-bye!\n");
    return NULL;
}

int main(void)
{
    int res;
    pthread_t thread_id;

    printf("Hello from main()!\n");
    pthread_create(&thread_id, NULL, thread_func, NULL);

    res = pthread_cancel(thread_id);
    printf("pthread_cencel() returned %d (%s)\n", res, strerror(res));
    res = pthread_join(thread_id, NULL);
    printf("pthread_join() returned %d (%s)\n", res, strerror(res));

    thread_id = pthread_self();
    res = pthread_cancel(thread_id);
    printf("self pthread_cencel() returned %d (%s)\n", res, strerror(res));
    res = pthread_join(thread_id, NULL);
    printf("sel pthread_join() returned %d (%s)\n", res, strerror(res));

    return 0;
}

$ gcc -lpthread -g p.c
$ ./a.out
Hello from main()!
pthread_cencel() returned 0 (No error information)
Hello! I`m thread_func!
pthread_join() returned 0 (No error information)
self pthread_cencel() returned 0 (No error information)
```

note. 
uwsgi.workers[].cores[0].thread_id is set pthread_self() id.

> uwsgi.workers[uwsgi.mywid].cores[0].thread_id = pthread_self();

https://github.com/unbit/uwsgi/blob/f262899bc5eb890007ba157d8b1da1e836049b74/core/uwsgi.c#L3632

maybe master process is require for sample program? 

</details>

bungoume:
@unbit Could you please merge this?
I am having the same issue.

It still occurs in [3.6.2-alpine3.6](https://hub.docker.com/_/python/
) docker image.

unbit:
It looks like pthread cancellation behave differently in every platform of the planet :) (well, i cannot even exclude that i misinterpreted the whole story here). By the way this patch is not needed in my common distros on linux x86 so it looks a bit invasive to me. What about adding an ad-hoc option for specifying the thread waiting strategy ? Something like, --threads-wait-by-spinning or something like that

afiram:
@unbit 
Sorry for the too late reply. I appreciate your reply.
but this PR is too old. I will close this. thanks.

PR #1457, opened 2017-01-27
Add the "License" classifier to setup.py

jdufresne:
It wasn't clear to me if uWSGI is GPL2 only or GPL2 or any later version. Can you clarify?

If the license includes any later version, the classifier should be:

License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -130,11 +130,12 @@ setup(
     description='The uWSGI server',
     author='Unbit',
     author_email='info@unbit.it',
-    license='GPL2',
+    license='GPLv2+',
     py_modules=['uwsgidecorators'],
     distclass=uWSGIDistribution,
     classifiers=[
         'Development Status :: 5 - Production/Stable',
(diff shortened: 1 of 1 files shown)


xrmx:
core/uwsgi.c says GPLv2+, @unbit care to clarify?

jdufresne:
Judging by the header of [uwsgi.c](https://github.com/unbit/uwsgi/blob/2.0.15/core/uwsgi.c#L7-L10) this should actually be GPLv2+. I have updated the line and rebased to resolve merge conflicts.

jdufresne:
It is also documented explicitly in the changelog that GPLv2+ should be used. So I updated the `license=` line as well:

https://uwsgi-docs.readthedocs.io/en/latest/Changelog-1.9.18.html#license-change

jdufresne:
Rebased to resolve merge conflicts. Travis is now green.

PR #1459, opened 2017-01-28
Fix invalid php session plugin behaviour

bartbes:
Make php session plugin return an empty string if the entry does not
exist.

As far as I can tell this has always been the right way, and php
accepting FAILURE was to support "broken" plugins. It seems that, at
least on my system, php 7.1 now requires correct behaviour.

### DIFF
--- a/plugins/php/session.c
+++ b/plugins/php/session.c
@@ -18,7 +18,7 @@ PS_READ_FUNC(uwsgi) {
 	char *value = uwsgi_cache_magic_get((char *)key, strlen((char *)key), &valsize, NULL, cache);
 #endif
 	if (!value) {
-		*val = ZSTR_EMPTY_ALLOC();
+		*val = STR_EMPTY_ALLOC();
 		return SUCCESS;
 	}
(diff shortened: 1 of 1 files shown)


bartbes:
Some more useful information, this behaviour can be reproduced using a php containing only `session_start()`, and symptoms include sessions not working at all. Looking at stats also shows the cache remains empty.

Here's an example log message:

`PHP Warning:  session_start(): Failed to read session data: uwsgi (path: sessions)`

With the patch I get no more warnings, and sessions work as normal.

xrmx:
@bartbes it looks like it doesn't build with the php 5 version we have on travis. Could you please fix it?

bartbes:
I knew I should have checked the php5 source...

xrmx:
@bartbes no worries, and thanks for the patch :)
@gdamjan care to review please?

gdamjan:
I have no knowledge about this, if it compiles on all phps all good

bartbes:
I can confirm #1646 fixes the same problem.

Issue #1466, opened 2017-02-06
Unable to build uwsgi with asyncio support in python3.x 


2mf:
Although that I refer to ticket #1100 and face the same problem with python 3.6.
I can't not reason with the comments of @kracekumar and @xrmx  regarding #1100 for just updating the docs....
The installation scripts should work in venv and vanilla. Just take one other example, you have a docker container with a custom compiled python and no system wide install of python present at all e.g. pyenv.

Please fix the install/compile scripts for Python 3.x and asyncio.

```
  Downloading uwsgi-2.0.14.tar.gz (788kB)
  Downloading from URL https://pypi.python.org/packages/11/ef/3d64655693e12f163f659bc3a45837893d7ebbb242e608feb9adc788d2ca/uwsgi-2.0.14.tar.gz#md5=d4b2c7db274dc20be54a977c9b040b7e (from https://pypi.python.org/simple/uwsgi/)
    100% |████████████████████████████████| 798kB 6.5MB/s 
  Running setup.py (path:/tmp/pip-build-b9hqp7vn/uwsgi/setup.py) egg_info for package uwsgi
    Running command python setup.py egg_info
    running egg_info
    creating pip-egg-info/uWSGI.egg-info
    writing pip-egg-info/uWSGI.egg-info/PKG-INFO
    writing dependency_links to pip-egg-info/uWSGI.egg-info/dependency_links.txt
    writing top-level names to pip-egg-info/uWSGI.egg-info/top_level.txt
    writing manifest file 'pip-egg-info/uWSGI.egg-info/SOURCES.txt'
    warning: manifest_maker: standard file '-c' not found

    reading manifest file 'pip-egg-info/uWSGI.egg-info/SOURCES.txt'
    writing manifest file 'pip-egg-info/uWSGI.egg-info/SOURCES.txt'
  Source in /tmp/pip-build-b9hqp7vn/uwsgi has version 2.0.14, which satisfies requirement uwsgi[asyncio] from https://pypi.python.org/packages/11/ef/3d64655693e12f163f659bc3a45837893d7ebbb242e608feb9adc788d2ca/uwsgi-2.0.14.tar.gz#md5=d4b2c7db274dc20be54a977c9b040b7e
  Installing extra requirements: 'asyncio'
  uWSGI 2.0.14 does not provide the extra 'asyncio'
Installing collected packages: uwsgi
[...]

Issue #1100, opened 2015-11-03
Unable to build uwsgi with asyncio support in python3.5


kracekumar:
```bash
➜  /tmp  python3.5 -m venv tmp
➜  /tmp  . tmp/bin/activate
(tmp) ➜  /tmp  pip install wheel
Collecting wheel
  Downloading wheel-0.26.0-py2.py3-none-any.whl (63kB)
    100% |████████████████████████████████| 65kB 192kB/s
Installing collected packages: wheel
Successfully installed wheel-0.26.0
(tmp) ➜  /tmp  pip install greenlet
Collecting greenlet
  Using cached greenlet-0.4.9.tar.gz
Building wheels for collected packages: greenlet
  Running setup.py bdist_wheel for greenlet
  Stored in directory: /Users/krace/Library/Caches/pip/wheels/08/9c/52/3d8ee766d654c43b131dcc954f255f8c6270bef617da0ac3e6
Successfully built greenlet
Installing collected packages: greenlet
Successfully installed greenlet-0.4.9
(tmp) ➜  /tmp  CFLAGS="-I/tmp/tmp/include/python3.5m" UWSGI_PROFILE="asyncio" pip install uwsgi --no-cache-dir
Collecting uwsgi
  Downloading uwsgi-2.0.11.2.tar.gz (782kB)
    100% |████████████████████████████████| 782kB 4.4MB/s
Installing collected packages: uwsgi
  Running setup.py install for uwsgi
    Complete output from command /private/tmp/tmp/bin/python3.5 -c "import setuptools, tokenize;__file__='/private/var/folders/xk/8lhz2_j91_3758cc65stj5z40000gn/T/pip-build-28wigc2x/uwsgi/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /var/folders/xk/8lhz2_j91_3758cc65stj5z40000gn/T/pip-urj0zxi1-record/install-record.txt --single-version-externally-managed --compile --install-headers /private/tmp/tmp/include/site/python3.5/uwsgi:
    running install
    plugins/greenlet/greenlet.c:2:10: fatal error: 'greenlet/greenlet.h' file not found
    #include <greenlet/greenlet.h>
             ^
    1 error generated.

    ----------------------------------------
[...]

PR #1467, opened 2017-02-07
fix include of "greenlet.h"

2mf:
Fix include `greenlet.h`
#1100 
#1466


xrmx:
Sorry, but I don't think we are going to bundle dependencies as git submodules.

2mf:
Please make a suggestion that is acceptable for you as a fix?

xrmx:
The fix would be more smart if the greenlet.h is on a position we can find out, if it's not then there's something off with how you are installing things and passing the relevant environment variable should be enough.

2mf:
So you are seriously saying, you want to depend on some system libs/headers that don't even exist in many systems even as installable package but has to be installed or put in some special place by hand.

The simplest solution to keep uwsgi installable via pip without special needs, is to either add the header file directly to the git repo or update the install script to pull the greenlet repo in case it is needed.

However, you wouldn't get a +1 if you need to read the documentations on why an install is not working or files are missing.

Please clarify that for me and others who face(d) the same thing.

Thank you.

xrmx:
Il 08/02/2017 12:59, mfranke ha scritto:
> So you are seriously saying, you want to depend on some system libs/headers that
> don't even exist in many systems even as installable package but has to be
> installed or put in some special place by hand.

On which systems greenlet is not packaged?

> The simplest solution to keep uwsgi installable via pip without special needs,
> is to either add the header file directly to the git repo or update the install
> script to pull the greenlet repo in case it is needed.

The dependency on greenlet looks good, but what would be nice to sort out is to 
get the virtualenv include path from python and not by constructing the path by 
hand. If that's not possible we can live with that though.

> However, you wouldn't get a +1 if you need to read the documentations on why an
> install is not working or files are missing.

I think a compiler error telling is not being able to find an header it's pretty 
self explanatory

You forgot to remove the externals dir from your PR.

2mf:
Sorry about the last comment, please forget about that :-)

I thought I removed that sub repo already...

2mf:
Ok I'll add a check for VIRTUAL_ENV but if not found, it might be good to default to sys.prefix construct

xrmx:
System python should be already be covered see *plugins/greenlet/uwsgiplugin.py*:
```
CFLAGS = [
    '-I' + sysconfig.get_python_inc(),
    '-I' + sysconfig.get_python_inc(plat_specific=True)
]
```

2mf:
according to PEP405 `sys.prefix` is the most reliable way, as of VIRTUAL_ENV can not be set if there was no activation of the venv.
So I think, I would leave it that way because gcc would not bother to have the include path a 2nd time.

2mf:
any other requirements for this pull request?

2mf:
btw. `sysconfig.get_python_inc()` do not point to right inculde directories e.g. the header of the installed greenlet package. So it is better to use `sys.prefix`.

unbit:
@xrmx seems good to to me, even that explicit check fro greenlet in setup.py is not bad as effectively the greenlet module is something python-specific (and setup.py is python-specific)

If you do not see other problems i will merge it

xrmx:
@unbit merged myself to be sure they got squashed :)

cicku:
so what if system has this library preinstalled?

Issue #1041, opened 2015-09-16
uwsgi.route segfaults with python3.4+


dakra:
I compiled uwsgi with python3.5 (I tried 3.4 as well) and every time I call
`uwsgi.route("uwsgi", "/tmp/foo,0,0")` from my app, uwsgi crashes.

```
 ./uwsgi --show-config --ini-paste-logged ~/atomx/api/uwsgi.ini 
[uWSGI] getting INI configuration from /home/daniel/atomx/api/uwsgi.ini

;uWSGI instance configuration
[uwsgi]
show-config = true
ini-paste-logged = /home/daniel/atomx/api/uwsgi.ini
http-socket = :6543
paste = config:/home/daniel/atomx/api/development.ini
paste-logger = /home/daniel/atomx/api/development.ini
enable-threads = true
virtualenv = /home/daniel/.virtualenvs/atomx35
offload-threads = 2
;end of configuration

*** Starting uWSGI 2.1-dev-59a76bb (64bit) on [Wed Sep 16 07:47:03 2015] ***
compiled with version: 5.2.0 on 16 September 2015 07:24:39
os: Linux-4.1.6-1-ARCH #1 SMP PREEMPT Mon Aug 17 08:52:28 CEST 2015
nodename: matrix
machine: x86_64
clock source: unix
pcre jit disabled
detected number of CPU cores: 8
current working directory: /home/daniel/github/uwsgi
detected binary path: /home/daniel/github/uwsgi/uwsgi
*** WARNING: you are running uWSGI without its master process manager ***
your processes number limit is 62938
your memory page size is 4096 bytes
detected max file descriptor number: 1024
lock engine: pthread robust mutexes
thunder lock: disabled (you can enable it with --thunder-lock)
uwsgi socket 0 bound to TCP address :6543 fd 4
Python version: 3.5.0 (default, Sep 15 2015, 16:09:16)  [GCC 5.2.0]
Set PythonHome to /home/daniel/.virtualenvs/atomx35
Python main interpreter initialized at 0x55c5994935a0
python threads support enabled
your server socket listen backlog is limited to 100 connections
your mercy for graceful operations on workers is 60 seconds
your request buffer size is 4096 bytes
mapped 72792 bytes (71 KB) for 1 cores
*** Operational MODE: single process ***
Loading paste environment: config:/home/daniel/atomx/api/development.ini
[...]

PR #1473, opened 2017-02-21
get uwsgi.route() working again from python

smason:
I was getting very similar crashes to that reported in #1041.  It crashes because `ur->condition_ub` and `ur->ovn` are `NULL` on entry to `uwsgi_routing_translate()`.

I'm presuming that these were supposed to be initialised somewhere else, but not sure where.  This was my first guess and appears to fix it for me!

### DIFF
--- a/core/routing.c
+++ b/core/routing.c
@@ -535,6 +535,7 @@ int uwsgi_route_api_func(struct wsgi_request *wsgi_req, char *router, char *args
 	return -1;
 found:
 	ur = uwsgi_calloc(sizeof(struct uwsgi_route));
+	uwsgi_fixup_routes(ur);
 	// initialize the virtual route
 	if (r->func(ur, args)) {
 		free(ur);


smason:
just to note, I was following along with the article on [Offloading Websockets and Server-Sent Events](http://uwsgi-docs.readthedocs.io/en/latest/articles/OffloadingWebsocketsAndSSE.html). 
 This crashed when I got to the section on [Simplifying things using the uwsgi api](http://uwsgi-docs.readthedocs.io/en/latest/articles/OffloadingWebsocketsAndSSE.html#simplifying-things-using-the-uwsgi-api-uwsgi-2-0-3).

Specifically the code that does:
```python
def subscribe(request):
    uwsgi.route("uwsgi", "/tmp/foo,0,0")
    return HttpResponse()
```
consistently crashes for me, inside the `route()` call.

unbit:
I have seen the same patch in your other pull request, just check the python_plugin.c part and i will absolutely merge it.

Thanks a lot

smason:
@unbit thanks for looking at this…  feel it's somewhat strange that nobody else has hit these issues, but guess that's life

I could have amended this pull-request rather than creating a new one.  would you like me to close this and make #1474 the main one?

coddingtonbear:
FWIW, @smason, I, too, ran into this issue.

PR #1474, opened 2017-02-22
Fixes for gevent after running with valgrind

smason:
My attempt at fixing reference errors when running with `gevent`.  I was experiencing strange crashes while running through the article on [uWSGI and SSE's](http://uwsgi-docs.readthedocs.io/en/latest/articles/OffloadingWebsocketsAndSSE.html).

The "real" change is in 3039e3501c0b294302970746dd74aa7320502802 and might not be correct, but the two calls to `Py_DECREF` look suspect to me.  I've arbitrarily removed one and a series of strange crashes stopped occuring for me.  Threading and gevent seems to be unsupported at the moment, hence saying there are two calls.  I don't know the code well enough to do any better with this at the moment!

Let me know if you think this is about right!

### DIFF
--- a/plugins/python/python_plugin.c
+++ b/plugins/python/python_plugin.c
@@ -1092,7 +1092,6 @@ void *uwsgi_python_create_env_holy(struct wsgi_request *wsgi_req, struct uwsgi_a
 }
 
 void uwsgi_python_destroy_env_holy(struct wsgi_request *wsgi_req) {
-	Py_DECREF((PyObject *)wsgi_req->async_environ);
 	Py_DECREF((PyObject *) wsgi_req->async_args);
 	// in non-multithread modes, we set uwsgi.env incrementing the refcount of the environ
 	if (uwsgi.threads < 2) {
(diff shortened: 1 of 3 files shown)


plugins/python/python_plugin.c
```diff
@@ -1092,7 +1092,6 @@ void *uwsgi_python_create_env_holy(struct wsgi_request *wsgi_req, struct uwsgi_a
 }
 
 void uwsgi_python_destroy_env_holy(struct wsgi_request *wsgi_req) {
-	Py_DECREF((PyObject *)wsgi_req->async_environ);
```
unbit: Not sure about this, in the holy mode we explicitely destroy the environment

unbit:
Thanks, see the comments about the python_plugin.c part

plugins/python/python_plugin.c
smason: yup, wasn't sure about this.  but there seem to be two `DECREF`s straight after each other

I don't know the code well enough to know where ownership belongs.  maybe if I put a small test together than can be run under different threading/async modes?

PR #1482, opened 2017-03-14
Add mule_send_msg success indicator

tiras-j:
This change is backwards compatible to existing mule uses and opens up the option for taking some action on failure to dispatch mule functions. 

Our particular use case involves offloading logging to a mule as python socket behavior can be quite... lethargic. However in some scenarios large amounts of logging information can be generated and we'd like to be able to decide how to respond to a full mule msg pipe rather than simply dropping the log data. 

Quick side note: I stuck with the
```c
Py_INCREF(Py_True);
return Py_True;
```
model for the python plugin, but is there any reason not to use the `Py_RETURN_TRUE` macros that encapsulate this behavior? Just wondering.

### DIFF
--- a/core/mule.c
+++ b/core/mule.c
@@ -12,7 +12,7 @@ extern struct uwsgi_server uwsgi;
 
 void uwsgi_mule_handler(void);
 
-void mule_send_msg(int fd, char *message, size_t len) {
+int mule_send_msg(int fd, char *message, size_t len) {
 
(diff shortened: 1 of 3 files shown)


unbit:
@xrmx safe for backport too

@tiras-j thanks a lot, regarding the Py_True returning, i started with it in 2007 (i do not remember if the macro was already there) then it remained as is for inertia :)

PR #1487, opened 2017-03-15
Make mule msg receive size configurable

tiras-j:
Adds the option `mule-msg-recv-size` to configure the maximum read size from the mule msg pipe. Unfortunately the previous `mule-msg-size` (defining the pipe buffer size) is slightly ambiguous, but what can you do. Also I've updated the `mule-msg-size` to be required_argument as I suspect getting a pipe buffer of 1 byte is *not* the expected behavior if someone forgets to define the size when using `--mule-msg-size`. Probably better to fail fast there, but I can take that out if backwards compatibility is a concern (who knows, maybe someone was depending on that 1 byte pipe size and can't be troubled to add an explicit `=1`?).

I've attached the receive buffer state to the mule's copy of the uwsgi global struct since it seemed most apt a location. If memory is more a concern than performance here we could simply re-allocate on each message read, that's an easy change just let me know.

### DIFF
--- a/core/mule.c
+++ b/core/mule.c
@@ -58,6 +58,8 @@ void uwsgi_mule(int id) {
 		uwsgi.mules[id - 1].pid = getpid();
 		uwsgi.mypid = uwsgi.mules[id - 1].pid;
 
+		uwsgi.mule_msg_recv_buf = uwsgi_malloc(uwsgi.mule_msg_recv_size);
+
 		uwsgi_fixup_fds(0, id, NULL);
 
 		uwsgi.my_signal_socket = uwsgi.mules[id - 1].signal_pipe[1];
@@ -170,8 +172,7 @@ void uwsgi_mule_handler() {
 	int rlen;
 	int interesting_fd;
 
-	// this must be configurable
-	char message[65536];
+	char *message = uwsgi.mule_msg_recv_buf;
 
 	int mule_queue = event_queue_init();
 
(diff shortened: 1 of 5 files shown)


uwsgi.h
```diff
@@ -2496,6 +2496,8 @@ struct uwsgi_server {
 	struct uwsgi_string_list *farms_list;
 	struct uwsgi_farm *farms;
 	int mule_msg_size;
+	int mule_msg_recv_size;
```
unbit: please move new structure fields at the end to maintain ABI compatibility

unbit:
The patch is ok for me, just move the struct fields at the end. Setting the buffer size to 1 seems an insane thing to do, so better to protect the user from the error here. Thanks

xrmx:
Friendly remainder to squash commits (at merge time is fine too)

tiras-j:
@xrmx Ah sorry about that, I'll be sure to remember in the future. I suppose here it'll be simpler to squash and merge.

xrmx:
@tiras-j no need to say sorry, multiple commits are nice for reviewing but are PITA to backport :)

core/mule.c
```diff
@@ -58,6 +58,10 @@ void uwsgi_mule(int id) {
 		uwsgi.mules[id - 1].pid = getpid();
 		uwsgi.mypid = uwsgi.mules[id - 1].pid;
 
+		if (!uwsgi.mule_msg_recv_size)
+			uwsgi.mule_msg_recv_size = 65536;
```
tiras-j: Would it make sense to place this assignment in `uwsgi_init_default`? That way the mule won't need to check whether the value is configured here and worker processes will have the default value (instead of 0) which is nice when querying the value from plugins etc.

tiras-j: Would it make sense to place this assignment in `uwsgi_init_default`? That way the mule won't need to check whether the value is configured here and worker processes will have the default value (instead of 0) which is nice when querying the value from plugins etc.

tiras-j:
I went ahead an moved the `uwsgi.mule_msg_recv_size` default into `uwsgi_init_default`. This way the mule doesn't need to check the value and configure it to the default when it's initialized. Furthermore it means workers querying the value (i.e. from the python plugin) will actually see a default value when it's not configured instead of simply 0.

Also went ahead and squashed :)

tiras-j:
@unbit Please let me know if there are any other changes I should make, or anything else I can do to help. Thanks!

PR #1498, opened 2017-03-23
plugins/redislog: support auth & select.

### LINKED ISSUE #1434: No option to pass AUTH in redislog plugin
While running uwsgi with redislog plugin defined, there is no option where  `AUTH` parameter can be specified for authentication with redis server if redis server is enabled with authentication.

`--logger redislog[:<host>,<command>,<prefix>]`

userproof:
Usage: logger = redislog[:[password@]host:port[/id][,cmd[,prefix]]]

fixes #1434


unbit:
Thanks

PR #1502, opened 2017-03-28
Make mule_msg_hook extensible

tiras-j:
This extends the python plugin mule msg hook to allow a list of functions which messages are dispatched to on the mule side, rather than a single function pointer (however it maintains the old pointer for compatibility).

The use case which prompted this change is that I needed a lazily configurable patch of a function which might offload to a mule based on consumed configuration files, thus I couldn't use a decorator directly. Furthermore this situation required serializing tuples of strings, which is tremendously faster with `marshal` than `cPickle`. To solve this issue I explicitly import `uwsgidecorators` to ensure installation of its mule msg hook, then overrode the hook myself, chaining the dispatch to the `uwsgidecorator` dispatch function in the event that the message was generated via `mulefunc` decorator etc.

However this is a rather delicate setup which requires careful import ordering. Instead allowing modules to install any function, which the plugin then dispatches to all hooks removes any such dependency/chaining. It's important that hooks use some method of disambiguation (i.e. serialization format + perhaps a service key like `uwsgidecorators` uses) however that's up to the developer to decide.

### DIFF
--- a/plugins/python/python_plugin.c
+++ b/plugins/python/python_plugin.c
@@ -1901,23 +1901,45 @@ int uwsgi_python_mule_msg(char *message, size_t len) {
 	UWSGI_GET_GIL;
 
 	PyObject *mule_msg_hook = PyDict_GetItemString(up.embedded_dict, "mule_msg_hook");
-        if (!mule_msg_hook) {
-                // ignore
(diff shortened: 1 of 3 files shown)


unbit:
Nice work, Thanks !

PR #1504, opened 2017-03-29
uwsgi python module: function to get wsgi_req->async_id

baverman:


unbit:
Hi, i think you can use environ['uwsgi.core'] (if you can easily access environ from your code)

baverman:
@unbit the whole idea is to be able to get request context at any time. But with environ['uwsgi.core'] an application should pass it through all frames. For example there is no easy way to get request context in low api levels to be able to log exception in Sentry with all details.

PR #1512, opened 2017-04-12
print without parentheses fix

talaikis:


xrmx:
Thanks for the patch but:
- please open a pr against master and not other branches
- import print_function from future
- do the same change to other files in the same pr

PR #1522, opened 2017-05-04
php: fix php-fallback-qs

unixwitch:
php-fallback-qs is broken; it chops off the last character from the query string.

I wrote this code originally, so that's my fault.  Sorry about that.


xrmx:
Any chance this can be tested? :)

unixwitch:
I can write a test, but I don't see any documentation on the test framework, and the contents of `tests/` is not enlightening.  (It looks like there aren't any PHP tests at all right now?)

Where would I start?

xrmx:
There's no documentation because the automatic test story is quite sad. As an example (in python though) you can take a look at the tests/ part of 24fd91e1000251fa44f0cfd630bcf3c867336264 or  7fb0aecaeb20d768010e3f1b7ce14bdd371f2102

PR #1524, opened 2017-05-04
Add http-path-info-no-decode-slashes option

comel:
The option disables encoded slash (%2F) decoding in PATH_INFO when using
HTTP(S) protocol. This is similar to AllowEncodedSlashes directive in Apache.


unbit:
Thanks

Issue #1534, opened 2017-05-17
Improper header checks in http plugin


funkybob:
http.c:191:        else if (uhttp.auto_gzip && !uwsgi_strnicmp("ACCEPT-ENCODING", 15, hh, keylen)) {
http.c:302:	else if (uhttp.auto_gzip && !uwsgi_strncmp("ACCEPT_ENCODING", 15, hh, keylen)) {

At least one of these is wrong.

One calls strnicmp, the other strncmp.

One has _ the other has -

I wonder if this is related to why I can never get http-auto-gzip to work.

unbit:
The right one is the 191 (we are managing the true header in case insensitive way, so the dash is the right char). Waiting for your pull request :)

funkybob:
Some clarity is needed on how headers are checked:

                else if (!uwsgi_strnicmp("SEC-WEBSOCKET-VERSION", 21, hh, keylen)) {
                else if (!uwsgi_strnicmp("SEC-WEBSOCKET-KEY", 17, hh, keylen)) {
        if (!uwsgi_strnicmp("CONTENT-LENGTH", 14, hh, keylen)) {
        else if (!uwsgi_strnicmp("TRANSFER_ENCODING", 17, hh, keylen)) {
        else if (uhttp.auto_gzip && !uwsgi_strnicmp("ACCEPT-ENCODING", 15, hh, keylen)) {
    if (uwsgi_strncmp("CONTENT_LENGTH", 14, hh, *keylen) && uwsgi_strncmp("CONTENT_TYPE", 12, hh, *keylen)) {

Can we assume that:
1) incoming headers are CGIified - that is, upper cased and '-' changed to '_'
2) outbound headers are all caps, and .... how are '-' and '_' handled?

If these are true, we can dispense with the case-insensitive compare [small speed up, maybe]

PR #1535, opened 2017-05-17
[#1534] Fix test for accept encoding

funkybob:


unbit:
Thanks, @xrmx safe for backport

gdamjan:
Looking at the `http_add_uwsgi_header` function, before the line that this commit changes, there are other uses of headers with underscores and with `uwsgi_strncmp()`.

The code is a bit hard to follow, but in `http_headers_parse`, http_headers_parse is used after something that's called http_header_to_cgi that normalizes the header names.

PR #1536, opened 2017-05-20
travis: bump ruby version for rack plugin

xrmx:
As rack build fails with ruby < 2.2.2. While at it simplify ruby
handling by using rvm.


xrmx:
Thanks for the review @StefanoOrdine !

PR #1541, opened 2017-05-29
Fix MULE MSG QUEUE IS FULL message hint

etataurov:
Error message for unsuccessful `mule_send_msg` had a confusing message(to tune `signal-bufsize`),
but the actual buffer used here is one tuned with `mule-msg-size` param


unbit:
Thanks

PR #1543, opened 2017-05-30
Port to python 3.5

ofosos:
Enable support for structured FrameSummary object as found in Python
3.5.

Warning: this will only work on Python 3.5 and later.


unbit:
Hi, thanks for your patch, it looks good and i am pretty sure that python3 support has been always broken, so even if only >=3.5 is supported, it will be more than what we support right now

unbit:
@xrmx safe for merge

xrmx:
Any reason why we can't use the  PyObject_GetAttrString in python2?

PR #1547, opened 2017-06-01
support k_minor has a _xxx suffix

pirDOL:
on my platform uname outputs 2.6.32_1-15-0-0, i think k_minor should be 32, but now 32_1 is extracted, which leads to int(k_minor) throws an exception


xrmx:
Merged, thanks!

PR #1548, opened 2017-06-04
Fix application loading without Plack after excluding "." from @INC in new Perl versions

jef-sure:
Without this change I had to add `PERL5OPT=-I.` to ini file.


unbit:
Thanks

Issue #1554, opened 2017-06-16
Emperor throttling logic confuses micro- and milliseconds


jeremy-hiatt:
According to the [uWSGI documentation](http://uwsgi-docs.readthedocs.io/en/latest/Options.html#emperor-throttle) the emperor-throttle option is specified in milliseconds. However the value to which it gets set is [passed directly to usleep](https://github.com/unbit/uwsgi/blob/master/core/emperor.c#L1101), which expects a value in microseconds!

Passing the desired value in microseconds causes problems with the blacklisting logic, as the throttle is assumed to be in millis and is multiplied by 1000. Thus if you've set the throttle level to 1 second (i.e. `emperor-throttle = 1000000`), your bad-behaving vassal won't be restarted until 1000 seconds later! Note that setting `emperor-max-throttle` does not prevent this from happening. One available workaround is to simply disable blacklisting altogether with the `emperor-no-blacklist` flag/option.

xrmx:
Good catch! I think that adding a `* 1000` to to both the usleep should do the trick

PR #1557, opened 2017-06-17
core/emperor: proper conversion to microseconds for usleep()

jeremy-hiatt:
Proposed fix for #1554 

I verified that throttling seems to work correctly now.


xrmx:
Thanks a lot!

PR #1561, opened 2017-06-24
Allow to override build date

bmwiedemann:
to allow for reproducible builds of uwsgi

See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.


xrmx:
You are also changing the time stored from local to utc though.

bmwiedemann:
That is intentional as it helps to get the same results independent of timezone

xrmx:
Sure and i like the change, was just noticing the behaviour change :)

xrmx:
@unbit is the timezone change a problem for you? If not let's merge this!

Issue #1558, opened 2017-06-18
Store verified TLS Peer Certificates in a wsgi `environ` in the Python plugin


paultag:
Filing this issue to get some feedback and feel out how receptive the uwsgi community is to such a change.

It would be great for any verified `x.509` TLS Peer Certifciates that have been validated to be passed into Python through the `wsgi` `environ` object, likely in a vendor-specific `uwsgi.` name. In a perfect world, this would also include the entire chain.

It would make the most sense to pass them back in DER format, for Python to parse as needed. This will allow the Python code to specify more particular checks (e.g. for a custom attribute like a `UPN`, `SAN` or even something totally custom.

I know most folks do this by extracting `x.509` attributes and setting them as `X-*` headers, but this always struck me as fragile, and depends on the underlying software doing the right thing, or understanding custom extensions.

If I remember my OpenSSL correctly, this is something you can pull from the `SSL_CTX`, but a quick grep makes me think the TLS support was so far abstracted that the Python code doesn't know about the `SSL_CTX` at all. Does anyone have a gut check on how hard this would be to implement?

paultag:
Thanks to @funkybob on IRC, `./plugins/http/https.c` seems to be storing stuff using `uwsgi_buffer_append_keyval` - is this something you can read from Python running with `wsgi`? If so, dumping the raw DER at that point could work

unbit:
@paultag Hi, the @funkybob approach would work only if you have to uWSGI http router on front. It would not work on any other webserver/proxy. If it is enough for you can add the required fields in this part of the code: https://github.com/unbit/uwsgi/blob/master/plugins/http/https.c#L180 (where we currently add a minimal amount of ssl infos)

paultag:
I hacked up a proof of concept in https://github.com/unbit/uwsgi/pull/1562 - this works well enough for my usecase!

pinodeca:
Hi, thanks for this feature!

However, I'm having trouble using it. request.environ['HTTPS_CLIENT_CERTIFICATE'] in my Flask application is correctly populated, but I'm not able to de-serialize/load it.

I've tried with both cryptography.x509.load_der_x509_certificate(cert_bytess,  default_backend()) and OpenSSL.crypto.load_certificate(OpenSSL.crypto.FILETYPE_ASN1, cert_bytes). Both complain about the format.

[...]

PR #1562, opened 2017-06-25
Write x509 DER to the uwsgi buffer

paultag:
This will write the full x.509 DER into the buffer for use by clients
during runtime. This feature is intended to allow clients to handle
per-user ACL with the direct x.509 Certificate, without having to
configure the webserver to extract the right bits, which may or may not
be custom extensions.

One such example would be using and extracting the UPN SAN, or some
other exotic extension.

### DIFF
--- a/plugins/http/https.c
+++ b/plugins/http/https.c
@@ -179,6 +179,12 @@ int hr_https_add_vars(struct http_session *hr, struct corerouter_peer *peer, str
 #endif
                 hr->ssl_client_cert = SSL_get_peer_certificate(hr->ssl);
                 if (hr->ssl_client_cert) {
+                        int client_cert_len;
+                        unsigned char *client_cert_der = NULL;
+                        client_cert_len = i2d_X509(hr->ssl_client_cert, &client_cert_der);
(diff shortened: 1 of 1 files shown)


paultag:
Example usage:

```
from cryptography.hazmat.backends import default_backend


    start_response("200 OK", [
        ("Content-Type", "text/plain"),
    ])
    cert = x509.load_der_x509_certificate(environ['HTTPS_CLIENT_CERTIFICATE'], default_backend())
    yield ", ".join(("{0}={1}".format(attr.oid._name, attr.value) for attr in cert.subject))
```

With the server invocation looking like:

```
./uwsgi --wsgi-file wsgi.py --processes 1 --workers 1 --https ':8000,server.crt,server.key,,!ca.crt'
```

paultag:
This will close #1558 if merged

unbit:
@xrms safe for backport

@paultag nice work, thanks

unbit:
Note: this patch had a leak, the fix is here:

https://github.com/unbit/uwsgi/commit/86a3f5a341438bb3547cf4bb752cde1ac8626f32

@xrmx patch already in branch 2.0

PR #1564, opened 2017-06-25
Add a --ssl-verify-depth flag

paultag:
Add ssl-verify-depth flag to set the max Client CA chain length

When verifying Certificates, OpenSSL will build a chain of trust from a
root CA to the Certificate, via intermediary CA(s). The chain was
hardcoded to 1 previously, which means that all Client Certificates
would have to be issued directly from a root CA.

This adds a new flag, --ssl-verify-depth with a default of 1 (existing
behavior) that will allow N intermediary CAs when building the chain of
trust.

### DIFF
--- a/core/ssl.c
+++ b/core/ssl.c
@@ -326,8 +326,7 @@ SSL_CTX *uwsgi_ssl_new_server_context(char *name, char *crt, char *key, char *ci
                 else {
                         SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, uwsgi_ssl_verify_callback);
                 }
-                // in the future we should allow to set the verify depth
-                SSL_CTX_set_verify_depth(ctx, 1);
+                SSL_CTX_set_verify_depth(ctx, uwsgi.ssl_verify_depth);
 
 		if (uwsgi.ssl_tmp_dir && !uwsgi_starts_with(client_ca, strlen(client_ca), "-----BEGIN ", 11)) {
 			if (!name) {
--- a/uwsgi.h
+++ b/uwsgi.h
@@ -2892,6 +2892,10 @@ struct uwsgi_server {
 	char *mule_msg_recv_buf;
 
 	int http_path_info_no_decode_slashes;
+
+#ifdef UWSGI_SSL
+	int ssl_verify_depth;
+#endif
 };
 
 struct uwsgi_rpc {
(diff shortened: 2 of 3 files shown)


paultag:
This is an easy one; mind giving me a review here, @unbit?

paultag:
Any feedback, @unbit? Things to fix?

unbit:
Is not better to make it configurable ?

paultag:
@unbit Yeah, I agree -- I didn't do it here because I wasn't sure what the correct way to change that in a non-invasive way. I can start working on that if that's a hard requirement -- I just didn't want to create an invasive diff changing function signatures without understanding more about uwsgi style and idioms.

I figured a short-term change from a const to another const that's slightly higher would have been less debt, but I can obviously work to correct this to be configurable.

Which do you prefer? I'm happy to start to fix this up, but it'll likely result in a bigger diff with more to review, and maybe a few rounds of feedback.

paultag:
@unbit ping ^

unbit:
I prefer adding the new option, it is pretty easy to add new option in uWSGI (this is why we have thousands of them :P)

paultag:
@unbit OK, I added a flag. This depends on #1603, since I didn't want to ignore compile warnings. I'm happy to split it out if that PR isn't wanted

core/ssl.c
```diff
@@ -326,8 +326,7 @@ SSL_CTX *uwsgi_ssl_new_server_context(char *name, char *crt, char *key, char *ci
                 else {
                         SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, uwsgi_ssl_verify_callback);
                 }
-                // in the future we should allow to set the verify depth
```
paultag: THE FUTURE IS NOW! :)

paultag:
(Rebased, thanks for the merge on the other PR!)

paultag:
@unbit any thoughts? It works well locally here and I think it's pretty minimal :)

unbit:
check my review comment about the ABI  :)

paultag:
@unbit did you click "finish review"? I don't see it  yet :)

unbit:
ops :P

uwsgi.h
```diff
@@ -2717,6 +2717,7 @@ struct uwsgi_server {
 #ifdef UWSGI_SSL
 	int ssl_initialized;
 	int ssl_verbose;
+	int ssl_verify_depth;
```
paultag: :+1: yeah, totally. I just assumed because it was used in the flags that it was part of the CLI and not library that grouping args would make more sense, but on second thought that makes total sense. I even see other SSL flags throughout the struct - sorry for not picking up on that!

paultag:
@unbit I do it all the time myself :)

Push'd a fix

unbit:
@xrmx safe for backport

PR #1570, opened 2017-07-03
plugin/http: Fix compilation

melvyn-sopacua:
Fix compilation when OpenSSL is compiled without NPN support. The flags
for NPN are always defined, but OPENSSL_NO_NEXTPROTONEG is defined if
the npn support is disabled at OpenSSL compile time.

This is the default for FreeBSD OpenSSL 1.0+ via ports.

### DIFF
--- a/plugins/http/common.h
+++ b/plugins/http/common.h
@@ -5,7 +5,7 @@ extern struct uwsgi_server uwsgi;
 #include "../corerouter/cr.h"
 
 #ifdef UWSGI_SSL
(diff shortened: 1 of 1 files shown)


melvyn-sopacua:
The reason for the double check is that versions before 1.0 have no knowledge of either definition. So just checking for *absence* of `OPENSSL_NO_NEXTPROTONEG` would yield the wrong behavior.

unbit:

Issue #1578, opened 2017-07-10
stats_pusher_statsd options struct missing END_OF_OPTIONS


thinkski:
This causes option parsing to occasionally work, if memory immediately following struct happens to be zeroed, but can spuriously fail.

PR #1579, opened 2017-07-10
Add END_OF_OPTIONS to stats_pusher_statsd struct

thinkski:
Struct was missing terminating END_OF_OPTIONS, causing options
parsing to fail if memory following struct was not zero. Addresses #1578.


unbit:
Thanks a lot. @xrmx safe for backport

xrmx:
We don't have that code in uwsgi-2.0 :)

PR #1589, opened 2017-07-21
Remove uwsgi.lazy as it's replaced

funkybob:
Since the docs firmly recommend against using --lazy, and --lazy-apps is working in its place, this drops the support/use of the --lazy flag.


funkybob:
There's a function in core/uwsgi.c that, besides the typo, may be simplified, but I wasn't sure:

```
// brutally reload
void reap_them_all(int signum) {

    // avoid reace condition in lazy mode
    if (uwsgi_instance_is_reloading)
        return;
```

It wasn't clear if this was for lazy, or lazy-apps.

xrmx:
By removing a field in the struct you are changing the ABI.

funkybob:
I've added the lazy flag back, but updated the commend to indicate it's deprecated.

PR #1592, opened 2017-07-26
Add worker running time metrics.

yasek:
We also needed average response time as arithmetic mean as was mentioned in #730 
Having total_running_time allows to calculate even more statistics like average cluster response time, percentiles and so on.


unbit:
Thanks !

@xrmx safe for backport

unbit:
backported to 2.0

Issue #1593, opened 2017-07-27
Using a hostname with Redis Logger


ahmbas:
I am using AWS elasticache Redis.
Using the following I can see logs coming into Redis

(Using IP address)
```
logger     =  redislog:172.30.31.232:6379
```

```
test-uwsgi.3o8ah4.0001.euw1.cache.amazonaws.com:6379> MONITOR
OK

1501156479.639613 [0 172.30.31.232:53396] "select" "0"
1501156479.639783 [0 172.30.31.232:53396] "publish" "uwsgi" "{\"app_name\": \"testo\", \"message\": \"*** Starting uWSGI 2.0.15 (64bit) on [Thu Jul 27 15:54:39 2017] ***\"}"
1501156479.639945 [0 172.30.31.232:53396] "publish" "uwsgi" "{\"app_name\": \"testo\", \"message\": \"compiled with version: 5.4.0 20160609 on 27 July 2017 15:24:01\"}"
```

But if I use the DNS entry provided by AWS test-uwsgi.3o8ah4.0001.euw1.cache.amazonaws.com
(Using hostname)
```logger     =  redislog:test-uwsgi.3o8ah4.0001.euw1.cache.amazonaws.com:6379```
Nothing shows up in redis.
```
172.30.31.232:6379> MONITOR
OK
```

Nslookup returns the IP normally
```
Non-authoritative answer:
Name:	test-uwsgi.3o8ah4.0001.euw1.cache.amazonaws.com
Address: 172.30.31.232
```
It seems like the redis logger only works with IPs.

This can be problematic if you are writing to a load balanced redis cluster or a redis instance with an A record.
Uncovered this when I was trying to write uwsgi logs to AWS elasticache which has a DNS entry

**Comment by mudasirmirza** [comment]
+1

**Comment by unbit** [comment]
Hi, you can add a call to uwsgi_resolve() here:

https://github.com/unbit/uwsgi/blob/master/plugins/redislog/redislog_plugin.c#L99

```c
uredislog->address = uwsgi_resolve_ip(logarg)
```

i have not tried, but it should work

waiting for your pull request :)

PR #1595, opened 2017-07-30
Added uwsgi_resolve_ip for redis host

ahmbas:
@unbit  https://github.com/unbit/uwsgi/issues/1593
Would appreciate if you take a look :)
Tested it locally and it works :+1:


unbit:
Thanks :)

PR #1596, opened 2017-08-01
Cheaping multiple workers command should be respected

shivawu:
When the master receives multiple cheaping command, like `---`, it should be able to try its best to cheap 3 workers, not only 1.

This also adds an additional check for worker aliveness after thunder lock to increase the effectiveness of worker cheaping.


shivawu:
Superseded by #1598

PR #1597, opened 2017-08-01
Cheap worker uses SIGHUP

shivawu:
We used SIGWINCH to cheap a worker, by waking it up and making itself realize it should go away. 

But in certain cases like thunder lock, a worker could be blocked by the thunder lock instead of the `accept()` syscall, and many lock engine's acquire operation is not interruptible, which means in this case, the cheaped worker will hang until SIGKILLed. 

Since the change that we only cheap idle workers after 2.0.14, it should be fine to just send SIGHUP to make the worker go away directly. If there's concerns around this, we could make this an option.

### DIFF
--- a/core/master_utils.c
+++ b/core/master_utils.c
@@ -206,7 +206,12 @@ safe:
 			uwsgi.workers[oldest_worker].rss_size = 0;
 			uwsgi.workers[oldest_worker].vsz_size = 0;
 			uwsgi.workers[oldest_worker].manage_next_request = 0;
-			uwsgi_curse(oldest_worker, SIGWINCH);
+			// Send SIGHUP to shutdown the worker. SIGWINCH doesn't work
+			// with workers that block on thunder lock (either pthread
+			// mutex or ipcsem, which uWSGI retries after EINTR).
+			// Since we're only choosing idle workers, it's fine to
+			// immediately shutdown with SIGHUP (end_me).
+			uwsgi_curse(oldest_worker, SIGHUP);
 		}
 	}
 


unbit:
Hi, thanks again for your patch. Yes, i would prefer an option to change this behaviour as it is common to override signal handlers in app code (it happened we did it too for specific customers need). Albeit under thunder-lock this is absolutely a good choice, it could break other areas when no thunder-lock is in place.

shivawu:
So in that case, do you think it makes sense to say "if thunder-lock enabled then SIGHUP else SIGWINCH", or a separate option would be better. I'm not viewing from the full picture here so there might be other usecases I'm missing that SIGHUP instead of SIGWINCH itself is useful standalone.

shivawu:
Ping?

unbit:
Hi, unfortunately i cannot know exactly how many signal 'mess' users are doing, so i would prefer an explicit option. Thanks again

core/master_utils.c
```diff
@@ -206,7 +206,12 @@ int uwsgi_calc_cheaper(void) {
 			uwsgi.workers[oldest_worker].rss_size = 0;
 			uwsgi.workers[oldest_worker].vsz_size = 0;
 			uwsgi.workers[oldest_worker].manage_next_request = 0;
-			uwsgi_curse(oldest_worker, SIGWINCH);
+			// Send SIGHUP to shutdown the worker. SIGWINCH doesn't work
+			// with workers that block on thunder lock (either pthread
+			// mutex or ipcsem, which uWSGI retries after EINTR).
+			// Since we're only choosing idle workers, it's fine to
+			// immediately shutdown with SIGHUP (end_me).
+			uwsgi_curse(oldest_worker, SIGHUP);
```
funkybob: Perhaps make the choice of signal contingent on if thunderlock (or a vulnerable lock engine) is in use?

PR #1598, opened 2017-08-01
Correctly handle multiple cheaping command

shivawu:
When the master process receives multiple cheaping command, like `---`, currently it only tries to close one single worker, which is not what it's supposed to do.

In this PR, we change it to:
- Start a loop to try to cheap as many as required
- If not all of them can be satisfied, preserve the left over commands to the next master loop
- Add an optimization to improve the efficiency of worker aliveness check after thunder lock

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -1527,6 +1527,16 @@ int wsgi_req_accept(int queue, struct wsgi_request *wsgi_req) {
 
 	thunder_lock;
 
+	// Recheck the manage_next_request before going forward.
+	// This is because the worker might get cheaped while it's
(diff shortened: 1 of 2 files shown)


core/utils.c
```diff
@@ -1527,6 +1527,16 @@ int wsgi_req_accept(int queue, struct wsgi_request *wsgi_req) {
 
 	thunder_lock;
 
+	// Recheck the manage_next_request before going forward.
```
unbit: nice catch !

unbit:
@shivawu Thanks, great work

@xrmx safe for backport

PR #1610, opened 2017-08-14
Add explicit breaks to avoid implicit passthrough.

paultag:
In addition to the changes in 0a14d0f0425f00421a69f0ca8e09a92cfdfc6a36, there were some additional case statements I missed because of my build configuration on my travel laptop.


unbit:
Thanks again :)

paultag:
@unbit totes, it's a pleasure!

PR #1614, opened 2017-08-18
Add --http-subscription-fallback-key and --fastrouter-subscription-fallback-key

funkybob:
When using a subscription server with fastrouter or http, and no subscribed nodes match, try the fallback key.

### DIFF
--- a/plugins/corerouter/cr.h
+++ b/plugins/corerouter/cr.h
@@ -293,6 +293,9 @@ struct uwsgi_corerouter {
 	socklen_t emperor_socket_addr_len;
 
 	int defer_connect_timeout;
+
+	char *fallback_key;
+	int fallback_key_len;
 };
 
 // a session is started when a client connect to the router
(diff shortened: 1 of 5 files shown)


plugins/corerouter/cr.h
```diff
@@ -293,6 +293,8 @@ struct uwsgi_corerouter {
 	socklen_t emperor_socket_addr_len;
 
 	int defer_connect_timeout;
+
+    char *http_subscription_fallback_key;
```
xrmx: tabs please

xrmx:
Would be cool to add all the context you write in the GH PR to the commit message so it is stored in git :)

plugins/corerouter/cr.h
funkybob: Gah, one slipped through!

funkybob:
Can we make it a trifecta of commits this afternoon? :)

unbit:
@xrmx safe for backport

PR #1621, opened 2017-08-26
[Fixes #1619] Stop putting env in embeeded dict.

### LINKED ISSUE #1619: Fatal Python error: deletion of interned string failed
I've hit this twice today, and can't nail down the issue [which I assume to be somewhat race-conditiony]
```
Fatal Python error: deletion of interned string failed
KeyError: 'wsgi.file_wrapper'

Current thread 0x00007f239f5e1740 (most recent call first):
DAMN ! worker 1 (pid: 11080) died, killed by signal 6 :( trying respawn ...
Respawned uWSGI worker 1 (new pid: 11084)
```
This is against current master 826bb8f35f4a0d6de9fc113c05b28402dcff3c2b
```
*** Starting uWSGI 2.1-dev-826bb8f3 (64bit) on [Fri Aug 25 15:19:27 2017] ***
compiled with version: 7.2.0 on 25 August 2017 02:06:26
os: Linux-4.12.0-1-amd64 #1 SMP Debian 4.12.6-1 (2017-08-12)
nodename: lazarus
machine: x86_64
Python version: 3.6.2 (default, Aug  4 2017, 14:35:04)  [GCC 6.4.0 20170724]
```

The key error has been different each time, but always on wsgi.

According to https://mail.python.org/pipermail/python-list/2004-April/258468.html it's a ref-counting issue.

Uwsgi config:
```
[uwsgi]
strict = true
master=true

http= :5000
http-auto-chunked=true
http-auto-gzip=true
http-keepalive=true

static-map=/static=%dapp/static/
static-gzip-all=true
module= manage:app
venv=%dvenv/
pp=%d

# Serve a logo from the specified physical path.
route = ^/client_logo static:%dtemplates/client_logos/$(CLIENT_LOGO)
route = ^/client_logo static:%dapp/static/images/logo.png

route-run = delheader:Content-Length

collect-header = Content-Type RESPONSE_CONTENT_TYPE
response-route-if = equal:${RESPONSE_CONTENT_TYPE};application/json addheader:uWSGI-Encoding: gzip
response-route-if = startswith:${RESPONSE_CONTENT_TYPE};text/html addheader:uWSGI-Encoding: gzip

enable-threads = 1
offload-threads = 1
```

funkybob:


funkybob:
Found a proper fix.

PR #1625, opened 2017-09-02
Alternative to marking explicit fallthrough.

funkybob:
GCC supports comments to mark this, as documented in:
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

This PR is only presented as illustrative of an alternative to an already accepted patch.

### DIFF
--- a/core/hash.c
+++ b/core/hash.c
@@ -42,14 +42,14 @@ static uint32_t murmur2_hash(char *key, uint64_t keylen) {
 	switch (keylen) {
 		case 3:
         		h ^= key[2] << 16;
-        		break;
+			/* FALLTHRU */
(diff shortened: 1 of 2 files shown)


xrmx:
We have already used that technique, see https://github.com/unbit/uwsgi/commit/043114964fd86f52cc9ac18d833cad067408476f

Looking at the code again the breaks seems wrong, as we want the pass through. That's actually a bugfix, good catch! Care to open a PR against master, do two separated comments referencing each the commit where the breaks have been introduced and use /* fall through */ ?

funkybob:
Oh, did you mean a single PR with each file fixed in a separate commit?

PR #1628, opened 2017-09-02
Update travis config for trusty

funkybob:
Since travis is now defaulting to using Trusty, certain versions have updated.


xrmx:
Cherry-picked the ruby version update and forced precise for now until we sort out the other errors.

PR #1637, opened 2017-09-14
Make --http-keepalive require an argument.

funkybob:
This flag accepts a timeout value, but is typically mistaken for a bool.
Update the description to help clarify, and require the option value.


xrmx:
Uhm it looks like there's places where it is used as boolean because `uwsgi_opt_set_int` will set it a 1 if not value is passed.
```
plugins/http/http.c:                        if (uhttp.keepalive && !uwsgi_strncmp("HTTP/1.1", 8, base, ptr-base)) {
plugins/http/http.c:                    if (uhttp.keepalive && !uwsgi_strncmp("HTTP/1.1", 8, base, ptr-base)) {
plugins/http/http.c:                    if (uhttp.keepalive > 1) {
plugins/http/http.c:                            http_set_timeout(peer->session->main_peer, uhttp.keepalive);
```

funkybob:
So perhaps I should drop this, and just work on improving the docs?

BTW - what _unit_ is that timeout in? I assumed seconds, but, well...

karmux:
Merge?
Boolean checks should work as expected after this change.
In another PR they can be changed to `if (uhttp.keepalive > 0)`.
Related to #1245.

PR #1640, opened 2017-09-16
Allow setting the response headers buffer size.

funkybob:
Default size is system page size. This should allow for setting larger/smaller.

### DIFF
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -1009,6 +1009,7 @@ static struct uwsgi_option uwsgi_base_options[] = {
 	{"config-py", no_argument, 0, "dump the uwsgiconfig.py used for building the core  (useful for building external plugins)", uwsgi_opt_config_py, NULL, UWSGI_OPT_IMMEDIATE},
 	{"build-plugin", required_argument, 0, "build a uWSGI plugin for the current binary", uwsgi_opt_build_plugin, NULL, UWSGI_OPT_IMMEDIATE},
 	{"version", no_argument, 0, "print uWSGI version", uwsgi_opt_print, UWSGI_VERSION, 0},
+	{"response-headers-limit", required_argument, 0, "set response header maximum size (default: 64k)", uwsgi_opt_set_int, &uwsgi.response_header_limit, 0},
 	{0, 0, 0, 0, 0, 0, 0}
 };
 
@@ -2156,6 +2157,9 @@ void uwsgi_setup(int argc, char *argv[], char *envp[]) {
 	uwsgi.page_size = getpagesize();
 #endif
 	uwsgi.binary_path = uwsgi_get_binary_path(argv[0]);
+
+	if(uwsgi.response_header_limit == 0)
+		uwsgi.response_header_limit = UMAX16;
 
 	// ok we can now safely play with argv and environ
 	fixup_argv_and_environ(argc, argv, environ, envp);
--- a/core/writer.c
+++ b/core/writer.c
@@ -71,7 +71,7 @@ int uwsgi_response_prepare_headers(struct wsgi_request *wsgi_req, char *status, 
 
 	if (!wsgi_req->headers) {
 		wsgi_req->headers = uwsgi_buffer_new(uwsgi.page_size);
-		wsgi_req->headers->limit = UMAX16;
+		wsgi_req->headers->limit = uwsgi.response_header_limit;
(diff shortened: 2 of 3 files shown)


april:
Thank you so much! Without this setting, part of my user-agent testing website is broken.  :)

core/uwsgi.c
```diff
@@ -1009,6 +1009,7 @@ static struct uwsgi_option uwsgi_base_options[] = {
 	{"config-py", no_argument, 0, "dump the uwsgiconfig.py used for building the core  (useful for building external plugins)", uwsgi_opt_config_py, NULL, UWSGI_OPT_IMMEDIATE},
 	{"build-plugin", required_argument, 0, "build a uWSGI plugin for the current binary", uwsgi_opt_build_plugin, NULL, UWSGI_OPT_IMMEDIATE},
 	{"version", no_argument, 0, "print uWSGI version", uwsgi_opt_print, UWSGI_VERSION, 0},
+	{"response-headers", required_argument, 0, "Set response head maximum size (Default is page size)", uwsgi_opt_set_int, &uwsgi.response_header_size, 0},
```
april: This should probably be `--response-headers-buffer-size`, to be consistent with the other options.  You may also want to word it like the others, being `set response header maximum size (default: page size)`.

april:
Note that you can see this branch currently running, using this option:
http://misbehaving.site/too-many-headers/16384

core/uwsgi.c
```diff
@@ -2157,7 +2158,10 @@ void uwsgi_setup(int argc, char *argv[], char *envp[]) {
 #endif
 	uwsgi.binary_path = uwsgi_get_binary_path(argv[0]);
 
-	// ok we can now safely play with argv and environ
+	if(uwsgi.response_header_size == 0)
```
xrmx: missing space

core/uwsgi.c
```diff
@@ -2157,7 +2158,10 @@ void uwsgi_setup(int argc, char *argv[], char *envp[]) {
...
+		uwsgi.response_header_size = uwsgi.page_size;
+
+		// ok we can now safely play with argv and environ
```
xrmx: indentation is off

xrmx:
@april  How much space do you need?

@funkybob is there a reason the PR is againt 2.0 and not master?

core/uwsgi.c
funkybob: Will have to blame that on VSCode...

funkybob:
@xrmx It's against 2.0 because April needed it, and was using 2.0.

april:
I'm not sure what you mean by how much space do I need?  It currently allows users to choose how many headers they want, and the setting I'm actively using is `--response-headers 8388608`.

xrmx:
Was wondering if a higher default would do but no, 8MB is quite a bit far from 4K.

april:
In the case of 16384 headers, it sends about 512KB of headers. So that's a bit more reasonable, but it would be nice to be able to go higher (since many user agents allow an even greater number of headers).

unbit:
Sorry for the late, but unfortunately the patch is conceptually wrong (albeit it should be easy to fix).

the uwsgi_buffer subsystem dynamically grows and shrink, your headers are limited because of this part:

```c
wsgi_req->headers->limit = UMAX16
```

just allow UMAX16 to be configurable and then i can merge for sure

Thanks again

funkybob:
Yeah, I had seen before that buffers were auto-growing, so I found it odd that headers explicitly ran out, hence opting for this solution. If I'd just looked one line further :)

I've renamed the option to "response-header-limit".

unbit:
Thanks, cherry-picked to master too

core/uwsgi.c
```diff
@@ -2157,6 +2158,9 @@ void uwsgi_setup(int argc, char *argv[], char *envp[]) {
 #endif
 	uwsgi.binary_path = uwsgi_get_binary_path(argv[0]);
 
+	if(uwsgi.response_header_limit == 0)
```
xrmx: you really have an adversion for spaces :)

funkybob: I'm a Python coder, I love spaces :)  I just don't know your formatting rules - yet :)

xrmx: I'm pretty sure you add spaces after ifs in python

PR #1642, opened 2017-09-18
[Fixes #1639] Log error if python home does not exist.

### LINKED ISSUE #1639: Error logs should inform you when the python "home" does not exist
When using emperor tyrant, if for some reason python 'home' is not correct, the instances die and you get no clue of what can be happening. You only get messages about killing or discarding cursed instances in the emperor uwsgi log, but the "slave" log doesn't even get created.

funkybob:
Will fail loading of Python plugin.

### DIFF
--- a/plugins/python/python_plugin.c
+++ b/plugins/python/python_plugin.c
@@ -242,6 +242,10 @@ int uwsgi_python_init() {
 	}
 
 	if (up.home != NULL) {
+		if (!uwsgi_is_dir(up.home)) {
+			uwsgi_log("Python Home is not a directory: %s\n", up.home);
+			exit(1);
+		}
 #ifdef PYTHREE
 		// check for PEP 405 virtualenv (starting from python 3.3)
 		char *pep405_env = uwsgi_concat2(up.home, "/pyvenv.cfg");


plugins/python/python_plugin.c
```diff
@@ -242,6 +242,10 @@ int uwsgi_python_init() {
 	}
 
 	if (up.home != NULL) {
+		if(!uwsgi_is_dir(up.home)) {
```
xrmx: space between if and parens :)

plugins/python/python_plugin.c
```diff
@@ -242,6 +242,10 @@ int uwsgi_python_init() {
...
+			uwsgi_log("Python Home is not a directory: %s\n", up.home);
+			return 0;
```
xrmx: I'm not sure anyone is looking at the plugin init return code, not sure if we can make this an exit(1)

funkybob: OK, I assumed since the "successful" path returned a 1, a 0 would mean fail.

funkybob: Other modules exit(1) on errors in init, so I'll follow.

plugins/python/python_plugin.c
```diff
@@ -242,6 +242,10 @@ int uwsgi_python_init() {
 	}
 
 	if (up.home != NULL) {
+		if (!uwsgi_is_dir(up.home)) {
+			uwsgi_log("Python Home is not a directory: %s\n", up.home);
+			exit(1);
```
xrmx: That was more of a question than a suggestion but i see we're racing each other :) I'm thinking of people that does not test their config before putting into production that may have the site working outside the home.

funkybob: Are you suggesting it's possible this works in some situations even with an invalid value?

funkybob: Am happy to make this a lower impact patch that simply logs, and doesn't exit.

unbit: may i suggest to have it only print an error in the logs for the 2.0 branch and to brutally exit (like the one you proposed) in master ? (where we suppose to move our choices to 2017 best-practices  instead of 2009 ;)

funkybob: Sounds perfect to me :)
Shall I leave it all to you from here, then? The difference being a single line.

unbit: yep, i'll do it

PR #1644, opened 2017-09-20
Leverage Linux's getcwd

funkybob:
According to the man page:

       As an extension to the POSIX.1-2001 standard, glibc's getcwd() allocates the buffer dynamically using malloc(3) if buf is NULL.  In this case, the allocated buf‐
       fer has the length size unless size is zero, when buf is allocated as big as necessary.  The caller should free(3) the returned buffer.

So to save us some work, just call getcwd(NULL, 0);

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -175,7 +175,9 @@ void daemonize(char *logfile) {
 
 // get current working directory
 char *uwsgi_get_cwd() {
-
+#if defined(__GLIBC__)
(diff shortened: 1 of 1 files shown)


xrmx:
That would break musl i think, if buf is null it goes on the stack there http://git.musl-libc.org/cgit/musl/plain/src/unistd/getcwd.c

PR #1650, opened 2017-09-24
[Fixes #1587] Support If-Range for static (partial)

### LINKED ISSUE #1587: Static files serving does not respect the If-Range header
uWSGI unconditionally returns a Partial Response for Range requests, even if the If-Range does not match the condition (with `--honour-range`).

References:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Range
- https://tools.ietf.org/html/rfc7233#section-3.2

funkybob:
Supports timestamps in If-Range on static assets.
Does not support ETags - will treat as implicitly correct.
None of the static system currently supports ETags.

### DIFF
--- a/core/protocol.c
+++ b/core/protocol.c
@@ -277,6 +277,11 @@ static void uwsgi_parse_http_range(char *buf, uint16_t len, enum uwsgi_range *pa
 
 static int uwsgi_proto_check_10(struct wsgi_request *wsgi_req, char *key, char *buf, uint16_t len) {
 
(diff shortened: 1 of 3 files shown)


unbit:
@xrmx safe for backport

xrmx:
@funkybob Had to backport a few dependencies, please give it a try to this branch please https://github.com/xrmx/uwsgi/tree/backportrange

funkybob:
This was a fix for someone else's issue ... @GaretJax ?

PR #1653, opened 2017-09-29
Add log-before option

funkybob:
Add an option to log the pid, remote address, request method and URI early, to aid in debugging.

### DIFF
--- a/core/protocol.c
+++ b/core/protocol.c
@@ -702,6 +702,14 @@ int uwsgi_parse_vars(struct wsgi_request *wsgi_req) {
 	}
 
 next:
+	if (uwsgi.log_before) {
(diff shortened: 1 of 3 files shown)


funkybob:
This feature was requested by MasterGberry on IRC, in aiding discovering which requests were causing their processes to be OOM killed.

I'm not 100% sure this is the right place for it, but it seems it will be harmless in other cases.

mastergberry:
Confirmed that the above feature would be extremely helpful in debugging an issue with requests not completing due to OOM.

unbit:
Is not it a bit too much ad-hoc ? Maybe it is better to add an hook to that part so the code can go in a plugin.

funkybob:
Yeah, seemed a bit specialist, and I didn't expect it to be accepted :)

Time to learn plugins... yay! :)

PR #1662, opened 2017-10-12
correctly handle IPv6 sources in proto/http ; branch on top of 2.0.x

Roguelazer:
This is #1203 rebased and squashed down on top of the 2.0 branch in case we get to do a 2.0.16, as requested by Curtis on the mailing list.

### DIFF
--- a/proto/http.c
+++ b/proto/http.c
@@ -175,8 +175,8 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
 	char *ptr = wsgi_req->proto_parser_buf;
 	char *base = ptr;
 	char *query_string = NULL;
-	char ip[INET_ADDRSTRLEN+1];
-	struct sockaddr_in *http_sin = (struct sockaddr_in *) &wsgi_req->c_addr;
+	char ip[INET6_ADDRSTRLEN+1];
+	struct sockaddr *http_sa = (struct sockaddr *) &wsgi_req->client_addr;
 	char *proxy_src = NULL;
 	char *proxy_dst = NULL;
 	char *proxy_src_port = NULL;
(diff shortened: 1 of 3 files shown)


Roguelazer:
I had to make changes because of eg the lack of http/1.1 support in 2.0.x. It's possible that something got confused in my changing of that stuff, I guess – especially with the unusual mixture of tabs and spaces in these files.

-- 
James Brown,
currently mobile

funkybob:
There is a separate PR to backport http/1.1 support to 2.0... so perhaps this will become simpler once that lands?

Roguelazer:
I just manually re-copied the main block of this from the ipv6 branch to fix-up indentation; not sure exactly how it got to be incorrect.

I also noticed that the `ip` field was being zeroed out twice in a row in the `AF_INET6` branch, so I removed the duplicate.

proto/http.c
```diff
@@ -293,14 +293,56 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
...
+				}
+			} else {
+				if (inet_ntop(AF_INET6, (void *) &http_sin->sin6_addr, ip, INET6_ADDRSTRLEN)) {
+					wsgi_req->uh->pktsize += proto_base_add_uwsgi_var(wsgi_req, "REMOTE_ADDR", 11, ip, strlen(ip));
+				} else {
+					uwsgi_error("inet_ntop()");
+					return -1;
+				}
+			}
+			}
```
funkybob: It would be less confusing to read if this brace were below the break, and one indent fewer.

Roguelazer: Fixed. I'm also going to rebase on the uwsgi-2.0 branch.

unbit:
Do we have some feedback about its stability in 2.0 ? Some of you using it in production ?

uwsgi.h
```diff
@@ -1362,7 +1362,8 @@ struct wsgi_request {
 	char *appid;
 	uint16_t appid_len;
 
-	//this is big enough to contain sockaddr_in
+	// This structure should not be used any more
+	// in favor of the union client_addr at the end
 	struct sockaddr_un c_addr;
```
funkybob: Would it make sense to rename this so any code trying to build against it would fail loudly?

Roguelazer: I don't know if anyone cares about symbol compatibility or not; if not, then that sounds like a good idea.

Roguelazer:
@unbit I have been using an older version based on 2.0.12 in production on several thousand instances without any issues.

PR #1681, opened 2017-11-10
Do not override environ if not necessary

myreg:
On Linux, initial process environment can be accessed from outside via /proc/<pid>/environ. We'd like to use this for uwsgi as well, but uwsgi clears the memory for this in the beginning so the file looks completely empty. The memory for argv and environ is used for a custom process name.

With the changes in this PR,
- argv_len is computed correctly (before, the length of all argv except the first was ignored due to off by one error)
- argv_len and environ_len is stored separately and if the new procname can fit into argv, environ is kept untouched

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -3124,33 +3124,36 @@ void uwsgi_set_processname(char *name) {
 
 #if defined(__linux__) || defined(__sun__)
 	size_t amount = 0;
+	size_t max_procname = uwsgi.argv_len + uwsgi.environ_len;
 
 	// prepare for strncat
 	*uwsgi.orig_argv[0] = 0;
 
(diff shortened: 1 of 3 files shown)


myreg:
@unbit any thoughts on this / could you help me find someone to review this?

myreg:
@unbit please please please? :)

unbit:
@myreg ooops, sorry for the huge late, can you move the new fields of uwsgi_server at the end of the struct to retain ABI compatibility ? Thanks a lot

myreg:
@unbit thanks a lot for looking, updated!

unbit:
@xrmx still need to check if the patch works on solaris, so wait before backporting. Thanks @myreg

PR #1684, opened 2017-11-18
core: format harakiri request start

### LINKED ISSUE #1682: harakiri message printing formatted time
With uwsgi 2.0.15 harakiri prints the time of the request in epoch while a formatted date maybe would be helpful:
```
Thu Nov 16 10:58:35 2017 - HARAKIRI [core 0] 93.50.186.51 - GET /foo since 1510826194
```

xrmx:
Instead of displaying the epoch print a formatted localtime.

Fix #1682


unbit:
I know its annoying, but changing the format of logs can break those scripts that parses logs. Can you wrap it into an option ?

xrmx:
Sure, will do

xrmx:
Done, was wondering if it would make sense to split the log so we can reuse most of it.

PR #1686, opened 2017-11-21
Enable execinfo on DragonFly BSD

liweitianux:
DragonFly BSD also has execinfo, so enable it.

I've tested this patch on DragonFly 5.1-master.

Cheers,
Aly


unbit:
Thanks

PR #1687, opened 2017-11-22
Fix for request URLs containing spaces

mialinx:
This patch fixes bug with spaces in request url.

How to reproduce bug:

* Use following WSGI app:
```
    return ["REQUEST_URI: {REQUEST_URI}\nPATH_INFO: {PATH_INFO}\n".format(**env)]
```
* Make following request (with literal spaces in URL)
`curl -v 'http://127.0.0.1:5555/h ello?a=o my'`

Expected result: server conforms HTTP protocol
Actual result: server returns malformed HTTP response:
```
[nuf@dsmal-work ~]$ curl -v 'http://127.0.0.1:5555/h ello?a=o my '
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 5555 (#0)
> GET /h ello?a=o my HTTP/1.1
> Host: 127.0.0.1:5555
> User-Agent: curl/7.47.0
> Accept: */*
> 
ello?a=o my god HTTP/1.1 200 OK
Content-Type: text/html

REQUEST_URI: /h
PATH_INFO: /h
* Connection #0 to host 127.0.0.1 left intact
```


mialinx:
After patch it works like
```
[nuf@dsmal-work ~]$ curl -v 'http://127.0.0.1:5555/h ello?a=o my'
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 5555 (#0)
> GET /h ello?a=o my HTTP/1.1
> Host: 127.0.0.1:5555
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/html
* no chunk, no close, no size. Assume close to signal end
< 
REQUEST_URI: /h ello?a=o my
PATH_INFO: /h ello
* Closing connection 0
```

proto/http.c
```diff
@@ -235,6 +236,7 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
 
 	// REQUEST_URI / PATH_INFO / QUERY_STRING
 	base = ptr;
+	QS:
```
xrmx: labels on column 0 please, also something more descriptive than QS may be helpful

xrmx:
It would be cool to have all the description of the problem in the commit instead of just github :)

mialinx:
fixed

mialinx:
@xrmx,  could you look at this patch again, please ?

Dzirtik:
@xrmx please accept this patch, it's very important security fix when uwsgi stands after nginx

unbit:
Sorry, maybe i am missing something, but should not we discard those kind of requests ? The client should escape spaces to be compliant with the standard

mialinx:
Yes, you are right, but from my point of views server should never respond with malformed http response. It may handle such requests or return 400 "bad request" error.

In my situation, client is hacker's script, that forces server to return malformed response to avoid SSI processing on proxy server. It's a security issue and I can't change client's behaviour.

unbit:
@mialinx what about validating the protocol ? if it contains spaces we discard the request

mialinx:
Ok, I'll change the code to return HTTP 400 in such cases.

PR #1694, opened 2017-12-19
not clear semaphore on forked master

brainfvck:
If using semaphore as lock engine and the forked master is in a crash loop, it'll eventually bring down the old worker processes too.
The problems is the forked master shouldn't try to clear the state(semaphore in this case) inherited from the old master.

To repro, set worker number = 1, and 
1. start server and send a request - let worker touch the semaphore
2. kill -9 the worker process - kill the last process touched it
3. change uwsgi.ini to a bad one(in my case, I setup a cache items > blocks) then do master fork - fail the master fork
4. now old master fails too, the exception is uwsgi_lock_ipcsem()/semop(): Invalid argument [core/lock.c line 547] because semaphore cleared by new master..

### DIFF
--- a/core/lock.c
+++ b/core/lock.c
@@ -612,7 +612,7 @@ pid_t uwsgi_lock_ipcsem_check(struct uwsgi_lock_item *uli) {
 
 void uwsgi_ipcsem_clear(void) {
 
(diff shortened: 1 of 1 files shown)


brainfvck:
not working as expected..

PR #1704, opened 2018-01-03
Add trove classifiers for all supported Python versions

jdufresne:
Help users know, at a glance, if uWSGI is compatible with their project.


xrmx:
Can't hurt i suppose, thanks! It looks we need to fix travis :(

PR #1705, opened 2018-01-03
Remove redundant default open mode 'r' from Python code

jdufresne:
In Python, `open()` opens files in read mode (`'r'`) by default. Specifying it is unnecessary and simply adds noise. It is Pythonic to instead rely on this default value.

See https://docs.python.org/3/library/functions.html#open

> mode is an optional string that specifies the mode in which the file is opened. It defaults to 'r' which means open for reading in text mode.


xrmx:
Checked that mode is default 'r' even for ancient python (2.5). Thanks!

PR #1707, opened 2018-01-03
Use context managers throughout Python code where available

jdufresne:
Context managers can wrap code blocks to help ensure resources are
always cleaned up properly, even in the case of an error. Instead of
manually managing resources, let the context manager do it is instead.
Available on all supported Python versions. Use a more modern Python
syntax.

https://docs.python.org/3/reference/compound_stmts.html#with


xrmx:
It looks like python 2.5 requires an import from __future__, do you still need 2.5 @unbit?
https://pymotw.com/2/contextlib/

jdufresne:
Python 2.5 is not included in the Travis tests nor the trove classifiers. Is it an officially supported version?

xrmx:
@jdufresne You added the trove classifiers just now i won't use them as reference :) python2.5 is not in the travis tests to save some time and quite frankly nobody should use it in 2017. But AFAIR Roberto has some customers using it and since it's the main developer of uwsgi the least we can do is not to break it. We can add python2.5 back after i unbreak travis.

jdufresne:
Rebased. Travis is now green.

PR #1708, opened 2018-01-03
Add classifier to document uWSGI is stable and ready for production

jdufresne:
Helps users know, at a glance, that uWSGI is actively maintained and
suitable for production.


xrmx:
If it's in the trove classifiers it should be true! :) Thanks!

PR #1709, opened 2018-01-03
Prefer uwsgi_opt_true to uwsgi_opt_set_int for boolean value

cdosborn:
Problem:
When I set `python-autoreload: true`, the app does not reload on changes
When I set `python-autoreload: 1`, the app reloads on changes

I didn't test this code, was hoping that someone with more experience in the project could give it a go. If the fix is correct, it looks like there are other places where the configuration could uniformly handle "true" in place of "1".


xrmx:
Well it would be nice to test the code and for instance check you are not breaking backward compatibility so that python-autoreload: 1 continue to work fine.

cdosborn:
Was trying to submit something a little more helpful than just an issue. Looks like set_opt_true function handles '0'/'1' just like 'true'/'false'.

PR #1734, opened 2018-02-06
move plugin post_jail hooks to post_jail start

bjne:
The tuntap router is the only included plugin that currently
uses the post_jail hooks to attach a the tuntap interface in
the new namespace. Move the execution of the plugin post_jail
hooks to before hook_post_jail/exec_post_jail to be able to
configure the interface just created before pivot_root

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -514,6 +514,13 @@ void uwsgi_as_root() {
 #endif
 
 	if (in_jail || uwsgi.jailed) {
+		int i;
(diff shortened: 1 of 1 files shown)


unbit:
Hi, thanks, i would really prefer if you add a new hook phase, there are third plugins that could assume the previous behaviour ;)

bjne:
I can do that, but it will then require change to the tuntap plugin. early-post-jail? Only for plugins? See also comments on the issue.

bjne:
new phase approach in #1735

PR #1735, opened 2018-02-06
new plugin hook: early_post_jail

bjne:
post_jail plugin hooks runs after all other hooks in post_jail,
so it was not possible to configure for example tuntap interface
in a post-jail exec/hook, since it was not yet available.

This patch adds a early_post_jail phase that plugins can attach to,
and moves initialization of tuntap interface to this phase.

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -514,6 +514,13 @@ void uwsgi_as_root() {
 #endif
 
 	if (in_jail || uwsgi.jailed) {
+		int i;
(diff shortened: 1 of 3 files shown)


unbit:
Thanks, seems good, just move the new uwsgi_plugin struct field at the end of the structure to maintain backward compatibility

bjne:
Like this?

unbit:
Perfect :)

@xrmx safe for backport

PR #1736, opened 2018-02-07
Add option to set post-buffering on HTTP fast router

funkybob:
I'm not aware of any reason this option should be available FastRouters and not HTTP fastrouters.


unbit:
Lol, it is incredible it was not there, Thanks :)

PR #1737, opened 2018-02-07
poc: unittest using check

pjstevns:
Not a real pull-request, but more a proof of concept since I have a lot of good experiences with the check unit-test framework.

a simple 'works-on-my-machine' proof of concept

- use the 'check' unit-test framework for C
- adds build config for building static libuwsgi.a
- run make tests to build the static lib, the tests, and run the tests
- uses a simple make file in the check directory for now which needs better
  integration with the custom build system

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ core/dot_h.c
 /build/
 /dist/
 /uWSGI.egg-info/
+check/check_*
+!check/*.c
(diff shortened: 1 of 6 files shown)


pjstevns:
Happy to help. We use and have used uwsgi for *all* our projects, past, present and future. It sure needs more than just the token love and good wishes.

funkybob:
On a relevant topic - https://www.youtube.com/watch?v=z-uWt5wVVkU

unbit:
OK, sorry for the late, but it took me a bit to analyze the various solutions available and now i think 'check' is the most 'friendly' one (love the fork() usage for test isolation).

Merging it as-is. Thanks a lot

PR #1738, opened 2018-02-09
Fix inverted logic in can-gzip checks

funkybob:
Turns out my previous patch didn't work, and actually broke ``static-gzip`` entirely :/


funkybob:
Obviously this should be back-ported before the 2.0.16 release

Issue #1717, opened 2018-01-13
Improve api for uwsgi websockets


danidee10:
Straight from the docs for `uwsgi.websocket_recv`:

> This is the real core of the whole implementation.
> 
> This function actually lies about its real purpose. It does return a websocket message, but it really also holds the connection opened (using the ping/pong subsystem) and monitors the stream’s status.

The function/logic for holding the connection open should be moved to another method. I propose something like:

`usgi.websocket_ping_pong`

In cases where you have a write only websocket (You're not reading any messages from it, just pushing to the client) it doesn't really make sense to call `uwsgi.websocket_recv` just to keep the connection. The logic should be abstracted into another method.

I don't know much about the `uWSGI` core but if this issue is considered i would be glad to work with the core devs in implementing this.

danidee10:
Okay @unbit Let me get my hands dirty and see what i can do.

I'll probably need to join the #IRC Channel

Issue #1533, opened 2017-05-17
no PONG received in 3 seconds


xiejieyi-bear:
I used uwsgi.websocket_recv_nb() to received message behind nginx.
but 	Irregularly Recur error: 'no PONG received in 3 seconds'.
Probability is 1/1000 or lower.
I can't find the reason.

can uwsgi exposes related api to close the ping/pong mechanism？

unbit:
I think it is is something we can expose. I leave the issue opened for taking track of it, thanks

maxim-s-barabash:
Fixed: https://github.com/unbit/uwsgi/commit/63c6b8688a9de8440f9883d1b49982a1514910cc

PR #1744, opened 2018-02-18
Create new websocket function websocket_check_pingpong

danidee10:
This adds a new method called `websocket_check_pingpong` #1717 

Right now, It's merely a wrapper around `uwsgi_websocket_recv_do` (like `websocket_recv` and `websocket_recv_nb`) and the non-blocking behaviour is hardcoded.

I tried to call the C function `uwsgi_websockets_check_pingpong` but I got errors and since the internals of `uWSGI` isn't documented, I didn't really know what to do. I believe some things must be done on the `wsgi_request` (e.g `uwsgi_websocket_parse_header(wsgi_req);`) before trying to send a ping or receive a pong.

The only solution was to copy most of the code from `uwsgi_websocket_recv_do` but that doesn't seem right.

I'm submitting this PR (Not to be merged in its current state) because I don't really know how to proceed and how to handle non-blocking and blocking mode that's an argument to `uwsgi_websocket_recv_do` (It's only used in one place in the function body so I feel it can easily be removed) .


danidee10:
This is loosely related to #1533 because with the new function I still experience the same `No PONG received in 3 seconds` and currently the only workaround is to kill the `uWSGI` process which forces it to restart (provided `uWSGI` is started with it's `--master` process). and on the client side try and re-establish a connection.

If there's a separate function for handling pings and pongs it'll be easier to solve the problem since we won't be debugging code that has to deal with receiving of packets.

maxim-s-barabash:
I get a problem with "No PONG received in 3 seconds"
Is there a solution to this problem?

maxim-s-barabash:
the proposed solution https://github.com/unbit/uwsgi/pull/1261

Issue #1677, opened 2017-11-08
mod_proxy_uwsgi fails with EFAULT on client read timeout


joshuahlang:
With a relatively simple Apache 2.4 config using mod_proxy_uwsgi (latest version from github) and mod_reqtimeout:
```
RequestReadTimeout body=10,MinRate=1000
ProxyPass / unix:/var/run/uwsgi/uwsgi.sock|uwsgi://uwsgi-service/ timeout=60
```

The following error is generated by Apache when there is a timeout reading request content from a client:
```
[(14)Bad address] sending data to httpd-UDS:0 failed
```

I'm generating the client timeout using netcat to send HTTP POST headers without the complete POST body.

This is caused by casting the return value of `ap_get_client_block`, which returns a _signed_ `long` value, to the _unsigned_ `apr_size_t` type [here](https://github.com/unbit/uwsgi/blob/master/apache2/mod_proxy_uwsgi.c#L202)

Once that is fixed, mod_proxy_uwsgi still returns `HTTP_INTERNAL_SERVER_ERROR` due to the timeout error. Ideally this would return an HTTP 408 status instead. Checking if the error is a timeout-like error and returning `HTTP_REQUEST_TIME_OUT` seems reasonable, but I'm not very familiar with Apache module development. See below for the exact code change:

```
diff --git a/apache2/mod_proxy_uwsgi.c b/apache2/mod_proxy_uwsgi.c
index d66243d..a9530df 100644
--- a/apache2/mod_proxy_uwsgi.c
+++ b/apache2/mod_proxy_uwsgi.c
@@ -197,20 +197,21 @@ static int uwsgi_send_body(request_rec *r, proxy_conn_rec *conn)
     if (ap_should_client_block(r)) {
         char *buf = apr_palloc(r->pool, AP_IOBUFSIZE);
         int status;
-        apr_size_t readlen;
+        long readlen;
 
         readlen = ap_get_client_block(r, buf, AP_IOBUFSIZE);
         while (readlen > 0) {
-            status = uwsgi_send(conn, buf, readlen, r);
+            status = uwsgi_send(conn, buf, (apr_size_t)readlen, r);
             if (status != OK) {
                 return HTTP_SERVICE_UNAVAILABLE;
             }
             readlen = ap_get_client_block(r, buf, AP_IOBUFSIZE);
         }
         if (readlen == -1) {
-            ap_log_rerror(APLOG_MARK, APLOG_ERR, 0, r,
[...]

PR #1754, opened 2018-03-06
Return a better HTTP status on timeout error when reading from client

joshuahlang:
https://github.com/unbit/uwsgi/issues/1677

Return a more accurate HTTP_REQUEST_TIME_OUT on any read timeout error
from the client instead of the generic HTTP_INTERNAL_SERVER_ERROR.


xrmx:
The apache module development has moved to the the apache httpd project. Not sure it's a good idea to update our deprecated copy.

PR #1768, opened 2018-03-29
Build in current Rawhide (rb_funcall bogus arguments + do not require old hacks)

Jakuje:
These changes makes it building with Fedora rawhide, where GCC is probably more strict about the bogus arguments.

The other patch is removing old workarounds and using again what is done by ngingx on new systems:

https://src.fedoraproject.org/cgit/rpms/nginx.git/tree/0001-unix-ngx_user-Apply-fix-for-really-old-bug-in-glibc-.patch

Now, I still have problems to build glusterfs plugin in rawhide, but it works in Fedora 28.


unbit:
Thanks a lot

Jakuje:
@ianw Release for other distros would certainly make sense. In the meantime, this change is already in Fedora 28:
https://src.fedoraproject.org/rpms/uwsgi/c/42f326b

PR #1772, opened 2018-04-03
plugins/glusterfs: Add path to the header files

Jakuje:
Related gluster commit removing this path from packageconfig variables:
http://git.gluster.org/cgit/glusterfs.git/commit/?id=684d6227
https://review.gluster.org/#/c/18576/


Jakuje:
This commit makes the 2.0.16 version build successfully in latest Fedora Rawhide.

unbit:
Thanks

Issue #1763, opened 2018-03-21
current_salt is a private member in glibc and libxcrypt does not have it


afazekas:
plugins/router_basicauth/router_basicauth.c copied a glibc 2.2.5 (year 2002) related workaround from nginx which references to  currently `undocumented` struct member.

https://www.mankier.com/3/crypt_rn

Portable code MUST NOT expect any other member than the `initialized` in the  crypt_data struct. Portable code should not do anything else than initilzing (zero) the crypt_data.

uwsgi builds are failing on Fedora28 with libxcrypt 4.0.0.5.fc28 and glibc-2.27-6.fc28 .  /usr/include/crypt.h is from the libxcrypt not from glibc.

plugins/router_basicauth/router_basicauth.c: In function ‘htpasswd_check’:
plugins/router_basicauth/router_basicauth.c:73:5: error: ‘struct crypt_data’ has no member named ‘current_salt’
   cd.current_salt[0] = ~cpwd[0];
     ^
make: *** [Makefile:4: all] Error 1

I wonder, is it possible simply deleting the offending line?

xrmx:
Thanks for reporting, any hint on which definition we can use to know that crypt.h is not the glibc one?

afazekas:
I found a similar ticket in nginx. The patch checks the glibc version:
https://trac.nginx.org/nginx/ticket/1469

pabelanger:
I can confirm we are seeing the same failure on fedora-28:
  http://logs.openstack.org/32/565632/4/check/devstack-platform-fedora-latest/a7e4979/job-output.txt.gz

ianw:
Note the patch applied to fedora for this is https://src.fedoraproject.org/cgit/rpms/nginx.git/tree/0001-unix-ngx_user-Apply-fix-for-really-old-bug-in-glibc-.patch and does as @afazekas suggets, basically ignoring it for recent glibc.

This is referenced in #1768 which fixes another problem for current fedora too; might be worth merging these to avoid confusion.

afazekas:
https://github.com/unbit/uwsgi/commit/00dce4265784903a3bf440834ab269fceaf5a14b  solved the crypt issue, but I am not sure the `#include <crypt.h>` really needs to be excluded.

afazekas:
[...]

PR #1780, opened 2018-05-03
Restore re-entrant crypt_r usage

afazekas:
The old glibc workaround removal
introduced by 00dce4265784903a3bf440834ab269fceaf5a14b also
removed the crypt_r usage.

This change limits the scope of workaround only
the known wrong glibc version and keep using the crypt_r on the others.

Very likely more libc and platform could use crypt_r instead of locking
but properly handling all platforms is out of the scope of this change.

Issue #1763: current_salt is a private member in libc

### DIFF
--- a/plugins/router_basicauth/router_basicauth.c
+++ b/plugins/router_basicauth/router_basicauth.c
@@ -2,10 +2,8 @@
 
 #ifdef UWSGI_ROUTING
 
-#if defined(__linux__) && (defined(__GLIBC__) && __GLIBC__ == 2) && \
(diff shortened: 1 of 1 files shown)


yodermk:
This fixes it for me on Fedora 28.  Having had this merged and released prior to now would have saved me some grief ...

mwconceicao:
Worked for me on Fedora 28!

JohnRoach:
Hi is there a way we can get this merged in?

rdeioris:
i will work on it tomorrow. sorry for the late

unbit:
Ok, finally merged, @xrmx safe for backport

xrmx:
Backported in uwsgi-2.0

PR #1782, opened 2018-05-04
Tiny validation on request_uri

fsitedev:
As a continue of ticket https://github.com/unbit/uwsgi/pull/1687.

This PR does a simple "request_uri" check and responds with "400 Bad request" to the client if the first line of request contains more than one space character.

p.s.
Litterally I couldn't find a common approach of responding on bad requests with 400 code, so applied a small patch directly in the "http_parse" function. Hope it's ok.

### DIFF
--- a/plugins/http/http.c
+++ b/plugins/http/http.c
@@ -1183,6 +1183,26 @@ ssize_t http_parse(struct corerouter_peer *main_peer) {
 			int skip = http_headers_parse_first_round(new_peer);
 			if (skip < 0) return -1;
 
+			// check for spaces in request_uri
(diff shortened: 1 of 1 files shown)


fsitedev:
Please react somehow on this PR. It's really important issue for those who use nginx as a reverse proxy and uwsgi behind it.

grauwolf32:
Any updates ?

fsitedev:
@unbit is there any chance of getting feedback on this PR?

unbit:
Hi, the http router is pretty brutal on broken requests and simply closes the connection. Maybe we should maintain this rule (albeit ugly) ?

Issue #1808, opened 2018-06-20
Handling unix signal with python code


almaslov:
I want to trigger some action in master with unix signal. 
Python signal handlers are a little tricky: if signal was caught in C subroutine, its python handler won't be executed until you return to Python code, or call PyErr_CheckSignals function. In case of master, it is never executed, since there are no python calls in master_loop.
As i can see, the best way to fix this behavior is to define master_cycle callback for python plugin, where PyErr_CheckSignals can be called. It takes only one extra if-statement per cycle, if no signals caught.

So, the question: is there some reason for master not to handle additional unix signals other than these: https://uwsgi-docs.readthedocs.io/en/latest/Management.html? 
If there are no reasons, i would be glad to send pull request.

PR #1809, opened 2018-06-21
plugins/python: Enable handling unix signals with python in master

almaslov:
Pull request for https://github.com/unbit/uwsgi/issues/1808
master_cycle callback added for python plugin. It calls PyErr_CheckSignals to check if any signal handlers scheduled. Exceptions in signal handlers are fatal.


unbit:
Hi, thanks for your patch. I think you need to reacquire the GIL before calling PyErr_CheckSignals(). You can make a quick test by adding --enable-threads

almaslov:
Done. Should i squash fix to initial commit?

unbit:
Thanks, merged as is :)

PR #1822, opened 2018-07-05
Allow to use autoport (socket :0) with custom socket backlog

pavelpatrin:
When uWSGI starts with configuration like this

    uwsgi:
        socket: 0.0.0.0:0
        listen: 1000

we see in startup log messages like this

    uwsgi socket 0 bound to TCP address 0.0.0.0:46490 (port auto-assigned) fd 3
    ...
    your server socket listen backlog is limited to 1024 connections

but when ours application stubs we see different backlog size

    *** uWSGI listen queue of socket "1.2.3.4:46490" (fd: 3) full !!! (101/100) ***

and real socket backlog size is 100.

This happens because uwsgi binds socket with auto port (uwsgi_new_socket() at socket.c:1044) BEFORE reading parameter `listen` and uses its default value 100.

### DIFF
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -176,7 +176,7 @@ static struct uwsgi_option uwsgi_base_options[] = {
 	{"connect-and-read", required_argument, 0, "connect to a socket and wait for data from it", uwsgi_opt_connect_and_read, NULL, UWSGI_OPT_IMMEDIATE},
 	{"extract", required_argument, 0, "fetch/dump any supported address to stdout", uwsgi_opt_extract, NULL, UWSGI_OPT_IMMEDIATE},
 
-	{"listen", required_argument, 'l', "set the socket listen queue size", uwsgi_opt_set_int, &uwsgi.listen_queue, 0},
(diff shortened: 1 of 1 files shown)


xrmx:
Why do you people write beautiful explanation but for some reason don't commit them? :)

pavelpatrin:
Guys, what i could improve in PR to make it better?

unbit:

xrmx:
Backported, thanks.

jnm:
Does this change to `UWSGI_OPT_IMMEDIATE` mean that the `$(ENV_VAR)` and `@(file_name)` syntax don't work for `listen` anymore? It was possible to use them in 2.0.17.1, but they don't work with 2.0.18.
```
jnm@world$ uwsgi --version
2.0.17.1
jnm@world$ OKAY=128 uwsgi --socket 127.0.0.1:1234 --listen '$(OKAY)' 2>&1 | grep backlog
your server socket listen backlog is limited to 128 connections
```
```
jnm@world$ uwsgi --version
2.0.18
jnm@world$ OKAY=128 uwsgi --socket 127.0.0.1:1234 --listen '$(OKAY)' 2>&1 | grep backlog
your server socket listen backlog is limited to 0 connections
```
OS is Ubuntu 18.04.3 LTS, kernel 4.15.0-36-generic. uWSGI was installed by pip 19.3.1 within a Python 3.6.8 virtualenv.

PR #1826, opened 2018-07-09
uwsgi compatibility with windows

DG007:
change the import os and os.uname function calls with import platform and platform.uname. This makes uwsgi compatible accross all platforms and can be used in windows without restrictions.

### DIFF
--- a/uwsgiconfig.py
+++ b/uwsgiconfig.py
@@ -5,10 +5,11 @@ uwsgi_version = '2.1-dev'
 import os
 import re
 import time
-uwsgi_os = os.environ.get('UWSGI_FORCE_OS', os.uname()[0])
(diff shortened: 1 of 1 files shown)


thomasf:
I think that this might fix compatibility with pipenv ( https://github.com/mirumee/saleor/issues/3355 )
Could we maybe have this in the next release @unbit ?

BrennanR:
I've been running into the problem with pipenv gathering dependencies under windows and its frustrating. Any reason this PR isn't being merged? It's been open just about a year now.

thomasf:
this by itself does not seem to fix that problem though IIRC..  I tried to publish a new uwsgi package with this patch on pypi to test  but it didnt work.. I  probably did something wrong though since the package did not work at all for linux either, I have not had time to look in to it again.

Issue #1844, opened 2018-08-02
php mountpoint support


vault:
It'd be nice if the php plugin let you map url-prefixes to different directories. Most other plugins seem to have some way of doing it (whether with the mount option or a plugin specific option).

vault:
Closing since the PR implementing this was merged.

PR #1846, opened 2018-08-03
Add php-mount option

vault:
Taking a stab at #1844.

This adds a php-mount option to the php plugin. This option maps a URL
prefix to a file-system path. The configured location will be used in
lieu of docroot or cwd if it's set.


unbit:
Thanks

PR #1848, opened 2018-08-09
Make pypy WSGIfilewrapper a true iterable.

purple4reina:
According to [PEP 333](https://www.python.org/dev/peps/pep-0333/#optional-platform-specific-file-handling)

> The callable must return an iterable object

However, the `WSGIfilewrapper` as defined in `pypy_setup.py` is not an iterable because it does not have a `next` or `__next__` method.

Here is a sample application that when run with pypy reproduces the issue:

```python
    status = '200 OK'
    response_headers = [('Content-type', 'text/plain')]

    file_wrapper = environ.get('wsgi.file_wrapper')
    return file_wrapper(open('hello_world.txt'))

class ConsumingMiddleware(object):
    def __init__(self, application):
        self.application = application

    def __call__(self, environ, start_response):
        data = []
        for item in self.application(environ, start_response):
            data.append(item + '?')
        return data

application = ConsumingMiddleware(application)
```

When you run this, you'll see an error like the following:

```
From cffi callback <function uwsgi_pypy_wsgi_handler at 0x00007f144c981178>:        
  File "c callback", line 472, in uwsgi_pypy_wsgi_handler                           
  File "./app.py", line 16, in __call__                                             
    for item in self.application(environ, start_response):                          
TypeError: iter() returned non-iterator
```

The solution is to add `next` and `__next__` methods to the `WSGIfilewrapper` object.

### DIFF
--- a/plugins/pypy/pypy_setup.py
+++ b/plugins/pypy/pypy_setup.py
@@ -382,6 +382,17 @@ class WSGIfilewrapper(object):
     def __iter__(self):
         return self
 
+    def __next__(self):
+        if self.chunksize:
+            data = self.f.read(self.chunksize)
+        else:
+            data = self.f.read()
+        if data:
+            return data
+        raise StopIteration()
+
+    next = __next__
+
     def sendfile(self):
         if hasattr(self.f, 'fileno'):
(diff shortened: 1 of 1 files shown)


unbit:
Thanks

Issue #816, opened 2015-01-16
Make pyuwsgi plugin python 3 compatible


msurdi:

unbit:
Do you use it or you are a packager ? I am asking because we were thinking about removing it, but if you have a use it for it we can start mantaining it again. Thanks

msurdi:
I was just looking for a way to run uwsgi from a freezed (cx_freeze) application. TBH, this is just some experimentation so I'm not using it right now, just investigating if this could work in this setup so I can replace gunicorn.

xtfxme:
I am using this. Combined with `setup.cpyext.py` in 2.1 you get a real python module, `uswgi.so`, and you can call `uwsgi.setup` return to python and do more work (in the master while interacting with uWSGI api) then call `uwsgi.run`.

IMO this should be the default for pip installations at the least, and fixes at least 2-3 pip related bugs in the tracker.

funkybob:
Is this issue still outstanding, or has it been fixed?

natefoo:
@unbit any chance this will see some work? If not, I have looked at it a bit, and can commit some more effort to it in a few months.

We do use pyuwsgi quite heavily, as it is (to my knowledge) the only way to provide a precompiled uwsgi that does not either: 1. require a dynamic libpython to be installed on the system, or 2. statically link libpython. This is how we build uWSGI wheels for [wheels.galaxyproject.org](https://wheels.galaxyproject.org) and aside from a few non-showstopping bugs, it works well for Python 2.

ipmb:
For anybody keeping an eye on this issue (and wanting to build wheels), [Lincoln Loop](https://lincolnloop.com/) is sponsoring development for it. @unbit just [committed a fix for building pyuwsgi on Python 3](https://github.com/unbit/uwsgi/compare/9055cf1a0db63adaba21189099e96a953992fbb2...eebe9f251646e83c92669777a725cbd058c7f72f). From the `uwsgi-2.0` branch, the shared library can be built with:

```bash
UWSGI_EMBED_PLUGINS=python make pyuwsgi PYTHON=python3
```

but I'm hoping we can get to the point where that just happens as part of the pip install. 🤞

natefoo:
@ipmb Great, this is [better than my efforts so far](https://travis-ci.org/natefoo/starforge-recipes/builds/426822660). Two potential outstanding issues with wheels in PyPI (that maybe should get their own issue(s) but I'm mentioning them here since you're currently working on the packaging):

[...]

PR #1863, opened 2018-09-10
Build the uWSGI shared module as pyuwsgi when building under Python 3

natefoo:
Otherwise, the so filename and importable don't match the module name:

```pytb
  File "/home/nate/.virtualenvs/pyuwsgi3/bin/uwsgi", line 7, in <module>
    from uwsgi import run
ImportError: dynamic module does not define module export function (PyInit_uwsgi)
```

With the recent completion of #816, this makes it possible to build a Python 3 pyuwsgi wheel using the standard tooling.

### DIFF
--- a/setup.cpyext.py
+++ b/setup.cpyext.py
@@ -96,6 +96,21 @@ class uWSGIBuildExt(build_ext):
                             ext.extra_compile_args.append(y)
 
 
+class uWSGIPy2BuildExt(uWSGIBuildExt):
+    pass
+
+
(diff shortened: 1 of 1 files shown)


natefoo:
Closing, it looks like setup.cpyext.py is superseded by [setup.pyuwsgi.py](https://github.com/unbit/uwsgi/blob/adb5436b1cc0075f275264781439d3157dbb9475/setup.pyuwsgi.py).

PR #1864, opened 2018-09-10
Packaging fixes

ipmb:
A couple tweaks prior to pushing to PyPI.

### DIFF
--- a/setup.pyuwsgi.py
+++ b/setup.pyuwsgi.py
@@ -112,5 +112,7 @@ setup(
     ext_modules=[
         Extension(uWSGIBuildExt.UWSGI_NAME, sources=[]),
         ],
-    scripts=['bin/pyuwsgi'],
+    entry_points={
+        'console_scripts': ['pyuwsgi=%s:run' % uWSGIBuildExt.UWSGI_NAME],
+        },
     )
--- a/setup.pyuwsgi.py
+++ b/setup.pyuwsgi.py
@@ -97,7 +97,7 @@ class uWSGIBuildExt(build_ext):
 
 
 setup(
-    name='uWSGI',
+    name='pyuwsgi',
     license='GPL2',
     version=uwsgiconfig.uwsgi_version,
     author='Unbit',
--- a/setup.pyuwsgi.py
+++ b/setup.pyuwsgi.py
@@ -110,9 +110,9 @@ setup(
         'uwsgidecorators',
         ],
     ext_modules=[
-        Extension(uWSGIBuildExt.UWSGI_NAME, sources=[]),
+        Extension('pyuwsgi', sources=[]),
         ],
     entry_points={
-        'console_scripts': ['pyuwsgi=%s:run' % uWSGIBuildExt.UWSGI_NAME],
+        'console_scripts': ['pyuwsgi=pyuwsgi:run'],
         },
     )
--- a/bin/pyuwsgi
+++ b/bin/pyuwsgi
@@ -1,4 +1,0 @@
-#!/usr/bin/env python
-
-import pyuwsgi
-pyuwsgi.run()


setup.pyuwsgi.py
```diff
@@ -110,7 +110,9 @@ def uwsgi_build(self):
         'uwsgidecorators',
         ],
     ext_modules=[
-        Extension(uWSGIBuildExt.UWSGI_NAME, sources=[]),
+        Extension('pyuwsgi', sources=[]),
         ],
-    scripts=['bin/pyuwsgi'],
+    entry_points={
+        'console_scripts': ['pyuwsgi=pyuwsgi:run'],
```
nsoranzo: @ipmb @unbit Would you consider changing the entry point name to `uwsgi` (or adding a second one called `uwsgi`)? We ([Galaxy](https://github.com/galaxyproject/galaxy/)) would like to move to pyuwsgi without breaking existing deployments.

CC: @natefoo

ipmb: I'd consider it, but I have some hesitation about "stealing" that script name. Curious to hear what others think.

@mattrobenolt would this be of any help for you wrt https://github.com/unbit/uwsgi/issues/1218#issuecomment-463679113

mattrobenolt: I haven't looked into this deeper yet, so my context is very limited, but if it acts and represents `uwsgi` the binary and is identical aside from packaging, yes. The way we do things in Sentry is by `exec` out to uwsgi.

https://github.com/getsentry/sentry/blob/master/src/sentry/services/http.py#L165

So if I can maintain this identical behavior, I'd be more inclined to test this out.

I think the only counter argument would be that this directly conflicts with `uwsgi` package itself, in which case, I think isn't relevant anyways since you shouldn't have both installed.

nsoranzo:
`uwsgi` and `pyuwsgi` kind-of "conflict" already in the sense that both install `uwsgidecorators.py` in the same position, e.g. `.venv/lib/python2.7/site-packages/uwsgidecorators.py`

setup.pyuwsgi.py
nsoranzo: `uwsgi` and `pyuwsgi` kind-of "conflict" already in the sense that both install `uwsgidecorators.py` in the same position, e.g. `.venv/lib/python2.7/site-packages/uwsgidecorators.py`

xrmx: @nsoranzo wouldn't  a rename break deployments for other people though? Will `ln -s pyuwsgi uwsgi` work?

nsoranzo:
@xrmx pyuwsgi has just been released, so that's probably not a big issue if we change it now.
Any way, as I mentioned above, adding a second identical `uwsgi` entry point would also work for us.

setup.pyuwsgi.py
ipmb: @mattrobenolt tangential, but with `pyuwsgi`, you could get rid of the `execvp` and your logic to detect the virtualenv. All that would be needed is:

```python
import pyuwsgi
pyuwsgi.run()
```

Here's an example of running it via a Django management command, https://github.com/lincolnloop/django-webserver/blob/cae80046baa5f72d76f42c213c55acf7353db64c/django_webserver/management/commands/pyuwsgi.py

xrmx:
@nsoranzo unless i am missing something, pyuwsgi is out since at least 5 months

setup.pyuwsgi.py
mattrobenolt: That's rad. I likely won't change to maintain compatibility, but I will probably use it on newer setups. :)

With that said, my last opinion here for uniformity is just that. I believe that if this package is meant to do what it's intending to do, it'd be an absolute drop in replacement in all ways, with the enhancement of packaging, and maybe other stuff in addition. But for it to be successful, imo, it should try to be a drop in as much as possible.

I've experienced success with psycopg2 in this regard with [psycopg2-binary](https://pypi.org/project/psycopg2-binary/). I'd imagine this is a similar goal, in which I fully support and would adopt.

ipmb:
@xrmx it's trivial to have it install scripts for both `pyuwsgi` and `uwsgi`. I'm just looking for feedback on if it is reasonable to take over the `uwsgi` name when installed. I see arguments both ways.

setup.pyuwsgi.py
ipmb: Ok, I'm convinced. I just cut a new release which can be executed as either `uwsgi` or `pyuwsgi` https://pypi.org/project/pyuwsgi/2.0.18.post0/

nsoranzo: Thanks @ipmb, much appreciated!

PR #1866, opened 2018-09-11
pyuwsgi more like default.ini and offer no SSL

ipmb:
@unbit let me know if there is a better way to do this

pyuwsgi will be the default for anyone building the wheel from an sdist
on PyPI, but we can build the wheels for distribution with
UWSGI_PROFILE=pyuwsginossl to avoid the security quagmire of bundling SSL.


ipmb:
Another thought I had is to just take the changes for `pyuwsgi.ini` and I can generate a non-ssl one during the wheel build process if you don't want to clutter up `buildconf` any more.

PR #1871, opened 2018-09-20
Fix for respawn with fork-server and emperor

Miniwoffer:
Fix for respawn with fork-server and emperor
A fix for emperor cursing vassels on respawn
The emperor will curse forked vassals when they
attempt a respawn, this will in turn cause the
vassal to get killed when it is finished
respawning. To prevent this emperor now checks
if it uses a fork-server and if the last_mod
on the vassal is less then 1 sec old.


unbit:
Thanks!

Issue #1471, opened 2017-02-17
Reloading through SIGHUP kills uwsgi, unable to chown socket


AMDmi3:
uwsgi running like this:

```
/usr/local/bin/uwsgi -M --manage-script-name --mount /=repology-app:app --pythonpath=/var/db/repology/repology --pidfile /var/run/uwsgi-repology.pid -s /var/run/repology/uwsgi.socket --chmod-socket=660 --chown-socket=www:www -d /var/log/uwsgi-repology.log --uid repology --gid repology
```
The config is pretty common - application is running under its own dedicated user ```repology```, and socket is owned by ```www``` as it should be accessed by frontend (nginx running as ```www```).

when issuing SIGHUP, it dies with
```
chown(): Operation not permitted [core/utils.c line 2623]
```
in the log, likely because it's already running as ```repology``` and is thus unable to change socket ownership to ```www```.

- Likely it should check socket permissions and don't try to change them if they are already correct.
- Definitely it should not die from SIGHUP, like all well-behaving servers. Such errors should not be considered fatal.

feld:
I'm seeing this as well. With the defaults on FreeBSD now uwsgi is started by root, creates the socket as www:www, and drops privileges to uwsgi user. This gives us nice isolation, but a SIGHUP causes it to close the unix and network sockets and attempt to re-open them. It fails because it no longer has root privileges so it can't recreate a socket owned by www:www, then terminates.

I don't understand why a reload requires destroying the sockets...

xrmx:
@AMDmi3 @feld what version of uwsgi is this?

AMDmi3:
2.0.14

feld:
Any updates on this?

AMDmi3:
Note that above mentioned FreeBSD ports commits didn't fix this.

DuncanFairley:
The socket permissions were screwed for me upon updating to 2.0.15 from 2.0.13 on FreeBSD.

theory14:
I have the same problem when I moved to 2.0.15 on FreeBSD 11.1 RELEASE.  I solved this just now by setting uwsgi to run as www and the socket owner to www:
uwsgi_uid="www"
uwsgi_gid="www"
uwsgi_socket_owner="www"

The socket is then owned by www and the SIGHUP reload works as expected.

Setting those to default UID to uwsgi and socket_owner to uwsgi results in failure.  The socket is owned by uwsgi and the uwsgi processes are running as uwsgi.

[...]

PR #1876, opened 2018-09-25
Fix Unix socket inheritance after reload on FreeBSD

citrin:
On FreeBSD Unix socket path returned by getsockname() in
uwsgi_add_socket_from_fd() can contain garbage trailing character(s).

Clear a structure before passing it to getsockname() as a workaround.

Problem with reloading through SIGHUP on FreeBSD was also reported in #1471

### DIFF
--- a/core/socket.c
+++ b/core/socket.c
@@ -1089,6 +1089,7 @@ void uwsgi_add_socket_from_fd(struct uwsgi_socket *uwsgi_sock, int fd) {
 	union uwsgi_sockaddr usa;
 	int abstract = 0;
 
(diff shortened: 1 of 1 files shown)


citrin:
As I can see in other code bases, clearing of structure before passing it to getsockname() is rarely done. But it can be found e. g. in libevent code: https://github.com/libevent/libevent/blob/29cc8386a2f7911eaa9336692a2c5544d8b4734f/sample/http-server.c#L387 though there is no comment why this was done.

rdeioris:
thanks a lot. @xrmx safe for backport

PR #1890, opened 2018-10-12
Calculate avg_response_time metric with right way

pavelpatrin:
Calculate avg_response_time metric with right way

Instead of this master version of metric calculates via avg(average, current).

For more info see example at https://www.geeksforgeeks.org/average-of-a-stream-of-numbers/


pavelpatrin:
I test it completely and next reopen.

pavelpatrin:
I am confused. Fixing this bug makes metric `avg_response_time` slightly changable when uwsgi works longer and longer.

Maybe metric `avg_response_time` is useless and we really needed in `moving_avg_response_time` metric calculated on last N minutes?

pavelpatrin:
TRG-41291

pavelpatrin:
We test this patch and there was no any positive changes in metrics.

PR #1896, opened 2018-10-16
Fixed builds against XCode 10

sergeykolosov:
The issue is `-mmacosx-version-min=10.5` which expects `libgcc_s.10.5` presence,
which isn't there in XCode 10.

The correct behavior is achieved by bumping the `-mmacosx-version-min` to 10.9
for releases of MacOS that support XCode 10 (High Sierra and above).

See also:
https://github.com/Homebrew/brew/pull/4941
https://github.com/Homebrew/brew/pull/4810


xrmx:
Looks good, thanks!

PR #1913, opened 2018-10-28
Fix multithreading in mono plugin

Evengard:
Let Mono handle the needed signals. Without theese it may end up in SIGSEGV or just hang a thread.


Evengard:
After testing I went live with this patch. I will report back the behaviour if there is any problems.

Evengard:
Added some more signals as per https://www.mono-project.com/docs/advanced/embedding/

Evengard:
Added unblocking of signals for the main thread too, as some threads ended up in "zombie" state because of Mono handling SIGCHLD.

PR #1915, opened 2018-10-29
Mono - make application automatically reload after unload

Evengard:
Make the application automatically reload after it gets unloaded (for example because of file edit) - almost IIS-like experience.


Evengard:
Kinda reviewed and normalized the whole file (which was in a mess even before me). Hope you don't dislike tab indentation, anyway I can replace it with simple find/replace to spaces if you need it.

PR #1917, opened 2018-10-30
ssl-option can't be set

mathieu-lacage:
The ssl-option variable can't be set: it is designed to take a numeric argument but is specified as a "no_argument" option which results in an error when trying to set this option. This changeset sets the proper type.

Given the changeset which introduced this option with the ssl-enable3 option which is of type no_argument, I suspect someone copy/pasted the ssl-enable3 configuration file without testing the change.


unbit:
Thanks, yes that options was added when the disabling of ssl3 became mandatory :)

PR #1923, opened 2018-11-06
Mono fix not full data sent

Evengard:
I decided to block the signals from Mono prior to drop to read/write routines. I'm not sure that I should rewrite the core, so this way we stay confined in the mono plugin boundaries, not breaking anything else.

Also I synchronized correctly the HttpRuntime.ProcessRequest (it CAN be asynchronous), and passed the current_wsgi_req to the Mono request object, because I've seen in the logs some weird calls with "detached" current_wsgi_req - I guess an ASP NET application may run some subroutines in its own threads.

### DIFF
--- a/plugins/mono/mono_plugin.c
+++ b/plugins/mono/mono_plugin.c
@@ -65,9 +65,11 @@ struct uwsgi_mono {
 	MonoClass *byte_class;
 
 	MonoClassField *filepath;
+	
+	MonoClassField *request;
 
 	// thunk
-	void (*process_request)(MonoObject *, MonoException **);
+	void (*process_request)(MonoObject *, void* req, MonoException **);
 
 	struct uwsgi_string_list *app;
 	struct uwsgi_string_list *exec;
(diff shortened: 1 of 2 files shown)


Evengard:
After all, I decided to rewrite the poll calls. I've got some hanging threads here and there, so maybe (just maybe!) blocking the signals in the actual poll call may help. If not, well then I guess we will need to somehow intercept them and pass to uwsgi_signal_handler... Unsure, will test more.

Evengard:
And another change. Now I do not try to block the signals at all. Just let them get through and even try to parse. If we don't have any signal actually, just loop back. Seems to work well for now. Will test more.

unbit:
Hi, i think the best thing to do is adding a "uwsgi_mono_protected_poll()" wrapper for your poll implementation that you can use in the mono plugin. Changing uWSGI internals for LTS (as i want to backport your changes to 2.0) is (generally) not allowed.

Evengard:
Why do I think that it should be included in the core: this method is basically a default one for dealing with signal handlers during poll. This problem may arise with any other signal, handled by any other plugin. I really think that the core should be fail-proof about theese things.

Also, I am not really sure about how should I deal with the poll wrapper, because the actual poll call is used deep in the core, and I don't see a way to "plug it in".

PR #1961, opened 2019-01-12
systemd_logger: Don't generate build warnings when all is OK

cjmayo:
libsystemd-journal is only found with old versions of systemd - check it
is installed before querying.

---

These messages are found in the build.log with a current systemd version:

```
Package libsystemd-journal was not found in the pkg-config search path.
Perhaps you should add the directory containing `libsystemd-journal.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libsystemd-journal' found
Package libsystemd-journal was not found in the pkg-config search path.
Perhaps you should add the directory containing `libsystemd-journal.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libsystemd-journal' found
```
This change prevents them and stops the user having to think whether they matter.

### DIFF
--- a/plugins/systemd_logger/uwsgiplugin.py
+++ b/plugins/systemd_logger/uwsgiplugin.py
@@ -2,9 +2,13 @@ import os
 
 NAME = 'systemd_logger'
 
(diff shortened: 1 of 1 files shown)


xrmx:
Looks good, thanks!

PR #1970, opened 2019-01-25
emperor: fix uwsgi_emperor_spawn_vassal protype

rbilovol:
clone() expects 'int (*fn)(void *)' as his argument
whereas uwsgi_emperor_spawn_vassal() is of type void.

This fixes build issue with recent gcc:

| core/emperor.c: In function 'uwsgi_emperor_vassal_start':
| core/emperor.c: error: cast between incompatible function types from 'void (*)(struct uwsgi_instance *)' to 'int (*)(void *)' [-Werror=cast-function-type]

Signed-off-by: Ruslan Bilovol <rbilovol@cisco.com>

### DIFF
--- a/core/emperor.c
+++ b/core/emperor.c
@@ -1193,7 +1193,7 @@ void emperor_add_with_attrs(struct uwsgi_emperor_scanner *ues, char *name, time_
 	}
 }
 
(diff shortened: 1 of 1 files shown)


xrmx:
Thanks!

PR #2007, opened 2019-05-01
core/uwsgi: Fix compilation with uClibc-ng

neheb:
uClibc-ng by default does not compile in execinfo support.


xrmx:
Merged, thanks!

PR #2008, opened 2019-05-03
Remove HTTPS_CLIENT_CERTIFICATE in favor of HTTPS_CC

paultag:
HTTPS_CLIENT_CERTIFICATE contains the DER encoded Client Certificate.

f065171aa14120c66e1ebb6076ab6d53dd070dc1 adds in HTTPS_CC, which is
a PEM encoded Client Certificate. For Python 3, the HTTPS_CLIENT_CERTIFICATE
field was read as a `string`, not `bytes`, which leads to the runtime
(rightfully!) complaining.

Since PEM encoding the field is a better idea, I'm removing this code in
favor of just using --https-export-cert. This was only usable in Python
2, so this is of diminishing utility as time goes on.


paultag:
This is an easy one if you have cycles @unbit

PR #2029, opened 2019-06-08
add mules add stats / add memory info to mules

awelzel:
Hello,

we're interested in having some basic mule stats fro the stats server output.
This PR adds this information. When the `--memory-report` option is set, a thread is started in each mule process collecting memory information every `mem-collector-freq` seconds. If `--memory-report` is not provided, the `rss` and `vsz` fields will be `0` in the stats output.

Following an example of the additional entries in the stats server output:
```
...
        ],
        "mules":[
                {
                        "id":1,
                        "pid":28214,
                        "signals":0,
                        "mode":"programmed",
                        "patch":"hello:mule",
                        "rss":14692352,
                        "vsz":142168064
                },
                {
                        "id":2,
                        "pid":28215,
                        "signals":0,
                        "mode":"programmed",
                        "patch":"./moo.py",
                        "rss":15179776,
                        "vsz":142172160
                },
                {
                        "id":3,
                        "pid":28216,
                        "signals":165,
                        "mode":"signal",
                        "rss":14798848,
                        "vsz":142168064
                }
        ]
}
```

Let me know if this is something you'd consider useful enough to be merged. I'd be also happy to hear about different ways to collect the memory usage, or whether a different approach should be used. Thanks!

There's a minor cleanup commit for the uwsgi python module related to message sending.

### DIFF
--- a/core/mule.c
+++ b/core/mule.c
@@ -32,6 +32,24 @@ int mule_send_msg(int fd, char *message, size_t len) {
 	return 0;
 }
 
+// Runs in a mule thread and periodically scans for memory usage.
+static void *mule_mem_collector(void *foobar) {
+	// block all signals
+        sigset_t smask;
+        sigfillset(&smask);
+        pthread_sigmask(SIG_BLOCK, &smask, NULL);
+	int id = uwsgi.muleid;
+	uwsgi_log_verbose("mem-collector thread started for mule %d\n", id);
+	for(;;) {
+		sleep(uwsgi.mem_collector_freq);
+		uint64_t rss = 0, vsz = 0;
+		get_memusage(&rss, &vsz);
+		uwsgi.mules[id - 1].rss_size = rss;
+		uwsgi.mules[id - 1].vsz_size = vsz;
+	}
+	return NULL;
+}
+
 void uwsgi_mule(int id) {
 
 	int i;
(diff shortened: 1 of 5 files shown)


core/mule.c
```diff
@@ -32,6 +32,24 @@ int mule_send_msg(int fd, char *message, size_t len) {
...
+static void *mule_mem_collector(void *foobar) {
+	// block all signals
+        sigset_t smask;
+        sigfillset(&smask);
+        pthread_sigmask(SIG_BLOCK, &smask, NULL);
+	int id = uwsgi.muleid;
+	uwsgi_log_verbose("mem-collector thread started for mule %d\n", id);
+	for(;;) {
+		sleep(uwsgi.mem_collector_freq);
+		uint64_t rss = 0, vsz = 0;
```
xrmx: this looks like a fix we should backport to the worker mem collector as on linux at least if there's not /proc/self/stat we are dereferencng that pointer

awelzel: Hello @xrmx ,

thanks for checking in. I've fixed the spaces and another fix related to the default frequency.

I'm not exactly clear what issue is solved that you'd want to backport - I have literally (including spaces) copied the mem collector from `core/uwsgi.c:3424`, but decided to always run it in a mule, not just when `-evil-reload-on-{as,rss}`.

Mind pointing me in the right direction?

Thanks,
   Arne

xrmx: See get_memusage:

```
        procfile = fopen("/proc/self/stat", "r");
        if (procfile) {
                i = fscanf(procfile, "%*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %*s %llu %lld", (unsigned long long *) vsz, (unsigned long long *) rss);
                ...
        }
        *rss = *rss * uwsgi.page_size;
```
And mem_collector:

```
                uint64_t rss, vsz;
                get_memusage(&rss, &vsz);
```

If we cannot read procfile rss is garbage

xrmx: Actually i was looking at uwsgi-2.0 branch which missed some backports for some stuff already fixed in master :)

PR #2041, opened 2019-07-14
fix Python 3.7 warning in pyuwsgi.c

orivej:
Python 3.7 made `PyUnicode_AsUTF8` return const: https://bugs.python.org/issue28769

---

This PR is for `uwsgi-2.0` branch. It is based on 3dab0e4e67c763793c372acae7a9a722dd362fc1 that did not reach the `master` branch.


unbit:
Maybe it is better to have an additional ifdef for 3.7 ?

xrmx:
@unbit it's only used for strlen which should be already doing the char * -> const char * promotion.

unbit:
@xrmx Right, thanks

PR #2052, opened 2019-08-04
Add PLUGIN_BASE_DIR variable

aduskett:
Currently, when cross-compiling, if the plugin_dir points to the target
directory, uwsgi will embed the full path during compiling. This whole path
results in uwsgi trying to load a full target path instead of /usr/lib/uwsgi
when running on the target.

For example with Buildroot, instead of using /usr/lib/uwsgi as the plugin_dir,
uwsgi would try to use /path/to/buildroot/output/target/usr/lib/uwsgi.

Creating a new PLUGIN_BASE_DIR variable and prefixing plugin_dir allows the
plugin to be installed to the appropriate directory but still have uwsgi load
the plugins from the correct folder when ran from on the cross-compiled target.

### DIFF
--- a/uwsgiconfig.py
+++ b/uwsgiconfig.py
@@ -27,7 +27,7 @@ try:
 except ImportError:
     import configparser as ConfigParser
 
-
+PLUGIN_BASE_DIR = os.environ.get('PLUGIN_BASE_DIR', '')
 PY3 = sys.version_info[0] == 3
 
 if uwsgi_os == 'Darwin':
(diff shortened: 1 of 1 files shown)


uwsgiconfig.py
```diff
@@ -27,7 +27,7 @@
 except ImportError:
     import configparser as ConfigParser
 
-
+PLUGINDIR_BASE = os.environ.get('PLUGIN_DIR_BASE', '')
```
xrmx: commit mesage say PLUGIN_BASE_DIR, code use PLUGIN_DIR_BASE and variable name is PLUGINDIR_BASE :) Can we use PLUGIN_BASE_DIR everywhere please?

aduskett: Ah! Yeah, of course. It was a long day. I will change that now.

aduskett: And done. I also rebased to make the commit history a bit cleaner.

PR #2057, opened 2019-08-12
python: fix object leak in uwsgi.workers()

awelzel:


awelzel:
@xrmx - is this something you'd pick up? Happy to adapt the testing part if you have a better idea here.

tests/testworkers.py
```diff
@@ -0,0 +1,26 @@
...
+Regression test for #2056 - uwsgi.workers() leaking objects.
+"""
+import uwsgi
+import gc
+
+
+def application(env, start_response):
+    gc.collect()
+    start_objs = len(gc.get_objects())
+
```
xrmx: What about adding here something like the following to check that we are actually returning something?

```
    workers = uwsgi.workers()
    assert workers
    for w in workers:
        assert w['apps']
```

awelzel: I had that at first, but then removed it to limit it to just test for the leaking part - added back now ;-)

xrmx:
Thanks @awelzel !

PR #2062, opened 2019-08-25
Python 3 compat

thomasgoirand:
Hi,

I'm maintaining uwsgi in Debian together with Jonas Smedegaard. We're planning to remove Python 2 from Debian Bullseye, and therefore, uwsgi also has to get rid of any Python 2 traces.

This patch fixes a lot of issues regarding Python 3 compatibility. Note that this also add a new six depends, which I think is pretty fine these days, but which I wasn't able to express (where the hell is it hidden in this setup.py thing?).

Please get this merged, so we can use it in Debian too, and get the package fixed.


contrib/spoolqueue/tasks.py
```diff
@@ -3,9 +3,9 @@
 
 @queueconsumer('fast', 4)
 def fast_queue(arguments):
-    print "fast", arguments
```
xrmx: without the from future import you are printing a tuple

aduskett:
With python 2 being EOLed next month this is going to be a major issue very soon!

thomasgoirand:
The above from __future__ import print_function was added. Please merge!

awelzel:
Hello @thomasgoirand ,

I do not know the plans for Python 2 and uwsgi, but have a naive question related to the following:

> I'm maintaining uwsgi in Debian together with Jonas Smedegaard.
> We're planning to remove Python 2 from Debian Bullseye, and therefore,
> uwsgi also has to get rid of any Python 2 traces.

The majority of your changes concern examples and tests - even if they are not Python 3 compatible, uwsgi would still be able to run in a Python3 only environment. I can see that it would be nice to have them Python 3 compatible as they appear to be put into `/usr/share/doc`. I'm mainly wondering if Debian Bullseye requires the whole source tree to be Python 2 free ?

Thanks,
   Arne

jonassmedegaard:
Hi @awelzel ,

Debian do _not_ require whole source trees to be void of Python2.  What is needed is for the source to properly work without Python2 neither on build system nor on runtime system.

thomasgoirand:
Hi,
Indeed, Debian doesn't require the whole source tree to be py3 compatible. But that's not what I'm proposing as a patch. I'm proposing a patch so that all remaining Py2 only code becomes py2 AND py3 compatible, because that's so much nicer for everyone. Why holding it off then?

Issue #1781, opened 2018-05-03
customize file mode for logfile plugin


dmrzzz:
Feature request: please make it possible to configure the mode for new files created by the logfile plugin (currently hardcoded to `S_IRUSR | S_IWUSR | S_IRGRP` in plugins/logfile/logfile.c)

I'm using the logfile plugin in order to separate request log from error log:

```
  logger: file:logfile=%(X_LOGDIR)/error.log,maxsize=2000000
  req-logger: file:logfile=%(X_LOGDIR)/access.log,maxsize=2000000
  logformat: %(addr) - %(user) [%(ltime)] "%(method) %(uri) %(proto)" %(status) %(size) "%(referer)" "%(uagent)" msecs=%(msecs) tmicros=%(tmicros)
```

but would like both sets of logfiles to be world-readable (not just group-readable), and the `logfile-chmod` option doesn't appear to help presumably because I'm using a plugin.

At the moment I'm clumsily working around this with a cron task to chmod each file after the fact.

Issue #1525, opened 2017-05-05
logfile has wrong permission when set both logfile-chmod and log-maxsize


wgjak47:
Here is my log config in uwsgi.ini
```
how-config=true
disable-logging=false
logfile-chmod=644
logfile-chown=wgj
log-maxsize=500000000
logto=%(chdir)logs/uwsgi_access.log
logger = longquery file:%(chdir)logs/uwsgi_long.log
log-route = longquery msec
```

The first log file is fine, But when it exceed 500MB, the new logfile permission is wrong like this:
```
-rw-r----- 1 wgj      wgj   314009695 5月   5 09:47 uwsgi_access.log
-rw-r--r-- 1 wgj      wgj   500001871 4月  29 01:52 uwsgi_access.log.1493401977
-rw-r----- 1 wgj      wgj   500003761 4月  30 22:50 uwsgi_access.log.1493563839
-rw-r----- 1 wgj      wgj   500000841 5月   2 20:13 uwsgi_access.log.1493727188
-rw-r----- 1 wgj      wgj   500001200 5月   4 10:56 uwsgi_access.log.1493866560
```
The read permission for other users is missing...

The function void uwsgi_log_do_rotate  in core/logging.c
```
int fd = open(logfile, O_RDWR | O_CREAT | O_TRUNC, S_IRUSR | S_IWUSR | S_IRGRP);
```
not open the logfile with the logfile-chmod options...

erezarnon:
Any update here?

PR #2071, opened 2019-09-08
Add mode option to `logfile` plugin

jacobtolar:
* Add a new option to the `logfile` plugin to accept a `mode` argument, used to set flags when creating a new log file.
* Add stricter validation to `core/utils.c:uwsgi_mode_t`.
* Update argument parsing for `chmod-socket` and `logfile-chmod` to call `uwsgi_mode_t`.

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -4558,22 +4558,25 @@ void uwsgi_setns(char *path) {
 mode_t uwsgi_mode_t(char *value, int *error) {
 	mode_t mode = 0;
 	*error = 0;
+	int len = strlen(value);
 
-        if (strlen(value) < 3) {
+	if (len < 3) {
 		*error = 1;
 		return mode;
 	}
 
-        if (strlen(value) == 3) {
-                mode = (mode << 3) + (value[0] - '0');
-                mode = (mode << 3) + (value[1] - '0');
-                mode = (mode << 3) + (value[2] - '0');
-        }
-        else {
-                mode = (mode << 3) + (value[1] - '0');
-                mode = (mode << 3) + (value[2] - '0');
-                mode = (mode << 3) + (value[3] - '0');
-        }
+	int i, start = 0;
+	if (len != 3) {
+		start = 1;
+	}
+
+	for (i = start; i < start + 3; i++) {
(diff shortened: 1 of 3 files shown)


jacobtolar:
Should address issues like these

https://github.com/unbit/uwsgi/issues/1781
https://github.com/unbit/uwsgi/issues/1525

awelzel:
Hey @jacobtolar , I took a quick look.

Two thoughts:
 1) There's the check subdirectory for unit tests using libcheck - uwsgi_mode_t seems like a good candidate: In 371da50 I tried to add some basic ones - feel free to cherry-pick or copy paste.

 2) In core/logging.c:uwsgi_do_log_rotate the logfile is reopened after renaming it with a hard-coded mode - you probably want to pass the mode to that call now. (In fact, I'm a bit surprised that doesn't try to use `uwsg.chmod_logfile_value` currently.

uwsgi.h
```diff
@@ -4894,7 +4894,7 @@ void uwsgi_master_fifo_prepare();
 int uwsgi_master_fifo();
 int uwsgi_master_fifo_manage(int);
 
-void uwsgi_log_do_rotate(char *, char *, off_t, int);
+void uwsgi_log_do_rotate(char *, char *, off_t, int, mode_t);
```
jacobtolar: Is this expected to be a public interface/maintain compatibility with any other existing code? If so I could create a new wrapper function and leave the original function signature alone.

xrmx: I'd say we should not worry https://github.com/unbit/uwsgi/commit/01a5611159825437235e7d98b8e0687937b523f8

xrmx:
Thanks for the PR, would be possible to extract the uwsgi_mode_t changes in another commit with @awelzel tests please, and then add the new option on top of that? I can cherry-pick the tests (except the one that tests the validation of course) if that would make it easier.

rammanoj:
Hi @xrmx/ @jacobtolar/ @awelzel, I am able to reproduce the similar issue in the latest version (2.0.19.1). Is there any other way through which I can accomplish this task ?

PR #2077, opened 2019-10-03
Enable python plugin to load and run compiled python files (.so).

Fran6nd:
# Introduction.
The core idea of this PR is to enable uwsgi to run with compiled python modules (via tools like [Nuitka](https://nuitka.net/)).
So we could improve performances while beeing able to ship professionnal uwsgi-based python solution without shipping the source code and without allowing the client to edit it (Wich is the worst nightmare of any support team!).

# How to use?

## Compiling a python file.
Here we will compile the "Wello World" example from [uwsgi documentation](https://uwsgi-docs.readthedocs.io/en/latest/WSGIquickstart.html).

```python
#foobar.py
    return [b"Hello World"]
```
Installing nuitka:
`python -m pip install nuitka`

Now let's build it:
`python -m nuitka --module --follow-imports foobar.py`

The so file is now created.
Then remove the python file:
`rm foobar.py`

Note: For multiple python file, we can basically compile each of them as module, it will be easier to maintain and debug than a single compiled module.

## Using it.
Basically as follow:
`uwsgi --http :9090 --wsgi-file foobar.so`
So now you can check it works with curl or any web browser (localhost):
`curl http://127.0.0.1:9090`

# How does it works?

Python allows natively loading compiled modules basically with `import my_module`. So basically we will create a temp python file containing one line of code: `from my_module import *`, run it and then delete it because Python interpreter cannot run directly compiled modules...
So the temp file is just importing all the module's method inside the uwsgi context.
So it will cost a bit more perfs on application startup, however it should be faster once started.

That's my first contribution to this project (and one of my first on GitHube)... I hope I didn't break any rules! (If so feel free to tell me!)
Waiting for your comments!

### DIFF
--- a/plugins/python/pyloader.c
+++ b/plugins/python/pyloader.c
@@ -580,6 +580,7 @@ PyObject *uwsgi_dyn_loader(void *arg1) {
 PyObject *uwsgi_file_loader(void *arg1) {
 
 	char *filename = (char *) arg1;
+	char tmp_filename[100] = {0};
 	PyObject *wsgi_file_module, *wsgi_file_dict;
 	PyObject *wsgi_file_callable;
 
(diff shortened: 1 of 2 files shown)


Fran6nd:
If anyone could take a look... would be great!

xrmx:
@Fran6nd you just closed the PR...

Fran6nd:
Ooops miss click!

awelzel:
@Fran6nd - do you know that your example works with vanilla `uwsgi` like this?
```
./uwsgi --http-socket :9090  --plugin ./python37_plugin.so --module foobar
```

awelzel:
Another comment - I initially tested a different example which uses `yield` instead of returning a list of `bytes`:
```
print("nuitka test", __file__, "imported")

    start_response('200 OK', [('Content-Type', 'text/plain')])
    yield b"hello\n"
```

`uwsgi` starts fine but explodes on the first request - that's probably some nuitka issue:

```
spawned uWSGI worker 1 (and the only) (pid: 21804, cores: 1)
!!! uWSGI process 21804 got Segmentation Fault !!!
*** backtrace of 21804 ***
./uwsgi(uwsgi_backtrace+0x2a) [0x55e5545efd3a]
./uwsgi(uwsgi_segfault+0x23) [0x55e5545f0163]
/lib/x86_64-linux-gnu/libc.so.6(+0x37840) [0x7f9208a28840]
.(+0x6efa) [0x7f9205ed4efa]
.(+0x104d3) [0x7f9205ede4d3]
/usr/lib/x86_64-linux-gnu/libpython3.7m.so.1.0(PyIter_Next+0xb) [0x7f9206675e8b]
./python37_plugin.so(uwsgi_response_subhandler_wsgi+0x4a) [0x7f920693e5ba]
./python37_plugin.so(uwsgi_request_wsgi+0x168) [0x7f920693cfe8]
./uwsgi(wsgi_req_recv+0xae) [0x55e55459dffe]
./uwsgi(simple_loop_run+0xc8) [0x55e5545eb9e8]
./uwsgi(simple_loop+0x10) [0x55e5545eb7f0]
./uwsgi(uwsgi_ignition+0x292) [0x55e5545f04b2]
./uwsgi(uwsgi_worker_run+0x276) [0x55e5545f3ac6]
./uwsgi(uwsgi_run+0x442) [0x55e5545f4032]
./uwsgi(+0x2c63e) [0x55e55459d63e]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xeb) [0x7f9208a1509b]
./uwsgi(_start+0x2a) [0x55e55459d66a]
*** end of backtrace ***
```

outscale-fne:
@awelzel thanks for answering... I'm checking it now!

awelzel:
```

**Comment by awelzel** [comment]
```
$ nuitka3 --version
0.6.5
Python: 3.7.3 (default, Apr  3 2019, 05:39:12) 
Executable: /tmp/venv/bin/python3
OS: Linux
Arch: x86_64
```
I don't think this is a `uwsgi` issue though and maybe it can be discussed elsewhere.

outscale-fne:
I've been working around the yield issue... As you said it seems to be a Nuitka issue... I'll close this PR now since the vanilla version can load modules. (I didn't find it in doc...)

PR #2079, opened 2019-10-08
plugins/cgi: adds dontresolve option

Ansuel:
This option permit to call the simbolic link instead of the file the simbolic link points.
All the security check are still done as the simbolic path is passed at the end after all the checks are passed. This is useful if some cgi app are used for multiple function based on the name they are called by.

Signed-off-by: Ansuel Smith <ansuelsmth@gmail.com>

### DIFF
--- a/plugins/cgi/cgi_plugin.c
+++ b/plugins/cgi/cgi_plugin.c
@@ -22,6 +22,7 @@ struct uwsgi_cgi {
 	int do_not_kill_on_error;
 	int async_max_attempts;
 	int close_stdin_on_eof;
+	int dontresolve;
 } uc ;
 
 static void uwsgi_opt_add_cgi(char *opt, char *value, void *foobar) {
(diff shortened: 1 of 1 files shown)


rdeioris:
Thanks, can you move the "int dontresolve;" declaration at the end of the struct ? (not strictly required for plugins but i prefer to maintain ABI compatibility)

Ansuel:
@rdeioris you mean after
`{0, 0, 0, 0, 0, 0, 0},`

Doesn't that needs to be at the end of the struct to set end of it?

rdeioris:
Hi, i mean the field in the uwsgi_cgi struct

Ansuel:
Ow... ok sorry then... Just give me a minute

Ansuel:
Done

rdeioris:
@xrmx safe for backport

PR #2084, opened 2019-10-21
Add docker-based build

jacobtolar:
Having a simple way to build uwsgi locally in Docker with all
dependencies might be helpful. (It's at least been helpful for me 🙂).

This PR adds a Dockerfile (based on the Travis configuration)
and updates the Makefile to let you run `make docker-test`.

The first time will be slow as the image builds, but future runs
should be quick as the docker cache will be used.

It would be simple to update this and the .travis.yml to use the same script
for setting up the environment if that's preferable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,11 @@ tests:
 	$(PYTHON) uwsgiconfig.py --build unittest
 	cd check && make && make test
 
+docker-test:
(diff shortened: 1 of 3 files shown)


xrmx:
Not sure i like this :) Not sure it makes a lot of sense to add a Dockerfile for running tests if we are not going to use to in the CI. And I don't know how a Dockerfile fits well with travis CI.

PR #2087, opened 2019-10-22
Bit more friendly log messages

jjanssen:
Lately I was debugging an application on the environment of a 3rd party when I saw some remarkable messaging coming by. Thought it could use a little more appropriate writing in case somebody who is not familiar with uwsgi finds this remarkable.


xrmx:
Looks good, thanks!

ThiefMaster:
booooooring.

PR #2089, opened 2019-10-25
register_rpc: fix overflow in rpc name

MRoci:


xrmx:
Merged and backported to uwsgi-2.0, thanks!

Issue #1453, opened 2017-01-19
Periodic error when reloading uwsgi with Master FIFO 


JonathanHuot:
Hi,
I have sometime errors when I'm sending a reload instruction to Master FIFO with uWSGI 1.9.18.2.

uWSGI configuration is :
```
uid = nginx
gid = nginx
master-fifo = /var/run/nginx/uwsgi.masterfifo
```

Permissions looks good :
```
ls -l /var/run/nginx/
total 4
drwxr-xr-x   2 nginx nginx 4096 Jan 19 04:31 .
drwxr-xr-x. 23 root  root  4096 Jan 19 15:21 ..
prw------- 1 nginx nginx 0 Jan 19 04:31 uwsgi.masterfifo
```

Command to send a reload (during logrotate) is :
`timeout 60 bash -c "echo 'r' > /var/run/nginx/uwsgi.masterfifo"`

Example of correct reload: 
```
1476156371 - ...gracefully killing workers...
1476156371 - Gracefully killing worker 2 (pid: 4059)...
1476156371 - Gracefully killing worker 3 (pid: 4060)...
1476156371 - Gracefully killing worker 1 (pid: 4058)...
1476156371 - worker 1 buried after 0 seconds
1476156371 - worker 2 buried after 0 seconds
1476156371 - worker 3 buried after 0 seconds
1476156371 - binary reloading uWSGI...
1476156371 - chdir() to /
1476156371 - closing all non-uwsgi socket fds > 2 (max_fd = 1024)...
(..)
```

Example of failed reload:
```
1476674915 - ...gracefully killing workers...
1476674915 - Gracefully killing worker 1 (pid: 32767)...
1476674915 - Gracefully killing worker 2 (pid: 32768)...
1476674915 - Gracefully killing worker 3 (pid: 32769)...
1476674915 - uwsgi_master_fifo()/mkfifo(): File exists [core/fifo.c line 86]
1476674915 - VACUUM: pidfile removed.
1476674915 - VACUUM WARNING: unix socket /var/run/nginx/uwsgi.socket changed inode. Skip removal
1476689381 - your processes number limit is 1024
1476689381 - your memory page size is 4096 bytes
(..)
```

If we look at `core/fifo.c:86`, we don't know if the unlink failed or the mkfifo, because we have none of the return codes are written into the logs.
I think, at minimum, uWSGI should write return codes of system calls and system libraries.

xrmx:
Please retry with latest stable release.

[...]

PR #2091, opened 2019-10-26
fifo: log non-ENOENT unlink() errors

awelzel:
Maybe this could help debugging #1453 .

### DIFF
--- a/core/fifo.c
+++ b/core/fifo.c
@@ -104,7 +104,9 @@ int uwsgi_master_fifo() {
 
 	char *path = uwsgi_fifo_by_slot();
 
-	unlink(path);
+	if (unlink(path) != 0 && errno != ENOENT) {
+		uwsgi_error("uwsgi_master_fifo()/unlink()");
+	}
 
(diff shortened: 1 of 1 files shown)


core/fifo.c
```diff
@@ -104,7 +104,12 @@ int uwsgi_master_fifo() {
 
 	char *path = uwsgi_fifo_by_slot();
 
-	unlink(path);
+	if (unlink(path)) {
+		if (errno != ENOENT) {
+			uwsgi_error("uwsgi_master_fifo()/unlink()");
+			// exit(1); // XXX: If unlink() failed, mkfifo() will probably as well
```
xrmx: Either exit() or don't, both are fine by me,  but let's not add commented out code please :)

xrmx: To be on the safe side better to don't exit though

awelzel: Sounds sensible - thanks. Also changed to `&& errno != ENOENT`.

xrmx:
Thanks!

Issue #1987, opened 2019-02-27
No app loaded when running php site on drupal 8.5 cms


shmel210:
Version of uwsgi:
`commit f17bdbd1b59565a0d6e81e7b825ee6e08ec8e29e
Author: Unbit <info@unbit.it>
Date:   Sat Feb 9 15:29:02 2019 +0100

    Update README
`
Emprerior config:
`
[uwsgi]
plugins-dir = /opt/uwsgi
emperor = /etc/uwsgi/vassals
uid = www-data
gid = www-data
logto = /var/log/uwsgi/emperor.log
check-static-docroot = true
emperor-stats = :1717
plugins-dir = /opt/uwsgi

`
Vassal config:
`
[uwsgi]
plugins-dir = /opt/uwsgi
plugins = php73
# variables
project_name = rud-garden.test
project_domain = %(project_name)
root = /home/shmel/projects
workdir = %(root)

# main config
chdir = %(root)/rud-garden.com/web
need-app = true
php-docroot = %(root)/rud-garden.com/web
php-index = index.php
php-allowed-ext = .php
php-allowed-ext = .inc
php-set = open_basedir=/tmp/:%(root)
php-set = post_max_size=100M
php-set = upload_max_filesize=100M
php-set = date.timezone=Europe/Kiev
#php-index = index.php
post-buffering = 1048576
http-socket-modifier1 = 14
#uid = nginx
#gid = nginx
master=true
lang = ru_RU.utf8
env = LANG = ru_RU.utf8

# control
pidfile = /run/uwsgi/%(project_name).pid
socket = /run/uwsgi/%(project_name).sock
stats = /run/uwsgi/%(project_name).stats.sock
chown-socket = www-data:www-data
chmod-socket = 664
#touch-reload = %(root)/logs/%(project_name).reload.txt
logto = /var/log/uwsgi/%(project_name).log
vacuum = true

# perfomance
cache2 = name=%(project_name).sessions,items=1000
php-set = session.save_handler=uwsgi
php-set = session.save_path=%(project_name).sessions
processes = %k
#enable-threads = true
#threads = 4
harakiri = 180
max-requests = 5000

`
Nginx config:
`
server {
    server_name rud-garden.test;
    charset UTF-8;
    set $domain rud-garden.test;
    set $user shmel;
    set $root_path /home/$user/projects/rud-garden.com/web;
    root $root_path;
    access_log /var/log/uwsgi/rud-garden.test.access.log;
[...]

PR #2106, opened 2019-12-08
php: Unset uwsgi.need_app in uwsgi_php_init() to keep running

awelzel:
Commit f2e0142 set the default for `uwsgi.need_app` to `1`. This results
in the termination of uWSGI when using the PHP plugin only as this plugin
does not register any app.

This feels a bit like a workaround.

I ran into this while attempting to test current master-tree (uwsgi-2.0 is not affected as f2e0142 isn't there) with the following PHP config file (and it was reported in #1987 as well).
```
[uwsgi]
master = true
workers = 16

plugins = http,php

http-modifier1 = 14
http = :8080

php-docroot = ./t/php/
php-index = test.php
```


gdamjan:
@xrmx please take a look at this too.

alternatively, if need-app is default now, the `--need-app` command option should be replaced with `--no-need-app`

Issue #2113, opened 2020-01-10
Python 3.8 Support


bpdev97:
Hi, I was unable to find anything related to python 3.8 support. Is python 3.8 support is already in the pipeline or can it already be used in some official capacity? Thanks!

awelzel:
Hey @bpdev97 ,

this is no statement about official support, but uwsgi builds and starts inside a `python:3.8-buster` docker container/environment, maybe give it a try?

When #2114 is merged Python 3.8 will be very "lightly" tested in the travis pipeline as well.

PR #2114, opened 2020-01-11
travis: Build plugins for Python 3.7 and 3.8

awelzel:
After issue #2113 came up - try building the Python plugin for Python 3.7 and 3.8.


unbit:
Thanks, @xrmx i think it should be backported

PR #2120, opened 2020-01-20
Add smart-daemon2 option to notify it on reloads

edufelipe:
We had a use-case of a smart-daemon that needed to be notified when the
uWSGI master was reloading.

That option needs to be explicitly enabled and requires the smart-daemon
to have a pidfile and a valid pid. It that is the case we send a signal
to it notifying the uWSGI master is reloading.


rdeioris:
Thanks, nice work. @xrmx safe for backport

PR #2122, opened 2020-01-25
yaml/libyaml: Fix keys ignored after sequence, support non-indented sequence entries

### LINKED ISSUE #2097: uWSGI with libyaml: static-map parsing
We are running uWSGI with libyaml, accomplished by specifying a profile as mentioned in: https://github.com/unbit/uwsgi/issues/863#issuecomment-240605936, and after much debugging found that the YAML configuration has some oddly specific requirements when it comes to listing `static-map` entries.
Specifically, the following configuration does not work:
```
uwsgi:
  static-map:
  - /static/style=/server/static/style/blue
  - /static=/server/static
  - /favicon.ico=/server/static/favicon.ico
```
appearing as if there is no static mapping, although it is technically valid YAML, but the following does work:
```
uwsgi:
  static-map:
    - /static/style=/server/static/style/blue
    - /static=/server/static
    - /favicon.ico=/server/static/favicon.ico
```
Additionally, we noticed that all the keys after `static-map` are inexplicably ignored.
Specifically, the following will properly parse all the configs:
```
uwsgi:
  buffer-size: 16384
  die-on-term: true
  enable-threads: true
  hook-master-start: 'unix_signal:2 gracefully_kill_them_all'
  http: 0.0.0.0:8080
  manage-script-name: true
  master: true
  offload-threads: 2
  processes: 1
  py-call-osafterfork: true
  threads: 4
  thunder-lock: true
  static-map:
    - /static/style=/server/static/style/blue
    - /static=/server/static
    - /favicon.ico=/server/static/favicon.ico
```
but, if the `static-map` is specified earlier in the config, everything after it seems to be ignored. For example, this will result in uWSGI signaling that thunder lock is disabled:
```
uwsgi:
  buffer-size: 16384
  die-on-term: true
  enable-threads: true
  hook-master-start: 'unix_signal:2 gracefully_kill_them_all'
  http: 0.0.0.0:8080
  manage-script-name: true
  master: true
  offload-threads: 2
  processes: 1
  py-call-osafterfork: true
  threads: 4
  static-map:
    - /static/style=/server/static/style/blue
    - /static=/server/static
    - /favicon.ico=/server/static/favicon.ico
  thunder-lock: true
```

This is specifically affecting us because we want to use valid YAML (so duplicate keys are problematic, hence the need for the libyaml version) and are generating the YAML config file through the Helm templating engine (ultimately running in a Kubernetes environment), and thus do not have direct control over the order of the keys or the indentation unless we special case the key which we are trying to avoid doing. Our presumption was that since it is libyaml, it should be able to parse any valid YAML config file, but that seems to not be the case. Are we missing something?

awelzel:
This patch re-purposes the `in_uwsgi_section` variable to track the depth
of the configuration. Previously, leaving a nested sequence would
stop the parsing process, ignoring any keys following the sequence.

Further, support non-indented sequences for which no
`YAML_BLOCK_SEQUENCE_START_TOKEN` is emitted. The entries are preceded by
the `YAML_BLOCK_ENTRY_TOKEN`, however, which can be used instead to
recognize these entries.

The following configuration is parsed sensibly after this patch when
compiled with libyaml:

    uwsgi:
      http-socket: ':4040'
      wsgi: tests.hello
      static-map:
        - /static/js=/server/static/js
        - /static/css=/server/static/css
      cache2: [
          "name=thumbnails,items=200",
          "name=videos,items=100"
      ]
      env:  # this is missing the SEQUENCE_START_TOKEN
      - ENVIRONMENT=prod
      - "DATABASE_URL=mysql://localhost"
      workers: 4
      thunder-lock: true

    # Verified with --show-config
    ./uwsgi -y ./config.yaml --show-config
    ...
    ;uWSGI instance configuration
    [uwsgi]
    yaml = ./config.yaml
    http-socket = :4040
    wsgi = tests.hello
    static-map = /static/js=/server/static/js
    static-map = /static/css=/server/static/css
    cache2 = name=thumbnails,items=200
    cache2 = name=videos,items=100
    env = ENVIRONMENT=prod
    env = DATABASE_URL=mysql://localhost
    workers = 4
    thunder-lock = true
    show-config = true
    ;end of configuration

Fixes #2097 (but *only* if libyaml is used).

### DIFF
--- a/core/yaml.c
+++ b/core/yaml.c
@@ -156,44 +156,49 @@ void uwsgi_yaml_config(char *file, char *magic_table[]) {
 			break;
 		case YAML_FLOW_SEQUENCE_START_TOKEN:
 		case YAML_BLOCK_SEQUENCE_START_TOKEN:
-			status = 3;
+			if (in_uwsgi_section)
+				in_uwsgi_section++;
+			// fallthrough
+		case YAML_FLOW_ENTRY_TOKEN:
+		case YAML_BLOCK_ENTRY_TOKEN:
+			status = 3;  // inside a sequence
 			break;
 		case YAML_BLOCK_MAPPING_START_TOKEN:
-			if (!in_uwsgi_section) {
-				if (key) {
-					if (!strcmp(section_asked, key)) {
-						in_uwsgi_section = 1;
-					}
+			if (in_uwsgi_section) {
+				in_uwsgi_section++;
+				break;
+			}
+			if (key) {
+				if (!strcmp(section_asked, key)) {
+					in_uwsgi_section = 1;
 				}
 			}
 			break;
 		case YAML_BLOCK_END_TOKEN:
-			if (in_uwsgi_section) {
-				parsing = 0;
-				break;
-			}
+		case YAML_FLOW_SEQUENCE_END_TOKEN:
+			if (in_uwsgi_section)
+				parsing = !!(--in_uwsgi_section);
+			key = NULL;
+			status = 0;
 			break;
 		case YAML_SCALAR_TOKEN:
-		case YAML_FLOW_ENTRY_TOKEN:
-		case YAML_BLOCK_ENTRY_TOKEN:
 			if (status == 1) {
 				key = (char *) token.data.scalar.value;
 			}
-			else if (status == 2) {
+			else if (status == 2 || status == 3) {
 				val = (char *) token.data.scalar.value;
 				if (key && val && in_uwsgi_section) {
 					add_exported_option(key, val, 0);
 				}
-				status = 0;
-			}
-			else if (status == 3) {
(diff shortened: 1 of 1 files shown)


awelzel:
Related to this:
> Fixes #2097 (but only if libyaml is used).

when the builtin yaml parser is used with the above sample configuration, `--show-config` produces:

```
;uWSGI instance configuration
[uwsgi]
yaml = ./config.yaml
http-socket = ':4040'
wsgi = tests.hello
show-config = true
;end of configuration
```

It might be easier to require `libyaml` in the future and remove the freestanding yaml parsing code instead of making them compatible (which they haven't been anyway).

rdeioris:
Thanks again for your patch. @xrmx safe for backport

Issue #2128, opened 2020-02-28
Chroot is called in each process after chroot of caller


ondratu:
As i write here http://lists.unbit.it/pipermail/uwsgi/2020-February/009028.html, i try to use chroot on my Debian Buster. I do some another example on another machine with the same problem. Here is another log when i use only chroot (no uid or gid).

```ini
[uwsgi]

# one day may be included
# ini = /srv/example_net/etc/uwsgi.ini
chroot = /srv/example_net

autoload = true
plugin = python3

# wsgi-file = /srv/app.py
# chdir = /srv
# module = app:app

check-static = /srv/www

master = true
threads = 5

python-autoreload = true
optimize = 2
# limit-as = 32

# uid = example_net
# gid = example_net

http = 127.0.0.1:31001
pidfile = /srv/example_net/var/run/uwsgi.pid
daemonize = /srv/example_net/var/log/uwsgi.log
#socket = /var/run/uwsgi.socket
stats = /var/run/uwsgi.stats
touch-reload = /var/run/uwsgi.reload
```

and the log after start
```
*** Starting uWSGI 2.0.18-debian (64bit) on [Fri Feb 28 08:30:03 2020] ***
compiled with version: 9.2.1 20200203 on 18 February 2020 19:38:45
os: Linux-5.3.0-3-amd64 #1 SMP Debian 5.3.15-1 (2019-12-07)
nodename: dell
machine: x86_64
clock source: unix
pcre jit disabled
detected number of CPU cores: 12
current working directory: /etc/uwsgi-emperor/vassals
writing pidfile to /srv/example_net/var/run/uwsgi.pid
detected binary path: /usr/bin/uwsgi-core
uWSGI running as root, you can use --uid/--gid/--chroot options
chroot() to /srv/example_net
*** WARNING: you are running uWSGI as root !!! (use the --uid flag) *** 
your processes number limit is 62689
your memory page size is 4096 bytes
detected max file descriptor number: 1024
building mime-types dictionary from file /etc/mime.types...567 entry found
lock engine: pthread robust mutexes
thunder lock: disabled (you can enable it with --thunder-lock)
uWSGI http bound on 127.0.0.1:31001 fd 4
uwsgi socket 0 bound to TCP address 127.0.0.1:42517 (port auto-assigned) fd 3
uWSGI running as root, you can use --uid/--gid/--chroot options
chroot() to /srv/example_net
chroot(): No such file or directory [core/utils.c line 592]
```

[...]

PR #2132, opened 2020-03-07
[RFC] chroot: Do not chroot multiple times when root

awelzel:
From 2.0.15 to 2.0.16 the uwsgi_setup() sequence was modified such that
uwsgi_as_root() may be called multiple times. If the `uid` option is
used, then on later invocation most of this function is skipped.

However, if run as root, then most everything in this function is run
again. In #2128 it was reported that the second chroot() call fails.

This patch attempts to fix this for `chroot()` only. However, this is
a bit naive - the other sandboxing calls inside uwsgi_as_root() should
probably not be run multiple times either (unshare() / capabilities),
but I'm not 100% sure about the semantics, particularly concerning
uwsgi.master_as_root, so just presenting this as an RFC.

Related, uwsgi 2.0.18 + chroot + uid is not working:

    $ sudo uwsgi --uid 1000 --chroot /
    *** Starting uWSGI 2.0.18 (64bit) on [Sat Mar  7 22:05:18 2020] ***
    ...
    detected binary path: /home/awelzel/venv/bin/uwsgi
    uWSGI running as root, you can use --uid/--gid/--chroot options
    chroot() to /
    setuid() to 1000
    ...
    cannot chroot() as non-root user

This patch happens to fix the above issue.

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -580,13 +580,15 @@ void uwsgi_as_root() {
 		}
 	}
 
-	if (uwsgi.chroot && !uwsgi.reloads) {
+	if (uwsgi.chroot && !uwsgi.is_chrooted && !uwsgi.reloads) {
 		if (!uwsgi.master_as_root)
 			uwsgi_log("chroot() to %s\n", uwsgi.chroot);
+
 		if (chroot(uwsgi.chroot)) {
 			uwsgi_error("chroot()");
 			exit(1);
 		}
+		uwsgi.is_chrooted = 1;
 #ifdef __linux__
 		if (uwsgi.logging_options.memory_report) {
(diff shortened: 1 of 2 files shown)


rdeioris:
Quick and dirty, but i do not see other ways to manage it :) @xrmx safe for backport

awelzel:
@rdeioris - hmm, ok - I would have removed the `[RFC]` part, but seems you already merged anyway... ;-)

xrmx:
Backported this and everyhing else and updated 2.0.19 changelog in uwsgi-docs, time for a new release?

rdeioris:
@xrmx waiting for customer's feedback about https://github.com/unbit/uwsgi/commit/46e1dec86f2d6d38ce30d2200c31d566b391534e . As soon as i get confirmation we can make an official release

PR #2138, opened 2020-03-31
plugins/{pypy,python}: get rid of unnecessary paste.script dependency

patrickdepinguin:
paste.script.util.logging_config is an old copy of logging.config from
Python 2.5.1 (see [1]). There are no paste-specific details in it.

By using logging.config directly, Python WSGI environments no longer need to
have pastescript installed.

[1] https://bitbucket.org/wilig/pastescript/commits/04d0db6b2f5ab360bdaa5b10511d969a24fde0ec


xrmx:
Thanks!

PR #2141, opened 2020-04-07
fix psgi plugin for GCC 10

thekad:
Following guide here https://gcc.gnu.org/gcc-10/porting_to.html

I am the fedora packager for uwsgi, this makes it possible for the psgi to build with gcc 10 again.

### DIFF
--- a/plugins/psgi/psgi_plugin.c
+++ b/plugins/psgi/psgi_plugin.c
@@ -3,11 +3,7 @@
 extern char **environ;
 extern struct uwsgi_server uwsgi;
 
-#ifdef __APPLE__
-extern struct uwsgi_perl uperl;
-#else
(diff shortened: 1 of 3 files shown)


plugins/psgi/psgi_plugin.c
```diff
@@ -3,11 +3,7 @@
 extern char **environ;
 extern struct uwsgi_server uwsgi;
 
-#ifdef __APPLE__
```
xrmx, line 6: Won't this break __APPLE__ though?

thekad: I don't think so? this is made so it matches what `__APPLE__` did

xrmx: right, let's merge this

PR #2142, opened 2020-04-09
psgi: fix impossibility of determining the end of the chunked stream

xenols:
Fix impossibility of determining the end chunked stream:
 
     while(1) {
             my $msg = uwsgi::chunked_read;
             last unless $msg;
             ......
     }

    At the end of the stream, the uwsgi_chunked_read function returns
    len 0 and a pointer to a buffer filled with old data.
    Based on the description: https://perldoc.perl.org/5.16.1/perlapi.html,
    The newSVpv function in the case when the len is zero uses the strlen.
    This makes it impossible to determine the end of the stream since the
    last chunk returns endlessly.

### DIFF
--- a/plugins/psgi/uwsgi_plmodule.c
+++ b/plugins/psgi/uwsgi_plmodule.c
@@ -885,7 +885,7 @@ XS(XS_chunked_read) {
 		XSRETURN_UNDEF;
         }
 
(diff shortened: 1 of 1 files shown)


rdeioris:
@xenols Thanks. @xrmx safe for backport

PR #2146, opened 2020-04-12
WIP cffi embedding API Python plugin

dholth:
I took the pypy plugin and updated it, in part, for the cffi embedding API. In theory this should be able to run on PyPy and CPython if the plugin config is updated to set different compiler flags for CPython.

A cffi bug prevents it from running on pypy 7.3.1. Should be fixed in the next version.

```
 ~/opt/pypy361/bin/python uwsgiconfig.py --plugin plugins/cffi nolang && \
./uwsgi --plugin cffi --http-socket 0.0.0.0:8081 --cffi-wsgi=helloworld --chdir=$HOME/opt/pypy361/bin --env=PYTHONPATH=$PWD:$PWD/plugins/cffi --cffi-init=cffi_dyn_init --master --touch-reload=$PWD/helloworld.py
```

websockets (linked websockets_chat_async.py to $PWD to avoid name collision):
```
python ./uwsgiconfig.py -p plugins/cffi nolang && ./uwsgi --plugin cffi --http-socket 0.0.0.0:8081 --cffi-wsgi=websockets_chat_async --chdir=$HOME/opt/pypy361/bin --env=PYTHONPATH=$HOME/prog/uwsgi:$HOME/prog/uwsgi/plugins/cffi --cffi-init=cffi_dyn_init --master --touch-reload=$PWD/websockets_chat_async.py --async 64
```


rdeioris:
Do you think it is possible to automatically generate cffi_plugin.py from uwsgi.h ?

dholth:
Yes to a point. The pypy plugin handles uwsgi.h more automatically. Then depending on how Pythonic you want the callback interface to be you might have to write adapters that convert things like char*, len to Python strings. But uwsgi isn't that big it's tempting to just do it manually.

I did one for sdl that uses a custom simple compiler to add python helpers. It works because that library has a very consistent API.

dholth:
I updated this so that it preprocesses and automatically wraps almost all of `uwsgi.h`.

dholth:
I was able to get the continulets async to work, but wasn't able to get the the asyncio plugin or the greenlets based async to work very well. Focusing on a simpler PR to get the basic wsgi & uwsgi API features working.

PR #2147, opened 2020-04-12
Add chunked request decoding to the CGI plugin

### LINKED ISSUE #2140: Chunked request decoding for CGI scripts
Hi all,

I'm trying to serve git-http-backend using uWSGI's cgi plugin. Arbitrarily large pulls and also small pushes (few KiB) work well so far, but as soon as I try to push around 2 MiB+, the request hangs and is finally killed by CGI timeout after 60 seconds. I tried the same push via Apache CGI and that worked even for 100 MiB.

Here is how to reproduce it easily:

uwsgi.ini:
```
[uwsgi]
plugin = cgi,router_uwsgi
master = true
http-socket = :8081
threads = 4
# Might need to be changed to /usr/libexec/git-core/git-http-backend for some distros
cgi = /usr/lib/git-core/git-http-backend
route-run = addvar:GIT_PROJECT_ROOT=.
route-run = addvar:GIT_HTTP_EXPORT_ALL=1
# git-http-backend only accepts pushes from authenticated users
route-run = addvar:REMOTE_USER=foo
route-run = uwsgi:,9,0
```

Then I ran uWSGI and did:
```
# git --version
git version 2.20.1
# git init --bare repo
Initialized empty Git repository in /home/roschi/repo/
# git clone http://localhost:8081/repo clone
Cloning into 'clone'...
warning: You appear to have cloned an empty repository.
# cd clone
# cp ../large_file .
# git add .
# git commit -m test
[master (root-commit) d581323] test
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 large_file
# git push
Enumerating objects: 3, done.
Counting objects: 100% (3/3), done.
Delta compression using up to 8 threads
Compressing objects: 100% (2/2), done.
error: RPC failed; curl 55 Send failure: Connection reset by peer
fatal: the remote end hung up unexpectedly
Writing objects: 100% (3/3), 2.43 MiB | 41.00 KiB/s, done.
Total 3 (delta 0), reused 0 (delta 0)
fatal: the remote end hung up unexpectedly
Everything up-to-date
```

Here is uWSGI's output:
```
# uwsgi --ini uwsgi.ini
[uWSGI] getting INI configuration from uwsgi.ini
*** Starting uWSGI 2.0.18-debian (64bit) on [Fri Apr  3 10:59:09 2020] ***
compiled with version: 8.2.0 on 10 February 2019 02:42:46
os: Linux-4.19.0-8-amd64 #1 SMP Debian 4.19.98-1 (2020-01-26)
nodename: xxx
machine: x86_64
clock source: unix
pcre jit disabled
detected number of CPU cores: 8
current working directory: /home/roschi
detected binary path: /usr/bin/uwsgi-core
*** dumping internal routing table ***
[rule: 0] action: addvar:GIT_PROJECT_ROOT=.
[rule: 1] action: addvar:GIT_HTTP_EXPORT_ALL=1
[rule: 2] action: addvar:REMOTE_USER=foo
[rule: 3] action: uwsgi:,9,0
*** end of the internal routing table ***
your processes number limit is 62445
your memory page size is 4096 bytes
detected max file descriptor number: 1024
lock engine: pthread robust mutexes
uwsgi socket 0 bound to TCP address :8081 fd 3
your server socket listen backlog is limited to 100 connections
your mercy for graceful operations on workers is 60 seconds
mapped 208576 bytes (203 KB) for 4 cores
*** Operational MODE: threaded ***
initialized CGI path: /usr/lib/git-core/git-http-backend
*** no app loaded. going in full dynamic mode ***
spawned uWSGI master process (pid: 19672)
spawned uWSGI worker 1 (pid: 19673, cores: 4)
[pid: 19673|app: -1|req: -1/1] 127.0.0.1 () {36 vars in 482 bytes} [Fri Apr  3 10:59:27 2020] GET /repo/info/refs?service=git-upload-pack => generated 38 bytes in 2 msecs (HTTP/1.1 200) 4 headers in 189 bytes (0 switches on core 1)
[pid: 19673|app: -1|req: -1/2] 127.0.0.1 () {36 vars in 484 bytes} [Fri Apr  3 10:59:48 2020] GET /repo/info/refs?service=git-receive-pack => generated 180 bytes in 6 msecs (HTTP/1.1 200) 4 headers in 190 bytes (0 switches on core 0)
[pid: 19673|app: -1|req: -1/3] 127.0.0.1 () {34 vars in 463 bytes} [Fri Apr  3 10:59:48 2020] POST /repo/git-receive-pack => generated 0 bytes in 2 msecs (HTTP/1.1 200) 4 headers in 183 bytes (0 switches on core 2)
CGI timeout !!!
fatal: the remote end hung up unexpectedly
[pid: 19673|app: -1|req: -1/4] 127.0.0.1 () {36 vars in 514 bytes} [Fri Apr  3 10:59:48 2020] POST /repo/git-receive-pack => generated 0 bytes in 60047 msecs (HTTP/1.1 200) 4 headers in 183 bytes (0 switches on core 1)
```

Things I tried so far, none of which seemed to have an effect:
* `--cgi-buffer-size=100000000`
* running without master
* `--processes=4`` instead of `--threads=4`
* `--ugreen --async=100` instead of `--threads=4`
* `--offload-threads=1`
* uWSGI on alpine via Docker instead of Debian host

Does anyone have an idea what's happening here?

Thanks and all the best
Robert

efficiosoft:
When a request using 'Transfer-Encoding: chunked' is routed to the CGI
backend, the request body is now read and sent to the script chunk by
chunk, allowing for requests with dynamic body length to pass. The CGI
process's stdin is always closed after the last chunk has been written,
regardless of --cgi-close-stdin-on-eof.

A caveat is that it's only started to read the CGI process's output
after the full request body has been consumed, meaning a CGI script may
not start writing a response before it has read the entire body to avoid
getting blocked during the write operation to stdout.

Closes #2140

### DIFF
--- a/plugins/cgi/cgi_plugin.c
+++ b/plugins/cgi/cgi_plugin.c
@@ -704,23 +704,50 @@ static int uwsgi_cgi_run(struct wsgi_request *wsgi_req, char *docroot, size_t do
 		uwsgi_socket_nb(cgi_pipe[0]);
 		uwsgi_socket_nb(post_pipe[1]);
 
-		// ok start sending post data...
-		size_t remains = wsgi_req->post_cl;
-		while(remains > 0) {
-                	ssize_t rlen = 0;
-                	char *buf = uwsgi_request_body_read(wsgi_req, 8192, &rlen);
-                	if (!buf) {
-				goto clear2;
-                	}
-                	if (buf == uwsgi.empty) break;
-                	// write data to the node
-                	if (uwsgi_write_true_nb(post_pipe[1], buf, rlen, uc.timeout)) {
-				goto clear2;
-                	}
-                	remains -= rlen;
-        	}
+		// Start sending request body
+		if (wsgi_req->body_is_chunked) {
+			// Write through to process chunk by chunk
+			while (1) {
+				struct uwsgi_buffer *ubuf = uwsgi_chunked_read_smart(wsgi_req, 8192, uwsgi.socket_timeout);
+				if (!ubuf) {
+					uwsgi_log("error reading chunk from CGI request !!!\n");
+					kill_on_error
+						goto clear2;
+				}
+				if (!ubuf->pos) {
+					// Last chunk received, go and close process's stdin
+					uwsgi_buffer_destroy(ubuf);
+					break;
+				}
+				// Write chunk to process's stdin
+				int err = uwsgi_write_true_nb(post_pipe[1], ubuf->buf, ubuf->pos, uc.timeout);
+				uwsgi_buffer_destroy(ubuf);
+				if (err) {
+					uwsgi_log("error writing chunk to CGI process !!!\n");
+					kill_on_error
+						goto clear2;
+				}
+			}
+		} else {
+			// Normal request with content length set
+			size_t remains = wsgi_req->post_cl;
+			while(remains > 0) {
+				ssize_t rlen = 0;
+				char *buf = uwsgi_request_body_read(wsgi_req, 8192, &rlen);
+				if (!buf) {
+					goto clear2;
+				}
+				if (buf == uwsgi.empty) break;
+				// write data to the node
+				if (uwsgi_write_true_nb(post_pipe[1], buf, rlen, uc.timeout)) {
+					goto clear2;
+				}
+				remains -= rlen;
+			}
+		}
 
-		if (uc.close_stdin_on_eof) {
+		// For chunked requests, close stdin to tell the script body has ended
+		if (uc.close_stdin_on_eof || wsgi_req->body_is_chunked) {
 			close(post_pipe[1]);
 			stdin_closed = 1;
 		}


rdeioris:
@xrmx this looks good to me, i think it can be backported too. Any objection ?

xrmx:
Indentation for the non-chunked branch is not wrong though. Fixed that is fine for me

efficiosoft:
> Indentation for the non-chunked branch is now wrong though. With that fixed is fine for me

I also noticed this, but didn't want to touch more lines than necessary to keep the diff small. Should I simply re-indent the whole file with tabs?

xrmx:
Reindent only the block you are putting in the else branch.

efficiosoft:
> Reindent only the block you are putting in the else branch.

I didn't touch that block originally, it was indented that way before. Anyway, re-indented it, though now the diff has grown a bit.

Anything else I could do to help get this released?

efficiosoft:
Ah, I could add something about it to docs as well.

Basically it should be said that requests with chunked transfer don't provide the Content-Length header and stdin is always closed to notify about end of data for such requests, regardless of the `--cgi-close-stdin-on-eof` setting.

Would that be appropriate?

efficiosoft:
BTW, why don't we just always close stdin on EOF? It would be more consistent that way and save us a setting.

The [CGI specification](https://tools.ietf.org/html/rfc3875#section-4.2) says:

> The server MAY signal an end-of-file condition after CONTENT_LENGTH bytes have been read or it MAY supply extension data.  Therefore, the script MUST NOT attempt to read more
   than CONTENT_LENGTH bytes, even if more data is available.  However, it is not obliged to read any of the data.

Given that uWSGI doesn't provide extension data (does it?), just signalling EOF would be appropriate. Or maybe there's a situation I haven't thought of yet in which it would make sense to keep the file open?

PR #2149, opened 2020-04-14
gevent plugin: adapt to removal of deprecated interfaces in gevent 1.5.0

### LINKED ISSUE #2144: uWSGI workers die under gevent 1.5.0
I'm experiencing strange issues with the newly released `gevent==1.5.0`.

Application workers get killed with
```
DAMN ! worker 1 (pid: 26925) died :( trying respawn ...
```
...after a seemingly successful start.

Using `gevent.monkey.patch_all()` helps to alleviate the issue, although previously unseen weird errors are displayed:
```
TypeError: 'module' object is not callable
TypeError: 'module' object is not callable
TypeError: 'module' object is not callable
```

Everything works as expected with `gevent==1.4.0`.

To reproduce, place `case.sh` and `test.py` under `/tmp/test/` in an Ubuntu 16.04 or 18.04 Docker container, and run `case.sh`:
[test.zip](https://github.com/unbit/uwsgi/files/4464288/test.zip)

I'm also getting the same behaviour when building uWSGI Python and Gevent support as plugins, i.e. along the lines of
```
PYTHON=python3.5 make PROFILE=nolang
PYTHON=python3.5 ./uwsgi --build-plugin plugins/python
PYTHON=python3.5 ./uwsgi --build-plugin plugins/gevent
```

vytas7:
Closes #2144 

Many thanks for @jamadden for finding the root cause of issues.

Note that I'm still having issues with uWSGI shutdown, looking into that.

And as noted by @jamadden , `gevent.version_info` will also be deprecated soon, I'll check if I could squash that too while at it.

### DIFF
--- a/plugins/gevent/gevent.c
+++ b/plugins/gevent/gevent.c
@@ -452,7 +452,12 @@ static void gevent_loop() {
 	if (!ugevent.spawn) uwsgi_pyexit;
 
 	ugevent.signal = PyDict_GetItemString(gevent_dict, "signal_handler");
-	if (!ugevent.signal) uwsgi_pyexit;
+	if (!ugevent.signal) {
+		// gevent.signal_handler appears in gevent 1.3.
+		// On older gevent, fall back to the deprecated gevent.signal.
+		ugevent.signal = PyDict_GetItemString(gevent_dict, "signal");
+		if (!ugevent.signal) uwsgi_pyexit;
+	}
 
 	ugevent.greenlet_switch = PyDict_GetItemString(gevent_dict, "sleep");
 	if (!ugevent.greenlet_switch) uwsgi_pyexit;
--- a/plugins/gevent/gevent.c
+++ b/plugins/gevent/gevent.c
@@ -435,9 +435,6 @@ static void gevent_loop() {
 	PyObject *gevent_dict = get_uwsgi_pydict("gevent");
 	if (!gevent_dict) uwsgi_pyexit;
 
-	PyObject *gevent_signal_dict = get_uwsgi_pydict("gevent.signal");
-	if (!gevent_signal_dict) uwsgi_pyexit;
-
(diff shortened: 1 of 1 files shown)


xrmx:
Does this break compatibility with older gevent? If so which version?

jamadden:
`gevent.signal.signal` has existed since gevent 1.1, circa 2015.

vytas7:
@xrmx I've now changed the proposed solution to use the simplified `gevent.signal_handler` instead of `gevent.signal.signal` as it seems it is what uWSGI is after... As well as added a compatibility shim to fall back to `gevent.signal` on `gevent` 1.1 and 1.2.

Judging from code syntax, `gevent` 1.0 does not seem to support Python 3, so I haven't looked too deep into it.

@jamadden is it OK to rely on `gevent.signal_handler`, or will it also be deprecated in the future?

jamadden:
`gevent.signal_handler` is documented and has no announced deprecation.

xrmx:
@vytas7 merged thanks!

vytas7:
@xrmx That would be appreciated in 2.0.x, but I don't know how uWSGI release schedule is looking like otherwise.
In a pinch, one can just pin `gevent` to 1.4.x, even if my fix doesn't make it.

seansfkelley:
I just spent a while chasing down this compatibility issue (those stack traces from C code are... not helpful) and am happy to find that it's just been fixed. Thanks @vytas7!

@xrmx I would love to have this in 2.0.x if possible, yes!

xrmx:
@seansfkelley It's there since a couple of days already :sweat_smile:

seansfkelley:
Sorry, I don't follow -- don't I need to wait for a 2.0.19 release to pick up these changes? Or are you suggesting I build it from source? I don't see any such release on pypi.

Rukeith:
Me too. wait for the release

PR #2150, opened 2020-04-14
Ensure '-lrt' is included on the link line after linking to python li…

ddowling:
…brary. Python depends on clock_gettime() and other functions in librt. Nothing else in uwsgi has a dependency on librt so if this library is only specified early on the link line then the library is dropped by the linker.

### DIFF
--- a/plugins/python/uwsgiplugin.py
+++ b/plugins/python/uwsgiplugin.py
@@ -69,6 +69,8 @@ if 'UWSGI_PYTHON_NOLIB' not in os.environ:
         # hack for messy linkers/compilers
         if '-lutil' in LIBS:
             LIBS.append('-lutil')
+        if '-lrt' in LIBS:
+            LIBS.append('-lrt')
     else:
         try:
             libdir = sysconfig.get_config_var('LIBDIR')


ddowling:
I cannot install uwsgi 2.0.18 using anaconda and pip.
Steps to repeat:
```
conda create -n utest
conda activate utest
conda install pip
conda install gcc_linux-64
pip install uwsgi

```
Build fails with a large number of undefined references to clock_gettime() and others

The Anaconda environment builds Python without the --enabled-shared flag so that the python executable is statically linked. This means the uwsgi build scripts will also try to statically link against libpython3.8.a and this link will fail unless -lrt is included on the link line after the python library.

xrmx:
When using conda should'nt you use something like this one https://anaconda.org/conda-forge/uwsgi?

ddowling:
Thinking about it I probably should use the conda-forge uwsgi but when I first tried to install it conda reported a mess of conflicts. I then went down the path of trying to use pip to install uwsgi. I perhaps should try to work out what is going on with conda.

MichaelKorn:
Same issue. We use conda for the dev environment of the developers. In the env yml we use the requirement files which are used for pip by the CI process. The way there is just one set of python dependencies definition.

MichaelKorn:
@xrmx 
I can confirm, that patch solves my building issue. It would be cleaner to put `libpython3.9.a` before the existing `-lrt`, but it works. As the if block is explicitly for static linking, I can not understand why this PR was not merged.

MichaelKorn:
@xrmx could you please merge this?

PR #2152, opened 2020-04-20
plugins/cgi: improve performance for santitizing file descriptors

### LINKED ISSUE #2053: CGI sanitize file descriptors is too slow in some circumstances
[This](https://github.com/unbit/uwsgi/blob/master/plugins/cgi/cgi_plugin.c#L782) block of code takes 1.5 seconds to execute under docker on my PROD machines:
```
for(i=3;i<(int)uwsgi.max_fd;i++) {
        close(i);
}
```

I don't have access rights necessary to debug this within the docker or debug the docker itself. I tried to reproduce this in other places where I can become root but it was not that bad.

When I replace that loop with reading available descriptors from `/proc/<pid>/fd` it works perfectly still successfully closing all the unwanted descriptors.

I suggest to replace that loop with reading from `/proc/<pid>/fd` on machines where `procfs` is available.

ncopa:
Only close file descriptors that are known to be open when /proc/self/fd
or /dev/fd exists. Fall back and loop over all when those are missing.

This improves performance significantly when number of max fds is high.

fixes #2053

### DIFF
--- a/plugins/cgi/cgi_plugin.c
+++ b/plugins/cgi/cgi_plugin.c
@@ -793,8 +793,21 @@ clear2:
 	close(cgi_pipe[1]);
 
 	// close all the fd > 2
-	for(i=3;i<(int)uwsgi.max_fd;i++) {
-		close(i);
+	DIR *dirp = opendir("/proc/self/fd");
+	if (dirp == NULL)
+		dirp = opendir("/dev/fd");
+	if (dirp != NULL) {
+		struct dirent *dent;
+		while ((dent = readdir(dirp)) != NULL) {
+			int fd = atoi(dent->d_name);
+			if ((fd > 2) && fd != dirfd(dirp))
+				close(fd);
+		}
+		closedir(dirp);
+	} else {
+		for(i=3;i<(int)uwsgi.max_fd;i++) {
+			close(i);
+		}
 	}
 
 	// fill cgi env


ncopa:
Difference is:
```
Requests per second:    9.58 [#/sec] (mean)
```
vs
```
Requests per second:    48.31 [#/sec] (mean)
```

Details:
Without fd patch:

```
ab -n 100 -c 10 https://cgit.alpinelinux.org/aports
This is ApacheBench, Version 2.3 <$Revision: 1874286 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking cgit.alpinelinux.org (be patient).....done


Server Software:        nginx/1.17.8
Server Hostname:        cgit.alpinelinux.org
Server Port:            443
SSL/TLS Protocol:       TLSv1.2,ECDHE-RSA-AES256-GCM-SHA384,2048,256
Server Temp Key:        X25519 253 bits
TLS Server Name:        cgit.alpinelinux.org

Document Path:          /aports
Document Length:        15138 bytes

Concurrency Level:      10
Time taken for tests:   10.434 seconds
Complete requests:      100
Failed requests:        0
Total transferred:      1534200 bytes
HTML transferred:       1513800 bytes
Requests per second:    9.58 [#/sec] (mean)
Time per request:       1043.399 [ms] (mean)
Time per request:       104.340 [ms] (mean, across all concurrent requests)
Transfer rate:          143.59 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:       74   90  15.3     88     221
Processing:   399  883 130.6    908    1075
Waiting:      369  825 124.4    857    1030
Total:        486  972 133.3    996    1204

Percentage of the requests served within a certain time (ms)
  50%    996
  66%   1051
  75%   1070
  80%   1087
  90%   1122
  95%   1145
  98%   1169
  99%   1204
 100%   1204 (longest request)
```

With fd patch:
```
ab -n 100 -c 10 https://cgit.alpinelinux.org/aports
This is ApacheBench, Version 2.3 <$Revision: 1874286 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking cgit.alpinelinux.org (be patient).....done


Server Software:        nginx/1.17.8
Server Hostname:        cgit.alpinelinux.org
Server Port:            443
SSL/TLS Protocol:       TLSv1.2,ECDHE-RSA-AES256-GCM-SHA384,2048,256
Server Temp Key:        X25519 253 bits
TLS Server Name:        cgit.alpinelinux.org

Document Path:          /aports
Document Length:        15138 bytes

Concurrency Level:      10
Time taken for tests:   2.070 seconds
Complete requests:      100
Failed requests:        0
Total transferred:      1534200 bytes
HTML transferred:       1513800 bytes
Requests per second:    48.31 [#/sec] (mean)
Time per request:       206.992 [ms] (mean)
Time per request:       20.699 [ms] (mean, across all concurrent requests)
Transfer rate:          723.82 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:       71  114  55.5     97     453
Processing:    55   78  17.4     73     149
Waiting:       26   39   9.0     37      69
Total:        135  192  57.1    176     526

Percentage of the requests served within a certain time (ms)
  50%    176
  66%    187
  75%    195
  80%    208
  90%    251
  95%    336
  98%    348
  99%    526
 100%    526 (longest request)
```

rdeioris:
Thanks @ncopa, /dev/fd should grant a good amount of *BSD support. @xrmx i do not know if we need to add a bunch of #ifdef as filesystem access/stat is pretty cheap.

ncopa:
> Thanks @ncopa, /dev/fd should grant a good amount of *BSD support. 

It should also cover MacOS.

> i do not know if we need to add a bunch of #ifdef as filesystem access/stat is pretty cheap.

I don't think `#ifdef`'ing is worth it here. There are BSDs with linux compat and it appears that linux also has `/dev/fd`.

We could do something like:
```c
	DIR *dirp;
#if defined(__linux__)
	dirp = opendir("/proc/self/fd");
￼	if (dirp == NULL)
#endif
		dirp = opendir("/dev/fd");
```
or
```c
	DIR *dirp = NULL;
#if defined(__linux__)
	dirp = opendir("/proc/self/fd");
#endif
￼	if (dirp == NULL)
		dirp = opendir("/dev/fd"); /* BSDs and MacOS */
```
but I would prefer not add lots of different `#ifdef`s.

xrmx:
I have /dev/fd on debian too, so agree adding #ifdefery is not worthwhile.

xrmx:
Let's merge this, thanks!

ncopa:
I've recently learned that it is not allowed to call async unsafe functions after fork, which may result in deadlocks. opendir() does malloc so it is not really allowed to call it after fork. We can either do the opendir before fork or we can use poll to speed up the closing of fds.

PR #2159, opened 2020-05-09
improve plugin errors with execfile

dholth:
Shows a filename and line number when an execfile'd python plugin has an error. Tested in Python 3.


xrmx:
Nice, thanks!

PR #2167, opened 2020-05-18
Fix possible deadlock in install

jacobtolar:
When  using`stderr=subprocess.PIPE, stdout=subprocess.PIPE`, you can (will) deadlock on install of uwsgi from source if one of the output pipes fills up before the process exits. (In fact, it does just that on my system: hangs forever because clang is blocking trying to write to a full stderr pipe).


xrmx:
Thanks!

PR #2168, opened 2020-05-19
Fixed logging issue - size of package shown as negative value

pawelmarkowski:
* Fixed logging issue - size of package shown as negative value when it was greater than INT_MAX

* uwsgi_size2str(size_t) method refactored


rdeioris:
thanks! @xrmx safe for backport

PR #2170, opened 2020-05-23
cffi+python plugin

dholth:
This is an update of the pypy plugin using the cffi embedding interface. Includes the unused declarations PR.

Cleaned up from the older PR. In this PR will try to port the pypy plugin without the more difficult asyncio bits.

### DIFF
--- a/plugins/cffi/cffitest.sh
+++ b/plugins/cffi/cffitest.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+# Run from base uwsgi directory
+set -e
+# export C_INCLUDE_PATH=/usr/local/opt/openssl\@1.1/include
+python ./uwsgiconfig.py -p plugins/cffi nolang
+./uwsgi --plugin cffi -T \
+    --http-socket 0.0.0.0:8080 \
+    --env=PYTHONPATH=$HOME/prog/uwsgi:$HOME/prog/uwsgi/plugins/cffi \
+    --cffi-init=cffi_dyn_init \
+    --cffi-wsgi=helloworld \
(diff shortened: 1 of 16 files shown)


xrmx:
@dholth coould you please rebase on master to drop the merge commit please?

plugins/cffi/cffitest.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
xrmx, line 1: You can add the plugin in .travis.yml to get compiled tested and this check in tests/travis.sh to have a smoke test on each pr

dholth:
I'm using a pypy3 nightly and pypy3 7.1.1 to test. 7.3.1 has a bug with the cffi embedding interface.

It compiles in CPython but it doesn't find the encoding module. Maybe a different startup environment variable, path, or working directory could fix the problem.

dholth:
So a neat thing happened. The version I have locally works. The cffi version of the asyncio event loop works like the C version, including the bugs. (Most of) the async features work with the new Python plugin.

One bug I'm having is that the `wsgi_req` used by websockets has to time out before it can be used again. If you reload the websockets chat page, disconnecting the websockets connection, the new connection will stall for 30 seconds before the new page loads.

I think I'll have to copy some of the async features, like handling `wsgi_req->async_force_again`, that weren't present in the outdated pypy plugin.

plugins/cffi/cffitest.sh
dholth: It looks like Travis is running on xenial which has a much too old pypy for this

PR #2173, opened 2020-06-01
Correctly handle SERVER_PORT for IPv6 sockets

cbay:
When listening in IPv6, `wsgi_req->socket->name` may contain multiple `:` characters: the ones from the IPv6 address and the port separator. The current code looks for the first `:` in the string and takes the second part as the port, which is wrong. With `http-socket = [fd00::1:2345]:8100`, it would return `:1:2345]:8100`. 

We need to find the last `:` in the string instead, i.e. the first one backwards.

### DIFF
--- a/proto/http.c
+++ b/proto/http.c
@@ -311,7 +311,7 @@ static int http_parse(struct wsgi_request *wsgi_req, char *watermark) {
                 }
 	}
 	else {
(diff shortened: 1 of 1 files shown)


xrmx:
Thanks for the patch. Does that fix an issue for you or do you have find it by review? That pattern is pretty widespread in the codebase.

cbay:
It does fix the issue for me.

PR #2174, opened 2020-06-03
fix execfile error in py3

benwu95:
If it's not in `__main__`, then `__builtins__` is `__builtin__.__dict__`

### DIFF
--- a/uwsgiconfig.py
+++ b/uwsgiconfig.py
@@ -1396,7 +1396,8 @@ try:
 except NameError:
     def execfile(path, up):
         with open(path) as py:
-            code = __builtins__.compile(py.read(), path, 'exec')
+            import builtins
+            code = builtins.compile(py.read(), path, 'exec')
         exec(code, up)
 
 


xrmx:
Thanks for the patch, For which python versione this does not work for you? Works fine on python 3.8 here

benwu95:
python 3.7 in docker container

Segmentational:
@xrmx - I believe this may impact all python3 builds; please see this last [commit on May 9th](https://github.com/benwu95/uwsgi/commit/8ed288bfbc1313f4b6af9db1384e879464556f9d#diff-ac3193b4c636ca586b5af0fab95b4c99)

Additionally, the issue is resolved removing 

```python
try:
    execfile
except NameError:
    def execfile(path, up):
        with open(path) as py:
            import builtins
            code = builtins.compile(py.read(), path, 'exec')
        exec(code, up)
```

and changing `get_plugin_up` to:

```python
def get_plugin_up(path):
    up = {}
    if os.path.isfile(path):
        bname = os.path.basename(path)
        # override path
        path = os.path.dirname(path)
        up['GCC_LIST'] = [bname]
        up['NAME'] = bname.split('.')[0]
        if not path:
            path = '.'
    elif os.path.isdir(path):
        uwsgi_plugin = open('%s/uwsgiplugin.py' % path, "rb").read()
        exec(uwsgi_plugin, up)
```

I apologize for writing these changes inline; I'm in a bit of a hurry at the moment (:

PR #2176, opened 2020-06-05
cffi + asgi

dholth:
This adds asgi with an asyncio event loop to uWSGI. ASGI is a new way to do http + websockets from asyncio Python. The websockets chat example works. However it continues to be very difficult to get the uWSGI internals right, to figure out the right control flow when perhaps the old Python plugin was written against an older version of uWSGI's internals and the more maintained event loops seem to make greater use of the `wsgi_req->async_*` values. For example the websockets seem to close okay but are not logged.

I thought the tornado event loop was the easiest to read but I'm not sure if its approach would work with asyncio.

The coolest thing about this IMO is that it can detect WSGI and ASGI applications and just host them, both natively, wherever you decide they should be mounted. Perfectly fine if you have a normal "quick requests" WSGI app.

### DIFF
--- a/plugins/cffi/cffi_asyncio.py
+++ b/plugins/cffi/cffi_asyncio.py
@@ -1,0 +1,333 @@
+"""
+Direct port of plugins/asyncio.c
+"""
+
+import sys
+import asyncio
+
+from _uwsgi import ffi, lib
+
+uwsgi = lib.uwsgi
+
+
+def print(*values):
+    lib.uwsgi_log(" ".join(str(v) for v in values).encode("utf-8") + b"\n")
+
+
+def get_loop():
+    return asyncio.get_event_loop()
+
+
+# testing
+def request_id():
+    return (
+        lib.uwsgi.workers[lib.uwsgi.mywid].cores[lib.uwsgi.wsgi_req.async_id].requests
+    )
+
+
+# keep alive
+_ASYNCIO = ffi.new("char[]", "asyncio".encode("utf-8"))
+
+timeout_handles = {}  # or array with # of cores?
+
+ob_timeout_counter = 0
+
+
+def store_ob_timeout(wsgi_req, ob_timeout):
+    global ob_timeout_counter
+    # TODO check if wsgi_req.async_id is consistent enough for us
+    ob_timeout_counter += 1
+    wsgi_req.async_timeout = ffi.cast("struct uwsgi_rb_timer *", ob_timeout_counter)
+    timeout_handles[ob_timeout_counter] = ob_timeout
+
+
+def get_ob_timeout(wsgi_req):
+    ob_timeout_ = ffi.cast("ssize_t", wsgi_req.async_timeout)
+    wsgi_req.async_timeout = ffi.NULL
+    return timeout_handles.pop(ob_timeout_)
+
+
+def print_exc():
+    import traceback
+
+    traceback.print_exc()
+
+
+def setup_asyncio(threads):
+    setattr(uwsgi, "async", threads)  # keyword
+    if uwsgi.socket_timeout < 30:
+        uwsgi.socket_timeout = 30
+
+    uwsgi.loop = _ASYNCIO
+
+
+def free_req_queue(wsgi_req):
+    """
+    A #define that uses wsgi_req from context.
+    #define free_req_queue uwsgi.async_queue_unused_ptr++; uwsgi.async_queue_unused[uwsgi.async_queue_unused_ptr] = wsgi_req
+    """
+    lib.uwsgi.async_queue_unused_ptr = lib.uwsgi.async_queue_unused_ptr + 1
+    lib.uwsgi.async_queue_unused[lib.uwsgi.async_queue_unused_ptr] = wsgi_req
+    print("free_req_queue", lib.uwsgi.async_queue_unused_ptr, wsgi_req)
+
+
+@ffi.def_extern()
+def uwsgi_asyncio_wait_read_hook(fd, timeout):
+    print("enter read hook")
+    wsgi_req = uwsgi.current_wsgi_req()
+    loop = get_loop()
+    loop.add_reader(fd, uwsgi_asyncio_hook_fd, wsgi_req)
+    try:
+        ob_timeout = loop.call_later(timeout, hook_timeout, wsgi_req)
+        try:
+
+            # to loop
+            if uwsgi.schedule_to_main:
+                print("wait_read_hook", fd)
+                uwsgi.schedule_to_main(wsgi_req)
+            else:
+                print("no read hook", uwsgi.schedule_to_mani)
+            # from loop
+        finally:
+            ob_timeout.cancel()
+
+        if wsgi_req.async_timed_out:
+            return 0
+
+        return 1
+
+    except:
+        print_exc()
+        return -1
+
+    finally:
+        # returns False if not added
+        loop.remove_reader(fd)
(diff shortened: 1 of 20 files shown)


plugins/cffi/cffi_asyncio.py
```diff
@@ -0,0 +1,854 @@
...
+    """
+    Could app be ASGI?
+    """
+    return inspect.iscoroutinefunction(app) or inspect.iscoroutinefunction(app.__call__)
+
+
+ASGI_CALLABLE = ffi.cast("void *", 2)
+
+
+@ffi.def_extern()
```
dholth, line 685: Here we redefine the request callback compared to the other part of the plugin `cffi_init.py`

rdeioris:
Great work. Yes, the python plugin is the one using the older (mainly non-async) api. PSGI and Lua ones are generally mode "modern".

dholth:
@rdeioris is it safe to call websocket send and websocket receive in the main greenthread (not the per-request thread)? Is there a way to force those functions to attempt to call schedule_to_req / schedule_to_main to test that?

Gronix:
Hey, any progress on this PR? It’s very important to us.

dholth:
I would be happy to collaborate with you on this. Please test it and let me know if it works for you. Do the command line arguments make sense? Is it reliable? Is the pypy-event-loop compatible with other plugins like static files? How about websockets versus ordinary connections? Does it find virtualenvs correctly, in CPython and in PyPy? Is it OK that this needs an unreleased nightly of pypy? Is it OK that it is Python 3 only?

I have a few versions of this pull request. ASGI+asyncio, trio, and just regular wsgi but with the cffi backend. The ASGI versions can also run regular wsgi in the same process. I think the trio version is the most interesting.

They run in CPython (more slowly than the classic python plugin) and in pypy (more quickly).

Sometimes the asyncio version (not the trio version) seems to get into a loop, where it always thinks the fd is readable around https://github.com/dholth/uwsgi/blob/cffi-trio/plugins/cffi/cffi_asyncio.py#L427 . The way that cancelled timeouts work could be improved?

If you care about it, run your application and let me know if it works.

Gronix:
Hello! Sorry it took so long to get back to you.

I have run the application with uWSGI from your PR. HTTP-Connection Upgrade occurs (event: `websocket.accept`), but there are some problems when the Python-SocketIO server tries to send data over a WebSocket connection - turns out the connection is already closed. Despite this, Http requests from the Socket.io client seem to be served OK.

More details:

`cffi_asyncio.py` : `handle_asgi_request` function
```python
...
async def send(event):
    nonlocal closed
    print("ws send", event)   # 1) Output: "ws send {'type': 'websocket.send', 'bytes': None, ..."
    if closed:
        print("ignore send on closed ws")
        ws_event.set()

...

async def receiver():
    ...
    while True:
        try:
            print("rx, closed=", closed)  # 2) Output: "rx, closed= False"
            msg = websocket_recv_nb(wsgi_req)  # 3) but here we get an Exception
        except IOError as e:
            closed = True
             ...
```
in console:
```
ws send {'type': 'websocket.send', 'bytes': None, 'text': '0{"sid":"6b1ba0b902544cb9bf413cca88f8698c","upgrades":[],"pingTimeout":60000,"pingInterval":25000}'}
ignore send on closed ws
```
Tb:
```python
  File "/usr/src/uwsgi/uwsgi-cffi-asgi/plugins/cffi/cffi_asyncio.py", line 592, in receiver
    msg = websocket_recv_nb(wsgi_req)
  File "/usr/src/uwsgi/uwsgi-cffi-asgi/plugins/cffi/cffi_asyncio.py", line 488, in websocket_recv_nb
    raise IOError("unable to receive websocket message")
OSError: unable to receive websocket message
```


Also faced with the following:

1. no way to build `cffi` plugin as embedded (with build config) - will raise error

2. `python` binary is required to install uWSGI. (before we use only system-wide python version 3.7 inside docker without virtualenvs) - I had to create symlink

    Traceback for points 1 and 2:
    ```python
        ...
        python constants.py         <---- "python" call
          File "constants.py", line 14, in <module>
            uwsgi_cflags = subprocess.check_output(["../../uwsgi", "--cflags"]).decode("utf-8")
          File "/usr/lib/python3.7/subprocess.py", line 411, in check_output
            **kwargs).stdout
          File "/usr/lib/python3.7/subprocess.py", line 488, in run
            with Popen(*popenargs, **kwargs) as process:
          File "/usr/lib/python3.7/subprocess.py", line 800, in __init__
            restore_signals, start_new_session)
          File "/usr/lib/python3.7/subprocess.py", line 1551, in _execute_child
            raise child_exception_type(errno_num, err_msg, err_filename)
        FileNotFoundError: [Errno 2] No such file or directory: '../../uwsgi': '../../uwsgi'
        make: *** [Makefile:5: _constants.h] Error 1
        ...
    ```
3. It is not possible to set an executable python file of the application through regular commands `file`/`wsgi`/... - only mount points, as in your examples. Otherwise raise this (I’ve understood after a while =/ ) :

>        *** no app loaded. GAME OVER *** 
    
4. `PYTHONPATH` with a path to `plugins/cffi` is required to run worker - that was unexpected.
5. setting the value of coroutines count (`--async` option) less than *32* causes an error _SegmentationFault_ on application startup!
    

I created a repo with a simple example of our app where you can reproduce the errors mentioned above. Notice: application code runs under Daphne without any issues.

Link to the repo: https://github.com/Gronix/uwsgi-asgi-test

dholth:
It looks like you are interested in running with CPython.

The plugin uses Python to do code generation based on uwsgi options, it wraps the part of the uwsgi API that has been compiled. There might be some flexibility to use the `.h` file instead of calling the `uwsgi` binary.

There is a `--cffi-wsgi` but the classic Python plugin owns the `--file` option. https://github.com/unbit/uwsgi/pull/2176/files#diff-0156a53c40615b2420b28af9df9687adR87

There's a hardcoded 32. The interaction between the event loop, the coroutine engine, and the plugin is very difficult to understand. We replace all three of those things with our new cffi Python plugin. https://github.com/unbit/uwsgi/pull/2176/files#diff-b56242d0c88cfe2ac6b9beebf845d402

The new asyncio plugin https://github.com/unbit/uwsgi/blob/96da7519cc0e675700c0d970eb9c815638170e98/plugins/cffi/cffi_asyncio.py

I'll see if I can push up the improved trio version of this plugin. I have some changes that embed the necessary Python code into the plugin and allow for improved tracebacks. I've been copying improvements back and forth between the trio and asyncio versions of this, and trying to simplify the core synchronous-wsgi version.

dholth:
I've pushed some changes to https://github.com/dholth/uwsgi which focuses on trio but also has asyncio. I'll let you know if I have the chance to run your app.

Gronix:
OK, thank you. Is there anything else I can help you with?

dholth:
@Gronix check out the documentation https://github.com/dholth/uwsgi-docs/blob/cffi/CFFI.rst

dholth:
There's a new pypy release. This plugin should work with the newest pypy without having to patch anything.

dholth:
I have put together a version that is dramatically easier to try. It has a Dockerfile and a vscode configuration to automatically build and launch that container. But it doesn't exercise the asyncio or trio based ASGI code yet. https://github.com/dholth/uwsgi-pypy-demo

dholth:
I'm closing this pull request in favor of a different branch. The other branch includes a trio variant.

PR #2182, opened 2020-06-13
use strrchr to get port

dholth:
Noticed an ipv6 --http-socket=[::]:80 had a bad port in the SERVER_PORT variable


xrmx:
Patch looks good. It is probably time to create an helper and switch all the user to that.

PR #2190, opened 2020-06-22
Request Transfer-Encoding

nbrachet:
2 bug fixes to Transfer-Encoding requests:
1. In raw mode (with no Content-Length), don't overwrite `hr->remains` and lose data
1. Make sure to grab the GIL before calling any Python methods


rdeioris:
@xrmx safe for backpo, @nbrachet  nice catch!, thanks

xrmx:
Picked the http one, the missing gil acquire does not apply to 2.0 since it is missing the chunked body stuff

PR #2193, opened 2020-06-26
Add path to header file

thekad:
The patch in #1772 never made it to the 2.0 branch.


thekad:
I suspect this also needs to be done for the 3.0 branch, just looked at it.

xrmx:
Thanks for reporting, I've cherry-picked the original commit in the uwsgi-2.0 branch

Issue #2202, opened 2020-07-09
Milliseconds in format log-encoder


maksym-iv:
A lot of third party logging aggregators expect timestamp to be in milliseconds. For this moment UWSGi `format` log encoder support:
* `unix` - seconds from EPOCH
* `micros` - microseconds from EPOCH

`format` should support `millis` also as it is in JSON log encoder. 

Similar with #1631

xrmx:
Can we close this right?

PR #2203, opened 2020-07-09
Adding millis to `format` log encoder

xmackex:
This PR adds `millis` timestamp to `format` log encoder. Also some whitespaces were stripped.

A lot of third party logging aggregators expect timestamp to be in milliseconds. For this moment UWSGi `format` log encoder support:
* `unix` - seconds from EPOCH
* `micros` - microseconds from EPOCH

`format` log encoder should support `millis` also as it is in JSON log encoder. 

Related to #2202

### DIFF
--- a/core/logging.c
+++ b/core/logging.c
@@ -325,7 +325,7 @@ void logto(char *logfile) {
 
 
 void uwsgi_setup_log() {
-	
+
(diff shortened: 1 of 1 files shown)


rdeioris:
@xrmx safe for backport.

@xmackex thanks!

PR #2208, opened 2020-07-16
proto: Close the socket cleanly

chaitu-tk:
In some cases we see that the uwsgi doesn't close the socket (eg:
sending FIN for TCP) that can cause timeouts as the other end waits for
the connection to close, use shutdown to make sure we close before
actually closing the socket FD.

This was observed when using Nginx + uWSGI, where nginx expects the
connection to be closed once uWSGI is done sending the data, but the
close just closes the local socket descriptor but doesn't send close
(verified with both UNIX and TCP sockets, no FIN sent for TCP).


chaitu-tk:
image
Attaching the screenshot of the Wireshark capture to illustrate the timeout because of socket closing issue.

1st marked packet (in black): After uWSGI sends all its data to Nginx, it closes the connection
2nd marked packet (in black): uWSGI sends its data to Nginx, but doesn't close the connection (not sure about the root cause though), Nginx waits for 10s (configured uwsgi_send_timeout in Nginx) and closes the connection logging a timeout.

PR #2212, opened 2020-07-30
Graceful Harakiri

bom-d-van:
There two commits for this new features, one in core, which contains the infrastructure for supporting graceful-harakiri. Another is psgi, which makes use of the new feature. Currently, the harakiri delay is only useful if plugins wants to do some special operations on the worker (like dumping stacktrace).

Bellow are the commit messages:

## core: introduce a harakiri graceful period mechanism

Usually when a harakiri timeout is triggered, uwsgi master fires an SIGKILL to
the worker immediately. There is no time for the app to perform some operations
on the worker like triggering a stacktrace dump or generates a notification.
This commit introduces a harakiri delay time to make it easy and possible. It is
up to plugins like psgi to set a harakiri_delayed_at timestamp to tell the
master if it wants to give some time for the plugin to perform some special
operations like dumping stacktraces before the worker is killed.œ

The new delay time is controlled by a new option: harakiri-delay. By default,
no harakiri delay is applied.

## plugins/psgi: add a harakiri handler and a pre-harakiri command

This commits makes use of the new harakiri graceful period mechanism. If a user
specifies the psgi-exec-pre-harakiri option, which is a separate program, then
psgi will trigger the program before harakiri and sets the harakiri_delayed_at,
which stops master from killing the worker immediately and delayed it after
harakiri-delay time (default is 0s).

There two environment variables is set for the `psgi-exec-pre-harakiri` command:

	UWSGI_HARAKIRI_WORKER_PID
	UWSGI_HARAKIRI_REQUEST_URI

A config example:

	harakiri-delay = 3
	psgi-exec-pre-harakiri = /bin/send_harakiri_event_with_stacktrace

### DIFF
--- a/core/init.c
+++ b/core/init.c
@@ -188,6 +188,8 @@ void uwsgi_init_default() {
 	uwsgi_master_fifo_prepare();
 
 	uwsgi.notify_socket_fd = -1;
+
+	uwsgi.harakiri_options.delay = 0;
 }
 
 void uwsgi_setup_reload() {
--- a/core/master_checks.c
+++ b/core/master_checks.c
@@ -186,9 +186,13 @@ int uwsgi_master_check_workers_deadline() {
 	for (i = 1; i <= uwsgi.numproc; i++) {
 		/* first check for harakiri */
 		if (uwsgi.workers[i].harakiri > 0) {
-			if (uwsgi.workers[i].harakiri < (time_t) uwsgi.current_time) {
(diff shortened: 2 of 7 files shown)


bom-d-van:
After this PR is accepted, I can create another PR for master (there are some conflicts against master at the moment).

bom-d-van:
Hi @rdeioris @unbit @aldur (sorry, not sure who else I should ping), what do you think of this change? Any chance of getting it reviewed and merged?

PR #2236, opened 2020-10-07
python: Use new Py_SET_SIZE API introduced in Python 3.9

darktohka:
Python 3.9 implements a new API method: `Py_SET_SIZE` which is used for setting a PyVarObject's type: https://docs.python.org/3.9/c-api/structures.html#c.Py_SET_SIZE

In Python 3.10, `Py_SIZE` was changed from a macro to a function, which changes its fundamental syntax: https://docs.python.org/3.10/c-api/structures.html#c.Py_SIZE

This causes a compile error using Python 3.10:

```
plugins/python/uwsgi_pymodule.c:1825:15: error: expression is not assignable
          Py_SIZE(ret) = rlen;
           ~~~~~~~~~~~~ ^
```

Therefore, we should migrate the usage of Py_SIZE to set the object size to the new Py_SET_SIZE API, which was specifically made for this purpose.

For all versions below Python 3.9, `Py_SET_SIZE` is implemented by a macro with the old functionality. This even allows us to remove an old Python 2.6 compatibility macro.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -38,3 +38,4 @@ Alexandre Bonnetain
 Darvame Hleran
 Sokolov Yura <funny.falcon@gmail.com>
 Marcin Lulek <info@webreactor.eu>
+Derzsi Dániel <daniel@tohka.us>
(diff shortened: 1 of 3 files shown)


xrmx:
@unbit do you still care about python2.6?

darktohka:
This PR is still compatible with Python 2.6. Merging it should not cause problems with Python 2.6.

unbit:
@xrmx i think the patch is fully backward compatible and safe for backport (and yes, unfortunately we still care 2.6 :( )

PR #2245, opened 2020-11-06
Fix logfile-chown and logfile-chmod when logfile is rotated

mkhon:


mkhon:
This also obsoletes #2204

PR #2248, opened 2020-11-09
Document default value for http-timeout option

### LINKED ISSUE #2183: Missing default for http-timeout option in docs
Hi folks,
I spent a significant amount of time trying to find out why uWSGI was breaking HTTP connections after 60 seconds, and at some point I found out that setting the `http-timeout` option to a number higher than 60 gave me more than 60 seconds for timing out. So I concluded that that option has a default of 60, but that's not documented in the uWSGI docs website.
I recommend that information to be added, to avoid other developers getting puzzled by this.
Cheers,
Diogo

tumb1er:
Closes #2183


idlesign:
Fixed in 997c6df503986b2152255fd9ef1b550abef81003

PR #2251, opened 2020-11-20
Enable the py-tracebacker for mules

maugt:
The current traceback system is great for giving visibility into the callstacks for serving requests, but when you use mules you don't get to see the same data.

This adds a new configuration option, --py-tracebacker-mule, which works the same way as --py-tracebacker. I added a new option because I didn't want to break anyone who was using the old behavior (although an earlier version of this change simply uses the same base path and adds -mule as a suffix between the path and the number). This results in in the same sockets being created as before and you can then read the stacks from them.

This is really useful in my usage, and I hoped others might find it useful enough that it should be a core feature.


maugt:
Hmm guess I merged this into the wrong branch, will resubmit

maugt:
Actually, I take that back, this does seem to be the right branch, the build is just broken

Issue #1950, opened 2018-12-20
uWSGI segfaults when WSGI environment lives longer than request and --wsgi-env-behavior=holy


untitaker:
## Application

Write to file called `app.py`:

```python
import threading, time, random

known_envs = []

def application(environ, start_response):
    for e in known_envs:
        repr(e)  # triggers segfault

    if not any(environ is e for e in known_envs):
        known_envs.append(environ)

    # This is just hello world
    status = '200 OK'
    response_headers = [('Content-type', 'text/plain')]
    start_response(status, response_headers)
    yield b'.'
```

## uWSGI options

Run this shell script:

```shell
export UWSGI_WORKERS=3
export UWSGI_THREADS=4

export UWSGI_DISABLE_LOGGING=true
export UWSGI_HTTP_SOCKET=0.0.0.0:9000
export UWSGI_HTTP_TIMEOUT=30
export UWSGI_WSGI_ENV_BEHAVIOR=holy
export UWSGI_PROTOCOL=http
export UWSGI_MODULE=app:application

exec uwsgi
```

## Client invocation

Just blast a bunch of concurrent requests against the server. For example with [ApacheBench](https://httpd.apache.org/docs/2.4/programs/ab.html):

```shell
ab -n 999 -c 10 http://127.0.0.1:9000/
```

## Output

<details><summary>Linux</summary>

```
!!! uWSGI process 15027 got Segmentation Fault !!!
!!! uWSGI process 15029 got Segmentation Fault !!!
*** backtrace of 15027 ***
uwsgi(uwsgi_backtrace+0x2a) [0x55aad7387b5a]
uwsgi(uwsgi_segfault+0x23) [0x55aad7387f43]
/lib/x86_64-linux-gnu/libc.so.6(+0x3ef20) [0x7fa052895f20]
/usr/lib/x86_64-linux-gnu/libpython3.6m.so.1.0(_PyDict_Next+0x150) [0x7fa052e62900]
/usr/lib/x86_64-linux-gnu/libpython3.6m.so.1.0(+0x222781) [0x7fa052e6a781]
/usr/lib/x86_64-linux-gnu/libpython3.6m.so.1.0(PyObject_Repr+0x53) [0x7fa052e575f3]
/usr/lib/x86_64-linux-gnu/libpython3.6m.so.1.0(_PyCFunction_FastCallDict+0x13a) [0x7fa052e5b96a]
[...]

Issue #1497, opened 2017-03-23
Crashes with wsgi-env-behavior=holy


ipmb:
We run a number of Python sites in separate uwsgi instances, both Python 2 and 3. After running into #1408 on one of our Python 3 apps, we set the `wsgi-env-behavior` to `holy` in all our uWSGI (v2.0.14) instances.

On a few (but not all) of our Python 2 instances, we started seeing #1111.

For now, we are running our Python 2 sites without the `holy` behavior and the Python 3 sites with it. I'm not clear whether or not this is the correct solution or if there is a deeper issue in uWSGI we're missing.

a-feld:
@ipmb uWSGI v2.0.15 contains a fix which prevents the `SystemError` issue with `wsgi-env-behavior` set to `cheat`.

If you upgrade to v2.0.15 and set `wsgi-env-behavior=cheat` does that eliminate the `SystemError`?

socketbox:
Can this be closed as a result of #2280 ?

ipmb:
I'm no longer using uWSGI on any Python 2 projects, so I'm unable to confirm.

Issue #1478, opened 2017-03-01
SEGVs + WSGIRequest object corruption on master


unixwitch:
Using master@f262899bc5eb890007ba157d8b1da1e836049b74, we've been seeing strange problems with Django applications.  The first is that WSGIRequest will be randomly missing attributes it should have:

```
AttributeError: 'WSGIRequest' object has no attribute 'environ'
```

or

```
AttributeError: 'WSGIRequest' object has no attribute 'path'
```

The second is intermittent SEGVs:

```
[kidneycareukwagtail] - !!! uWSGI process 16106 got Segmentation Fault !!!
[kidneycareukwagtail] - *** backtrace of 16106 ***
[kidneycareukwagtail] uWSGI worker 1(uwsgi_backtrace+0x30) [0x46a040]
[kidneycareukwagtail] uWSGI worker 1(uwsgi_segfault+0x21) [0x46a441]
/lib/x86_64-linux-gnu/libc.so.6(+0x350e0) [0x7f36294e70e0]
/opt/tbx/python/3.4/lib/libpython3.4m.so.1.0(+0x19295d) [0x7f362890c95d]
/opt/tbx/python/3.4/lib/libpython3.4m.so.1.0(+0xa67af) [0x7f36288207af]
/opt/tbx/python/3.4/lib/libpython3.4m.so.1.0(+0x192c27) [0x7f362890cc27]
/opt/tbx/python/3.4/lib/libpython3.4m.so.1.0(_PyObject_GC_Malloc+0xeb) [0x7f362890de9b]
/opt/tbx/python/3.4/lib/libpython3.4m.so.1.0(PyType_GenericAlloc+0x3b) [0x7f362883ef0b]
/opt/tbx/python/3.4/lib/libpython3.4m.so.1.0(+0xca223) [0x7f3628844223]
/opt/tbx/python/3.4/lib/libpython3.4m.so.1.0(PyObject_Call+0x68) [0x7f36287e1b48]
/opt/tbx/python/3.4/lib/libpython3.4m.so.1.0(PyEval_EvalFrameEx+0x4bcc) [0x7f36288c33fc]
/opt/tbx/python/3.4/lib/libpython3.4m.so.1.0(PyEval_EvalCodeEx+0x886) [0x7f36288c70f6]
/opt/tbx/python/3.4/lib/libpython3.4m.so.1.0(PyEval_EvalFrameEx+0x608a) [0x7f36288c48ba]
[...]

Issue #1111, opened 2015-11-11
worker dies by signal 6 + Fatal Python error: Inconsistent interned string state.


thestick613:
I keep getting these in my logs:

    DAMN ! worker 1 (pid: 12639) died, killed by signal 6 :( trying respawn ...
    Respawned uWSGI worker 1 (new pid: 12644)
    DAMN ! worker 5 (pid: 12641) died, killed by signal 6 :( trying respawn ...
    Respawned uWSGI worker 5 (new pid: 12645)
    DAMN ! worker 12 (pid: 12642) died, killed by signal 6 :( trying respawn ...
    Respawned uWSGI worker 12 (new pid: 12646)
    DAMN ! worker 9 (pid: 12643) died, killed by signal 6 :( trying respawn ...
    Respawned uWSGI worker 9 (new pid: 12647)

And lots of:

    Fatal Python error: Inconsistent interned string state.

I'm running

    uwsgi 2.0.9
    Python 2.7.10
    nginx version: nginx/1.6.2

uwsgi is running like this:

    uwsgi --ini /home/www-data/webserver/uwsgi_conf.conf --wsgi-env-behaviour=holy

/home/www-data/webserver/uwsgi_conf.conf

    [uwsgi]
    socket = /socket/my_app.uwsgi
    stats = /socket/stats_socket.uwsgi
    chdir = /home/www-data/
    limit-as = 14000
    processes = 32
    threads = 1
    master = true
    max-requests = 2048
    max-worker-lifetime = 1800
    wsgi-file = webserver.py

nginx relevant config

    server {
            listen 80;
            server_name Minifier;
            location / {
                    include uwsgi_params;
                    uwsgi_pass unix:///socket/my_app.uwsgi;
            }
    }

my app is something very basic, such as:

    import gevent.pywsgi as pywsgi
    import wsgiref

    def application(environ, start_response):
        if YYY:
            start_response("302 Found", headers + [("Location", "XXX"),])
        else:
            start_response('404 Not Found', headers)
        return [""]
    if __name__ == "__main__":
            server = pywsgi.WSGIServer(('', 8888), application=application, backlog=1024, log=None)
            server.serve_forever()

I have quite some errors in nginx logs:

    2015/11/11 13:26:29 [error] 6623#0: *199487 recv() failed (104: Connection reset by peer) while reading response header from upstream
    2015/11/11 12:37:56 [error] 6626#0: *109585 upstream prematurely closed connection while reading response header from upstream
[...]

PR #2280, opened 2021-01-22
fix segfaults when using --wsgi-env-behavior=holy.

antocuni:
In the issue trackers there are tons of issue related to this, e.g. #1497, #1950, #1478, #1111, etc. I think there is a good chance that this PR fixes them all.

The crash is very obvious and very reproducible if you use a debug build of python. On my machine, I can reproduce it easily with this, with uwsgi's current master (fb22797d):
```
# main.py
def app(environ, start_response):
    status = '200 OK'
    response_headers = [('Content-type', 'text/plain')]
    return [b'Hello world!\n']

# uwsgi.ini
[uwsgi]
cheaper = 1
processes = 2
module = main
callable = app
show-config = true
wsgi-env-behavior = holy
```

Without this PR, as soon as do a request, you see the following output:
```
...
spawned uWSGI master process (pid: 2104137)
spawned uWSGI worker 1 (pid: 2104138, cores: 1)
spawned uWSGI http 1 (pid: 2104139)
plugins/python/python_plugin.c:1110: _Py_NegativeRefcount: Assertion failed: object has negative ref count
Enable tracemalloc to get the memory block allocation traceback

object address  : 0x7f99b57fca70
object refcount : -1
object type     : 0x7f99b644ca00
object type name: dict
object repr     : <refcnt -1 at 0x7f99b57fca70>

Fatal Python error: _PyObject_AssertFailed
Python runtime state: initialized

Current thread 0x00007f99b5f04740 (most recent call first):
<no Python frame>
DAMN ! worker 1 (pid: 2104138) died, killed by signal 6 :( trying respawn ...
Respawned uWSGI worker 1 (new pid: 2104157)
...
```
With this PR, the problem is solved.

I also saw commit https://github.com/unbit/uwsgi/commit/80bcf021242b3985ecd366c6ebfb44baed87adba but I think it's just a workaround and it's conceptually wrong. This should be the correct fix.

### DIFF
--- a/plugins/python/wsgi_subhandler.c
+++ b/plugins/python/wsgi_subhandler.c
@@ -255,6 +255,7 @@ void *uwsgi_request_subhandler_wsgi(struct wsgi_request *wsgi_req, struct uwsgi_
 
 	// call
 	if (PyTuple_GetItem(wsgi_req->async_args, 0) != wsgi_req->async_environ) {
+	    Py_INCREF(wsgi_req->async_environ);
 	    if (PyTuple_SetItem(wsgi_req->async_args, 0, wsgi_req->async_environ)) {
 	        uwsgi_log_verbose("unable to set environ to the python application callable, consider using the holy env allocator\n");
 	        return NULL;


xrmx:
Thanks!

PR #2284, opened 2021-01-31
core, proto: replace uwsgi.h system includes

q3k:
This replaces all `#include <uwsgi.h>` in core/ and proto/ with `#include "uwsgi.h"`.

This is to make it easier to build uWSGI with build systems that are more opinonated about local-vs-system include directories (eg. Bazel), and is without detriment to uWSGI itself (at least as far as I can tell).


xrmx:
Thanks!

Issue #2259, opened 2020-11-29
Usage of NID_X9_62_prime256v1 results in weaker key exchange rating


spacesynth:
Usage of ```NID_X9_62_prime256v1``` results in weaker key exchange rating.
It should be dropped in favour of the more secure ```NID_secp384r1```.

As Let's Encrypt certs and all state of the art browser Chrome, Firefox, etc. support P384, you downgrade the security by performing the actual key exchange only over a 256-bit prime field.

I will attempt to recompile an own version supporting it in the meantime. Literally NGINX and regular SSL context support defining either ```set_ecdh_curve secp384r1;``` or ```sslcontext.set_ecdh_curve("secp384r1")```.

It should be an option in here as well.

vaporwave9:
Two line fix at:
https://github.com/unbit/uwsgi/blob/master/core/ssl.c#L278
```
#ifdef NID_secp384r1
        EC_KEY *ecdh = EC_KEY_new_by_curve_name(NID_secp384r1);
```
Followed by building from source:
```python3 uwsgiconfig.py --build```
Yields:
![grafik](https://user-images.githubusercontent.com/64961250/100613742-4fbc4380-3315-11eb-9ab0-8eb8c11cd7f8.png)
![grafik](https://user-images.githubusercontent.com/64961250/100613752-534fca80-3315-11eb-864a-2fc38bad857e.png)

Pros:
Future ready for crypto freaks.
Con:
Some 10 year old legacy device might have trouble connecting.

vaporwave9:
I am still watching this and interested in an official statement.

vaporwave9:
I am still watching this with interest in an official statement.

PR #2285, opened 2021-02-14
Increase curve strength from 256-bits to 384-bits of security.

spacesynth:
Increase curve strength from 256-bits to 384-bits of security.

This closes issue: https://github.com/unbit/uwsgi/issues/2259

### DIFF
--- a/core/ssl.c
+++ b/core/ssl.c
@@ -275,8 +275,8 @@ SSL_CTX *uwsgi_ssl_new_server_context(char *name, char *crt, char *key, char *ci
         }
 #if OPENSSL_VERSION_NUMBER >= 0x0090800fL
 #ifndef OPENSSL_NO_ECDH
-#ifdef NID_X9_62_prime256v1
-        EC_KEY *ecdh = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1);
+#ifdef NID_secp384r1
+        EC_KEY *ecdh = EC_KEY_new_by_curve_name(NID_secp384r1);
         if (ecdh) {
                 SSL_CTX_set_options(ctx, SSL_OP_SINGLE_ECDH_USE);
                 SSL_CTX_set_tmp_ecdh(ctx, ecdh);


vaporwave9:
I am still watching this and interested in an official statement.

vaporwave9:
I am still watching this with interest in an official statement.

core/ssl.c
```diff
@@ -275,8 +275,8 @@ SSL_CTX *uwsgi_ssl_new_server_context(char *name, char *crt, char *key, char *ci
         }
 #if OPENSSL_VERSION_NUMBER >= 0x0090800fL
 #ifndef OPENSSL_NO_ECDH
-#ifdef NID_X9_62_prime256v1
-        EC_KEY *ecdh = EC_KEY_new_by_curve_name(NID_X9_62_prime256v1);
+#ifdef NID_secp384r1
```
xrmx, line 278: What about keeping support for both by moving this check before the other and transforming the other one in al elif ? I don't know if both are available for the same version of openssl so i think it would be better to avoid any eventual downgrading if that's the case.

vaporwave9: @xrmx 
Yes support for both is preferable and your idea would be the ideal solution for me. Considering I ran it like this on older Debian releases, this should support really old OpenSSL versions. I highly doubt you can realistically find old OpenSSL versions in use as Debian backports all security relevant releases.

Both Buster and Bullseye use openssl-1.1.1n, original was openssl-1.1.1d in 2019. Since there is only TLS1.2 and TLS1.3 now both are fully supported anyway, as TLS1.x is fully deprecated. What distros and python distributions did you have in mind with old openssl dependencies? Since Python 2.x is deprecated as well now.

I'm just a security freak by the way, using as strong crypto as software can possibly offer me. It felt sensible to use P384 key exchange and P384 signature schemes when both offered and supported by letsencrypt. In the wild you will find odd combinations like x25519 + RSA as on the Guardian's website.

spacesynth:
Closing due to inactivity of the developer.

PR #2296, opened 2021-03-19
using platform except os uname for windows support

m-moein98:


xrmx:
Thanks for the pr. uwsgi does not support windows so having python scripts work on windows would not be something to pursue in my opinion.

Issue #2238, opened 2020-10-12
cannot build 2.0.18 with clang 11


drook:
I cannot build 2.0.18 with clang 11 on FreeBSD 13.x:

OS:
````
# uname -a
FreeBSD cg-mr-prod-stg09 13.0-CURRENT FreeBSD 13.0-CURRENT #0 r365569: Fri Sep 11 08:35:53 UTC 2020     root@cg-mr-prod-stg09:/usr/obj/usr/src/amd64.amd64/sys/GENERIC  amd64
````
clang:
````
# clang -v
FreeBSD clang version 11.0.0 (git@github.com:llvm/llvm-project.git llvmorg-11.0.0-rc2-0-g414f32a9e86)
Target: x86_64-unknown-freebsd13.0
Thread model: posix
InstalledDir: /usr/bin
````

````
[...]

PR #2305, opened 2021-04-15
Fix compilation with clang11 on FreeBSD 13.0

karolyi:
This fixes the compilation error mentioned in https://github.com/unbit/uwsgi/issues/2238#issuecomment-819929450.

### DIFF
--- a/plugins/gevent/gevent.c
+++ b/plugins/gevent/gevent.c
@@ -389,7 +389,7 @@ static void gil_gevent_get() {
 }
 
 static void gil_gevent_release() {
-	PyGILState_Release((PyGILState_STATE) pthread_getspecific(up.upt_gil_key));
+	PyGILState_Release((long) pthread_getspecific(up.upt_gil_key));
 }
 
 static void monkey_patch() {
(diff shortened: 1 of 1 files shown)


karolyi:
@xrmx can you please merge this ASAP? it is blocking a lot of my my uwsgi deployments.

karolyi:
bump

plugins/gevent/gevent.c
```diff
@@ -389,7 +389,7 @@ static void gil_gevent_get() {
 }
 
 static void gil_gevent_release() {
-	PyGILState_Release((PyGILState_STATE) pthread_getspecific(up.upt_gil_key));
+	PyGILState_Release((long) pthread_getspecific(up.upt_gil_key));
```
xrmx, line 392: Please keep the proper type `(PyGILState_STATE)(long)`

karolyi: Nice! I'm not a C expert so I wouldn't know.

I have looked into the code and the function `pthread_getspecific` has different casts throughout all the code. Is that intentional? If not, it's just as a security risk as this one, imo.

xrmx: It returns void * so you are supposed to cast it to whatever you set it before to.

xrmx:
Merged, thanks!

PR #2314, opened 2021-05-20
uwsgiconfig: Drop Homebrew-specific workaround

andreabolognani:
This is a tentative pull request because the necessary changes to Homebrew haven't been accepted yet.

In https://github.com/Homebrew/homebrew-core/pull/74516, the libvirt developers have been attempting to get the yajl issues that https://github.com/unbit/uwsgi/commit/80c3d6528b257ab9e173fda130b55141e7697f3c was intended to work around addressed in Homebrew. We have our own, similar workaround already in place, but we believe that the issue would be better fixed once and for all in Homebrew rather than worked around separately in each project that uses yajl and wants to support macOS.

The Homebrew maintainers have, quite understandably, been reluctant to introduce and carry a downstream-only patch. In [a recent comment](https://github.com/Homebrew/homebrew-core/pull/74516#issuecomment-843698702) however, the maintainers declared themselves willing to consider incorporating the fix under the condition that at least one more project in addition to libvirt would ask for that to happen.

Would uwsgi developers be willing to join the libvirt developers in asking for this bug in yajl to be addressed in Homebrew instead of being worked around in each project that uses the library?

Thank you for your consideration!

### DIFF
--- a/uwsgiconfig.py
+++ b/uwsgiconfig.py
@@ -1253,8 +1253,6 @@ class uConf(object):
                 else:
                     jsonconf = spcall("pkg-config --cflags yajl")
                     if jsonconf:
-                        if jsonconf.endswith('include/yajl'):
-                            jsonconf = jsonconf.rstrip('yajl')
                         self.cflags.append(jsonconf)
                         self.cflags.append("-DUWSGI_JSON")
                         self.gcc_list.append('core/json')


xrmx:
Won't it break with people with an older version installed from brew though?

andreabolognani:
> If we remove the workaround from our side won't it break with people with an older version of yajl installed from brew though?

Yeah, depending on your platform support policy you might not be able to drop the workaround right away, but after some sensible delay it's going to be okay to assume people have upgraded their system. The exact definition of "sensible" in this context is left to you of course ;)

With that in mind, would you be willing to voice your support for this change being made in Homebrew, even though it might be a while before you can actually reap the benefits? If so, please drop a comment in https://github.com/Homebrew/homebrew-core/pull/74516 indicating this and clearly stating your association with the uwsgi project. Thank you!

andreabolognani:
Hi, just a polite ping :)

Any chance you could go to https://github.com/Homebrew/homebrew-core/pull/74516 and show your support for working around the yajl pkg-config issue in Homebrew instead of yajl-using applications?

Thanks!

andreabolognani:
Hi, another polite ping :)

It'd be great if you would consider supporting this change in Homebrew! Doing so should only take a few minutes of your time.

Thanks :)

PR #2316, opened 2021-05-28
Log SIGINT/SIGTERM triggered kill_them_all

delenamalan:
Log `SIGINT/SIGTERM` instead of `SIGTERM/SIGQUIT` in `kill_them_all` because `SIGINT` and `SIGTERM` are the two signals that will run that function:

https://github.com/unbit/uwsgi/blob/d58a832c81c2c96ae0f6e72614e1cc47f4b5d332/core/master.c#L374-L378


xrmx:
Thanks!

PR #2323, opened 2021-06-07
Ssl socket activation

steelman:
If I am not mistaken, it isn't possible at the moment to configure uWSGI to receive sockets from systemd **and** use them as SSL sockets. These patches (actually the last one) makes uWSGI grab SSL context from `uwsgi_socket` configured in a config file and apply it to a socket received from systemd (while detaching it from the original structure).

For an SSL configuration to be applied to a socket received from systemd its computed name must match the name configured with `https-socket`.

**WARNING!** Although I have tested it for both IPv4 and IPv6 and it works for me, I am not entirely sure this is the right way to do it. Please, review the patches carefully.

### DIFF
--- a/core/setup_utils.c
+++ b/core/setup_utils.c
@@ -14,7 +14,7 @@ void uwsgi_setup_systemd() {
 			if (listen_fds) {
 				int systemd_fds = atoi(listen_fds);
 				if (systemd_fds > 0) {
-					uwsgi_log("- SystemD socket activation detected -\n");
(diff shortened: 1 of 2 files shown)


unbit:
@xrmx looks good to me, any thoughts ? thanks @steelman

steelman:
I don't exactly understand why the error occurred. Is there anything I need to fix? Can you explain?

steelman:
Ping?

PR #2331, opened 2021-06-22
Allow to compile on Python versions with more that two digits (Python 3.10)

jrd:
The fix is really simple. I took the value `30` as big enough to carry further various version schemas.

### DIFF
--- a/plugins/python/pyloader.c
+++ b/plugins/python/pyloader.c
@@ -22,7 +22,7 @@ PyMethodDef uwsgi_eventfd_write_method[] = { {"uwsgi_eventfd_write", py_eventfd_
 void set_dyn_pyhome(char *home, uint16_t pyhome_len) {
 
 
-	char venv_version[15];
+	char venv_version[30];
 	PyObject *site_module;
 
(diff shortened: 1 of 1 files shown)


plugins/python/pyloader.c
```diff
@@ -22,7 +22,7 @@ PyMethodDef uwsgi_eventfd_write_method[] = { {"uwsgi_eventfd_write", py_eventfd_
 void set_dyn_pyhome(char *home, uint16_t pyhome_len) {
 
 
-	char venv_version[15];
+	char venv_version[30];
```
xrmx, line 25: Shouldn't one more char be enough?

jrd: Yes you could use `16` chars if you like. But what if one day there is a Python `10.402`? What is the gain of having *just-enough* bytes? If you prefer, you can use `16`. I left the right for the reviewer to make corrections.

jrd:
What is blocking thiis MR?

unbit:
@xrmx safe for backport

PR #2332, opened 2021-06-25
take into account new naming for LIBPL since python 3.6

ilrico:
related to issues 2270 and 2293


unbit:
@xrmx safe for backport. @ilrico thanks !

Issue #1935, opened 2018-11-29
uwsgi routes / to the first mounted app, even though path doesn't match route


reece:
uwsgi 2.0.17.1, Python 3.6

# Problem
uwsgi routes `/` to the first mounted app, even though / does not match the route mount path. 

# Steps to reproduce:
```
snafu$ python3.6 -mvenv venv/3.6
snafu$ . venv/3.6/bin/activate
(3.6) snafu$ pip install -U pip setuptools
(3.6) snafu$ pip install uwsgi flask

(3.6) snafu$ head bug1.py uwsgi.ini
==> bug1.py <==
from flask import Flask
app = Flask(__name__)

@app.route("/")
def hello():
    return "You may call me " + __name__ + "\n"

==> uwsgi.ini <==
[uwsgi]
mount = /bug1=bug1.py
callable = app
http-socket = :9090

```

Start with `uwsgi --ini=uwsgi.ini`. In another window:

```
(3.6) snafu$ for url in http://localhost:9090/{bug1/,bug/,}; do (set -x; curl $url); done
+ curl http://localhost:9090/bug1/
You may call me uwsgi_file_bug1
+ curl http://localhost:9090/bug/
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>404 Not Found</title>
<h1>Not Found</h1>
<p>The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again.</p>
+ curl http://localhost:9090/
You may call me uwsgi_file_bug1

```

Logs in first window:
```
[pid: 19540|app: 0|req: 1/1] 127.0.0.1 () {24 vars in 242 bytes} [Wed Nov 28 21:21:34 2018] GET /bug1/ => generated 32 bytes in 0 msecs (HTTP/1.1 200) 2 headers in 79 bytes (1 switches on core 0)
[pid: 19540|app: 0|req: 2/2] 127.0.0.1 () {24 vars in 240 bytes} [Wed Nov 28 21:21:34 2018] GET /bug/ => generated 233 bytes in 8 msecs (HTTP/1.1 404) 2 headers in 72 bytes (1 switches on core 0)
[pid: 19540|app: 0|req: 3/3] 127.0.0.1 () {22 vars in 232 bytes} [Wed Nov 28 21:21:34 2018] GET / => generated 32 bytes in 0 msecs (HTTP/1.1 200) 2 headers in 79 bytes (1 switches on core 0)
```

[...]

PR #2343, opened 2021-09-10
Change errror code to 404 if no app is found

ApproximateIdentity:
Currently an error code of 500 is returned when no app is found for a route, but this changes it to a 404 which seems more appropriate. See this issue for more context:

https://github.com/unbit/uwsgi/issues/1935

Honestly I'm not sure if this change is really possible since maybe the currently functionality is required for compatibility. I tried running tests, but I had some issues. At the very least this appears to fix the problems for me so I figured I would open up a PR to solicit feedback.


ApproximateIdentity:
I made a slight change to the PR so that a 404 is only returned if no-default-app is set. I think my original change was a bit too broad.

PR #2363, opened 2021-11-02
Add PY_SSIZE_T_CLEAN define for Python 3.10 support

theacodes:
Python 3.10 started requiring `PY_SSIZE_T_CLEAN` to be defined when using `PyArg_ParseTuple` with formats that deal with object sizes (see [release notes](https://docs.python.org/3.10/whatsnew/3.10.html#id2)). Without this, invoking a Python extension module method that hits that code path (like `uwsgi.mule_msg`) causes a SystemError:

```
  ...
  File "./venv/lib/python3.10/site-packages/uwsgidecorators.py", line 191, in __call__
    return self.real_call(*args, **kwargs)
  File "./venv/lib/python3.10/site-packages/uwsgidecorators.py", line 183, in real_call
    uwsgi.mule_msg(data, self.mule)
SystemError: PY_SSIZE_T_CLEAN macro must be defined for '#' formats
```

This just defines the macro, since the Python module itself has already been updated to use `Py_ssize_t` instead of `int`. The define tells Python that none of your invocations are using an `int`.

### DIFF
--- a/plugins/python/uwsgi_python.h
+++ b/plugins/python/uwsgi_python.h
@@ -1,4 +1,6 @@
 #include <uwsgi.h>
+/* See https://docs.python.org/3.10/whatsnew/3.10.html#id2 */
+#define PY_SSIZE_T_CLEAN
 #include <Python.h>
 
 #include <frameobject.h>


HLFH:
Can it be merged? cc @xrmx 

Thanks!

xrmx:
Merged, thanks!

HLFH:
@xrmx I'd suggest to release/tag it as uwsgi 2.0.21.

MaxwellDPS:
Howdy,

Is there any plan to release this soon? Ref https://github.com/miguelgrinberg/python-engineio/issues/268

corydolphin:
@xrmx thank you for merging this change and supporting the uwsgi project in general! For other teams' planning purposes, is there a planned next release date we can orient around?

Thanks again for all you've done to build and share uwsgi,
Cory

minhaz1:
Is there a release planned that includes these changes? Although this doesn't completely break like in 3.10, it does produce deprecation warnings in 3.8 and 3.9 and would help to remove those.

bungoume:
I too am eagerly awaiting the release of this update.
This is the limitation that prevents us from updating to Python 3.10.

PR #2380, opened 2021-12-03
fix: avoid deadlocks in Python forked workers

majorgreys:


majorgreys:
This is unfortunately not ready yet to share but will do so soon!

Issue #2179, opened 2020-06-09
RuntimeError: cannot release un-acquired lock - when every new worker is spawned


tylerecouture:
I am running uwsgi (with django in a docker container, python 3.8), and every worker that spawns comes with this error:

    spawned uWSGI master process (pid: 1)
    web_1          | spawned uWSGI worker 1 (pid: 18, cores: 2)
    web_1          | Exception ignored in: <function _after_at_fork_child_reinit_locks at 0x7fd944954ca0>
    web_1          | Traceback (most recent call last):
    web_1          |   File "/usr/local/lib/python3.8/logging/__init__.py", line 260, in _after_at_fork_child_reinit_locks
    web_1          | spawned uWSGI worker 2 (pid: 19, cores: 2)
    web_1          |     _releaseLock()  # Acquired by os.register_at_fork(before=.
    web_1          |   File "/usr/local/lib/python3.8/logging/__init__.py", line 228, in _releaseLock
    web_1          |     _lock.release()
    web_1          | RuntimeError: cannot release un-acquired lock

Here is my uwsgi **.ini** file:

    [uwsgi]
    strict          = true          ; only explicitly understood configration allowed
    module          = myapp.wsgi
    need-app        = true          ; don't start if can't find or load django app
    
    ### PROCESS SETTINGS ###############
    
    master          = true          ; gracefully re-spawn and pre-fork workers, consolidate logs, etc
    enable-threads  = true
    
    processes       = 10             ; maximum number of worker processes
    threads         = 2
    thunder-lock    = true
    
    socket          = 0.0.0.0:8000  ; too complicated for me to get a unix socket to work with docker, this is fine.
    #socket          = /tmp/myapp.sock
    
    vacuum          = true          ; on exit, clean up any temporary files or UNIX sockets it created, such as HTTP sockets, pidfiles, or admin FIFOs
    single-interpreter = true
    die-on-term     = true
    
    ### WORKER MANAGEMENT ##########
    max-requests    = 1000          ; Restart workers after this many requests
    max-worker-lifetime = 3600      ; Restart workers after this many seconds
    reload-on-rss   = 2048          ; Restart workers after this much resident memory
    worker-reload-mercy = 60        ; How long to wait before forcefully killing workers
    
    harakiri        = 60            ; Forcefully kill workers after 60 seconds
    py-call-osafterfork = true      ; Allow workers to receive signals such as signal.alarm

    ### LOGGING ####################

[...]

PR #2388, opened 2021-12-26
fix(python-plugin): update internal interpreter state immediately upon forking 

### LINKED ISSUE #2387: cpython: deadlocked workers if thread started before forking
## Problem

This issue has been observed with CPython 2.7 and above. If a thread is started by the Python interpreter in the master (parent) process before the master process forks a worker (child) process, a worker can fail to respawn when the child process does not gracefully shutdown. The reproduction below was used to investigate the failure to respawn and we found that child processes deadlock on acquiring the GIL after they are forked. 

This issue is a duplicate of:

- https://github.com/unbit/uwsgi/issues/1149
- https://github.com/unbit/uwsgi/issues/1369

## Reproduction

See source code for this reproduction at https://github.com/majorgreys/uwsgi-deadlock-workers/.

This issue can reliably be reproduced by starting a thread in a `sitecustomize.py` that enters a busy loop. This ensures that the current thread in the Python interpreter is not the main thread when uWSGI forks a child process. The result is an invalid thread state in the child process, specifically the GIL is locked and will not have been recreated.

- We install the debug version of CPython so that we can debug uWSGI and the Python plugin with gdb. 
- The reproduction consists of a simple hello world wsgi application and a `sitecustomize` that starts a thread which enters a busy loop for 5 seconds. 
- The `uwsgi.ini` enables a master process and threads along with one worker process with a max worker lifetime of 15 seconds. This allows us to more easily observe the failed attempt to respawn the worker process.

To build and run the reproduction with CPython 2.7 and uWSGI 2.0.20, run:

```
docker run --rm -it $(docker build -q .)
```

If the worker was respawning properly, we would expect the output of the process to be:

```
spawned uWSGI master process (pid: 35)
spawned uWSGI worker 1 (pid: 37, cores: 1)
spawned uWSGI http 1 (pid: 38)
[thread=140112973076224] run: finished.
worker 1 lifetime reached, it was running for 16 second(s)
Respawned uWSGI worker 1 (new pid: 39)
worker 1 lifetime reached, it was running for 16 second(s)
Respawned uWSGI worker 1 (new pid: 40)
```

However, uWSGI will hang waiting on the worker to respawn with uWSGI 2.0.2 in this reproduction:

```
spawned uWSGI master process (pid: 36)
spawned uWSGI worker 1 (pid: 38, cores: 1)
spawned uWSGI http 1 (pid: 39)
[thread=140082567735040] run: finished.
worker 1 lifetime reached, it was running for 16 second(s)
```

If we attach to the process with `gdb` we find that the cause is a deadlock on the GIL acquisition:

```
#0  futex_abstimed_wait_cancelable (private=0, abstime=0x0, expected=0, futex_word=0x560bef494fd0) at ../sysdeps/unix/sysv/linux/futex-internal.h:205
#1  do_futex_wait (sem=sem@entry=0x560bef494fd0, abstime=0x0) at sem_waitcommon.c:111
#2  0x00007f1348060988 in __new_sem_wait_slow (sem=sem@entry=0x560bef494fd0, abstime=0x0) at sem_waitcommon.c:181
#3  0x00007f13480609f9 in __new_sem_wait (sem=sem@entry=0x560bef494fd0) at sem_wait.c:42
#4  0x00007f13475f72c4 in PyThread_acquire_lock (lock=0x560bef494fd0, waitflag=1) at Python/thread_pthread.h:356
#5  0x0000560bee9d30d9 in gil_real_get ()
#6  0x0000560bee9cba6c in uwsgi_python_master_fixup ()
#7  0x0000560bee97f5c6 in uwsgi_respawn_worker ()
#8  0x0000560bee97d83b in master_loop ()
#9  0x0000560bee9bb78a in uwsgi_run ()
#10 0x0000560bee968e7e in main ()
```

## Root cause

In CPython, the thread state of a child process and its parent process must be updated after forking if there will be subsequent calls to the Python interpreter. This is handled by `PyOS_AfterFork_Child` in CPython 3+ and `PyOS_AfterFork` in CPython 2.7. 

In uWSGI, there is an attempt to acquire the GIL in the `master_fixup` Python plugin hook before `PyOS_AfterFork[_Child]` is called in `post_fork` hook. This leads to a deadlock which explains the failure for workers to respawn.

majorgreys:
Fixes #2387

This PR adds `pre_uwsgi_fork` and `post_uwsgi_fork` hooks to `uwsgi_plugin` in order to update the internal interpreter state of CPython immediately before and immediately after the call to `uwsgi_fork`. This follows the implementation of the standard library `os.fork()` in CPython. 

While these two hooks effectively replace existing code to update the internal interpreter state during the lifecycle of master and worker processes, we preserve the current behavior for uWSGI LTS. We instead provide a new option `py-call-uwsgi-fork-hooks` to enable the new feature.

We add test coverage for the issue of deadlocked workers raised in #2387 parameterized for relevant configurations options.

### DIFF
--- a/uwsgi.h
+++ b/uwsgi.h
@@ -1095,6 +1095,9 @@ struct uwsgi_plugin {
 	int (*worker)(void);
 
 	void (*early_post_jail) (void);
+
+	void (*pre_uwsgi_fork) (void);
+	void (*post_uwsgi_fork) (int);
 };
 
 #ifdef UWSGI_PCRE
--- a/core/master_utils.c
+++ b/core/master_utils.c
@@ -747,9 +747,22 @@ int uwsgi_respawn_worker(int wid) {
 		pthread_mutex_lock(&uwsgi.threaded_logger_lock);
 	}
 
+
+	for (i = 0; i < 256; i++) {
+		if (uwsgi.p[i]->pre_uwsgi_fork) {
+			uwsgi.p[i]->pre_uwsgi_fork();
+		}
+	}
+
 	pid_t pid = uwsgi_fork(uwsgi.workers[wid].name);
 
 	if (pid == 0) {
+		for (i = 0; i < 256; i++) {
+			if (uwsgi.p[i]->post_uwsgi_fork) {
+				uwsgi.p[i]->post_uwsgi_fork(1);
+			}
+		}
+
 		signal(SIGWINCH, worker_wakeup);
 		signal(SIGTSTP, worker_wakeup);
 		uwsgi.mywid = wid;
@@ -812,6 +825,12 @@ int uwsgi_respawn_worker(int wid) {
 		uwsgi_error("fork()");
 	}
 	else {
+		for (i = 0; i < 256; i++) {
+			if (uwsgi.p[i]->post_uwsgi_fork) {
+				uwsgi.p[i]->post_uwsgi_fork(0);
+			}
+		}
+
 		// the pid is set only in the master, as the worker should never use it
 		uwsgi.workers[wid].pid = pid;
 
--- a/plugins/python/python_plugin.c
+++ b/plugins/python/python_plugin.c
@@ -195,6 +195,7 @@ struct uwsgi_option uwsgi_python_options[] = {
 #endif
 
 	{"py-call-osafterfork", no_argument, 0, "enable child processes running cpython to trap OS signals", uwsgi_opt_true, &up.call_osafterfork, 0},
+	{"py-call-uwsgi-fork-hooks", no_argument, 0, "call pre and post hooks when uswgi forks to update the internal interpreter state of CPython", uwsgi_opt_true, &up.call_uwsgi_fork_hooks, 0},
 
 	{"early-python", no_argument, 0, "load the python VM as soon as possible (useful for the fork server)", uwsgi_early_python, NULL, UWSGI_OPT_IMMEDIATE},
 	{"early-pyimport", required_argument, 0, "import a python module in the early phase", uwsgi_early_python_import, NULL, UWSGI_OPT_IMMEDIATE},
@@ -428,12 +429,20 @@ realstuff:
 
 void uwsgi_python_post_fork() {
 
+	// Need to acquire the gil when no master process is used as first worker
+	// will not have been forked like others
+	// Necessary if uwsgi fork hooks called to update interpreter state
+	if (up.call_uwsgi_fork_hooks && !uwsgi.master_process && uwsgi.mywid == 1) {
+		UWSGI_GET_GIL
+	}
+
 	if (uwsgi.i_am_a_spooler) {
 		UWSGI_GET_GIL
 	}	
 
 	// reset python signal flags so child processes can trap signals
-	if (up.call_osafterfork) {
+	// Necessary if uwsgi fork hooks not called to update interpreter state
+	if (!up.call_uwsgi_fork_hooks && up.call_osafterfork) {
 #ifdef HAS_NOT_PyOS_AfterFork_Child
 		PyOS_AfterFork();
 #else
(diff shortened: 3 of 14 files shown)


uwsgi.h
```diff
@@ -1039,6 +1039,8 @@ struct uwsgi_plugin {
 	void *data;
 	void (*on_load) (void);
 	int (*init) (void);
+	void (*pre_uwsgi_fork) (void);
```
unbit, line 1042: Can you move it at the end of the structure for maintaining ABI?

core/uwsgi.c
```diff
@@ -3338,14 +3338,6 @@ int uwsgi_start(void *v_argv) {
 		}
 	}
 
-	// master fixup
```
unbit, line 3341: you should leave this part in for not breaking other plugins

core/mule.c
```diff
@@ -69,12 +69,6 @@ void uwsgi_mule(int id) {
 
 		uwsgi_close_all_sockets();
 
-		for (i = 0; i < 256; i++) {
```
unbit, line 72: you should leave this part in for not breaking other plugins

unbit:
@majorgreys thanks a lot, my only doubt (in addition to the review comments) is about the old (and ugly, obviously)  up.call_osafterfork. Basically it loses any meaning with your patch so i think we can safely remove this (at least in master). For LTS i would probably introduce this new behaviour (albeit definitely correct) as an option. I know it is annoying but LTS grants the user that we do not break/change old stuff (honestly i am not sure about the implications, very probably only app-managed signal management will be affected, and in a positive way). @xrmx any thought?

core/uwsgi.c
majorgreys: @unbit I did not see the `master_fixup` hook used elsewhere in the repository but that doesn't mean it's not used by a plugin a user might have, so I can add this back. Are you aware of such a plugin?

majorgreys:
I've updated the change now to make it so the old behavior can be retained, specifically with regards to `call_osafterfork`. 

Given the needs of LTS, I'm inclined to start with the latter so it can be easily backported to the 2.0 branch and open a separate PR for the breaking changes so that they can end up on master. @unbit Does that work for what you had in mind for how to introduce these changes?

majorgreys:
@unbit @xrmx I have updated the PR to introduce this as an opt-in feature and avoid any breaking changes so it can be backported into the LTS 2.0 branch.

brettlangdon:
@unbit or @xrmx have you had a chance to look over this PR again? Is there anything else you would like us to update?

majorgreys:
👋🏽 @unbit Any update on this PR? I addressed the concerns you raised with regards to merging this into LTS.

xrmx:
Merged and broke master CI :)
Search for "ERROR: uWSGI did not start"
https://github.com/unbit/uwsgi/runs/5633992704?check_suite_focus=true

majorgreys:
@xrmx Thanks for the heads up. I'll see if I can reproduce this locally.

Superskyyy:
@majorgreys Hello, thanks for fixing the issue, but I was trying the new option `py-call-uwsgi-fork-hooks` and the following ignored exception prints out.
```

Exception ignored in: <function _after_at_fork_child_reinit_locks at 0x7fd01e7ecdc0>
  File "/home/gitpod/.pyenv/versions/3.8.13/lib/python3.8/logging/__init__.py", line 264, in _after_at_fork_child_reinit_locks
    _releaseLock()  # Acquired by os.register_at_fork(before=.
  File "/home/gitpod/.pyenv/versions/3.8.13/lib/python3.8/logging/__init__.py", line 232, in _releaseLock
    _lock.release()
RuntimeError: cannot release un-acquired lock
```

This was discussed here but remains unresolved, can you help to see if it is my env problem or something else? Really appreciated.

majorgreys:
@Superskyyy I need to still review the linked issue but I can confirm that I can reproduce this error. I don't see it when `--py-call-uwsgi-fork-hooks` is not used along with `--enable-threads`.

#2179 also reported the same exception when `--py-call-osafterfork` was used.

Code for minimal reproduction: https://gist.github.com/majorgreys/3ee1716b107ad80226b12145ae87276a

Issue #2356, opened 2021-10-14
Missing arginfo with PHP 8


jSML4ThWwBID69YC:
Hello, 

uWSGI with PHP 8 shows the following errors. The errors show up in the php error log. 

```
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_version() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_setprocname() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_worker_id() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_masterpid() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_signal() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_rpc() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_cache_get() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_cache_set() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_cache_update() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_cache_del() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_cache_clear() in Unknown on line 0
[14-Oct-2021 19:24:25 UTC] PHP Warning:  Missing arginfo for uwsgi_cache_exists() in Unknown on line 0
```
Simply starting uWSGI with the PHP 8 plugin loaded is enough to trigger this. This does not happen on PHP 7.
 
### System information
OS: FreeBSD 13.0p4
uWSGI: 2.0.20
PHP: PHP 8.0.11 (cli) (built: Oct  7 2021 16:35:42) ( NTS )

### uWSGI and plugin build command
```
make PROFILE=nolang
./uwsgi --build-plugin "plugins/php php"
```

gdamjan:
[...]

PR #2391, opened 2021-12-29
fix: missing arginfo when compiling against PHP 8

gdamjan:
fixes https://github.com/unbit/uwsgi/issues/2356


gdamjan:
should be backported to the uwsgi-2.0 branch too

xrmx:
Thanks @gdamjan !

PR #2402, opened 2022-02-02
Remove stray license headers

### LINKED ISSUE #2401: Licensing issue - some files have headers specifying a different license from the global one
While performing our OSS-related due diligence, we found that the following files are not marked with the GPL 2 linking exception:
- [apache2/mod_Ruwsgi.c](https://github.com/unbit/uwsgi/blob/master/apache2/mod_Ruwsgi.c)
- [apache2/mod_proxy_uwsgi.c](https://github.com/unbit/uwsgi/blob/master/apache2/mod_proxy_uwsgi.c)
- [plugins/router_xmldir/router_xmldir.c](https://github.com/unbit/uwsgi/blob/master/plugins/router_xmldir/router_xmldir.c)
- [apache2/mod_uwsgi.c](https://github.com/unbit/uwsgi/blob/master/apache2/mod_uwsgi.c)
- [core/uwsgi.c](https://github.com/unbit/uwsgi/blob/master/core/uwsgi.c)

They are marked solely as under GPL 2. Is this intentional or an oversight? We're just trying to fully understand our obligations with respect to this package. Thanks!

NielsDoucet:
Remove stray license headers that don't comply with the global GPL2 + linking exception licensing model adopted with version 2.

Fixes #2401

### DIFF
--- a/apache2/mod_Ruwsgi.c
+++ b/apache2/mod_Ruwsgi.c
@@ -1,27 +1,3 @@
-/*
-
- *** uWSGI/mod_Ruwsgi ***
-
- Copyright 2009-2010 Roger Florkowski <rflorkowski@mypublisher.com>
- Copyright 2009-2010 Unbit S.a.s. <info@unbit.it>
-
- This program is free software; you can redistribute it and/or
- modify it under the terms of the GNU General Public License
- as published by the Free Software Foundation; either version 2
- of the License, or (at your option) any later version.
-
- This program is distributed in the hope that it will be useful,
- but WITHOUT ANY WARRANTY; without even the implied warranty of
- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- GNU General Public License for more details.
-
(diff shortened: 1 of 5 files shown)


unbit:
Thanks a lot ! @xrmx safe for 2.0 obviously

xrmx:
I can understand removing the GPL headers where you were the sole copyright holder, but dropping names of the other copyright holders?

unbit:
@xrmx you are right, i did not noticed it, and maybe that two plugins should remain under GPL2 unless we have confirmation from the authors

NielsDoucet:
My apologies, I missed that too. Should I reinstate those 2 files' header?

What would be the way forward to make sure the package is fully compliant with the license as stated in the documentation?

lindenksv:
So...is this to say these files are under GPL 2 with no exceptions? Or that the original response from @rdi-mindesk on #2401 saying the lack of an exception was an oversight is still correct?

PR #2445, opened 2022-06-12
Use threading.current_thread, has existed since Python 2.6

hugovk:
`threading.currentThread` is deprecated since Python 3.10 and scheduled for removal in Python 3.12 (October 2023).

* https://github.com/python/cpython/pull/25174

We have an `ifdef` guard, but the `current_thread` version has existed since Python 2.6, the [oldest version supported here](https://github.com/unbit/uwsgi/blob/ca8bc6acc98b2c062b5b00668acd851e210fcbca/setup.py#L141), so we can simplify things.

* https://docs.python.org/2.6/library/threading.html#threading.current_thread

### DIFF
--- a/plugins/python/python_plugin.c
+++ b/plugins/python/python_plugin.c
@@ -1423,16 +1423,12 @@ void uwsgi_python_enable_threads() {
 }
 
 void uwsgi_python_set_thread_name(int core_id) {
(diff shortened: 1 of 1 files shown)


xrmx:
Looks like a nice cleanup, thanks!

Issue #2447, opened 2022-06-21
gcc crash(‽) building wheel on python:3.9.13-slim-bullseye


PeterJCLaw:
I'm hitting an error building wheels for uWSGI on Bullseye using the Python 3.9 slim docker image. I found this after moving from the equivalent buster image to the bullseye image, so I'm guessing that the debian version (or GCC version) are somehow related.

I'm assuming this is a crash as `gcc` seems to be exiting with code `-1` and no error messages. uWSGI's `uwsgiconfig.thread_compiler` function then error-exits, causing the build to fail.

This was originally observed in building a Docker image in GitHub Actions, however I've reproduced it in a cut-down image hosted locally.

I have not been able to reproduce this for:
- build targets other than `bdist_wheel` -- when `pip install` falls back to `setup.py install` that always works, as does running `setup.py install` directly
- single threaded builds (via `CPUCOUNT=1` environment value, which suggests a possible workaround)
- compilers other than the one inferred by `sysconfig` (i.e: passing the `CC` environment variable at all, even with the same value that `sysconfig.get_config_var('CC')` returns, causes this not to fail)

<details>
<summary>Dockerfile</summary>

``` dockerfile
FROM python:3.9.13-slim-bullseye as base

ENV \
    DEBIAN_FRONTEND noninteractive \
    PYTHONUNBUFFERED 1 \
    LC_ALL=C.UTF-8 \
    LANG=C.UTF-8

RUN apt-get update -y && \
    apt-get install -y --no-install-recommends \
        build-essential \
        gcc

RUN python -m venv /opt/venv
ENV PATH="/opt/venv/bin:$PATH"

RUN pip install -U pip setuptools wheel

RUN pip install uwsgi==2.0.20
```

</details>

<details>
<summary>Output log</summary>

``` plain
× python setup.py bdist_wheel did not run successfully.
│ exit code: 1
╰─> [87 lines of output]
    /opt/venv/lib/python3.9/site-packages/setuptools/_distutils/dist.py:257: UserWarning: Unknown distribution option: 'descriptions'
      warnings.warn(msg)
    running bdist_wheel
    running build
    running build_py
    creating build
    creating build/lib
    copying uwsgidecorators.py -> build/lib
    /opt/venv/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
      warnings.warn(
[...]

PR #2373, opened 2021-11-22
Replace os.system by subprocess.call in `thread_compile`

lukasjuhrich:
Please don't ask me why, but unintrusive one-liner seems to circumvent a nondeterministic bug in building the wheel (see [this comment](https://github.com/agdsn/pycroft/pull/508#issuecomment-975101269)).

It's probably a good idea anyway because the subprocess module is considered to be the replacement of `os.system`.

The relevant log portion looks as follows:
```
[…]
   [thread 1][gcc -pthread] core/regexp.o
  [thread 0][gcc -pthread] core/routing.o
  [thread 1][gcc -pthread] core/yaml.o
  [thread 1][gcc -pthread] core/ssl.o
  [thread 1][gcc -pthread] core/legion.o
  [thread 0][gcc -pthread] core/xmlconf.o
  [thread 0][gcc -pthread] core/dot_h.o
  ----------------------------------------
  ERROR: Failed building wheel for uwsgi
```

…without any further stacktrace or error message. Modifying the lines below to print the return code taught me that `os.system` returned `-1`, but I didn't quite understand the docs concerning whether this meant „terminated by signal“ or „terminated regularly with error code -1“.

PR #2448, opened 2022-06-23
Replace all os.system usage with subprocess.call

PeterJCLaw:
This aims to fix intermittent build failures on some platforms (most notably Debian Bullseye) which appear to be due to the use of `os.system` from background threads. The exact nature of the issue is unclear, however moving to `subprocess.call` appears to fix it.

Fixes https://github.com/unbit/uwsgi/issues/2447.

While there's more that could be done here to modernise these calls, this aims to be a minimal change which introduces subprocess without much intrusion.

This PR goes further than https://github.com/unbit/uwsgi/pull/2373 as a measure of future-proofing the change and as it's unclear to me which of these calls actually could happen in threaded contexts. Since `subprocess` is the preferred mechanism for this in Python anyway this feels like a reasonable set of changes.

### DIFF
--- a/examples/mjpeg_stream.py
+++ b/examples/mjpeg_stream.py
@@ -1,4 +1,5 @@
 import os
+import subprocess
 
 
 def application(env, start_response):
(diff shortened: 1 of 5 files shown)


PeterJCLaw:
@xrmx thanks for merging this. Do you have a sense for when a bugfix release containing this change might be published? I'm not familiar with what this project's release schedule looks like.

PeterJCLaw:
@xrmx is there anything I can help do to get this backported? We're hitting the issue fairly often in CI, so it would be great to have a fix released.

PeterJCLaw:
@xrmx just wondering if there's an update on when this might be included in a release? We're unfortunately starting to hit https://github.com/unbit/uwsgi/issues/2447 quite frequently in CI now. Do let me know if there's anything I can do to help :)

xrmx:
@PeterJCLaw sometimes I do some uwsgi work in my free time sometimes I prefer to spend my free time doing something else. Lately it has been something else :)

Issue #2446, opened 2022-06-19
Build error in master on macOS: `pos` set but not used


corwin-of-amber:
In this function
https://github.com/unbit/uwsgi/blob/ca8bc6acc98b2c062b5b00668acd851e210fcbca/core/utils.c#L3726

a variable `pos` is defined. It is only used on Linux. So in macOS we get
```
      core/utils.c:3726:6: error: variable 'pos' set but not used [-Werror,-Wunused-but-set-variable]
              int pos = 0;
                  ^
```

Because this is an error in `clang` with `-Wall -Werror`.

corwin-of-amber:
PS. Here is a workaround for anyone wishing to install uWSGI via pip (esp. for the purpose of https://github.com/unbit/uwsgi/pull/2363):
```
CFLAGS=-Wno-unused-but-set-variable pip install git+https://github.com/unbit/uwsgi
```

xrmx:
Puzzled because pos is used, sure it can pushed down into a smaller scope but still.

corwin-of-amber:
Hi @xrmx. Notice that the only use of `pos` is inside `#if defined(__linux__) || defined(__FreeBSD__) || defined(__GNU_kFreeBSD__)`. So on macOS, the compiler never sees the use and is correct in issuing the diagnostic.

xrmx:
Then the ifdefery at 3473 should be moved at the top of the function (and the closing at the end) because AFAICT it's not doing anything other than logging messages:

```
#if defined(__linux__) || defined(__FreeBSD__) || defined(__GNU_kFreeBSD__)
```

corwin-of-amber:
Contemplating whether a PR fixing this has any chance of being reviewed.

PR #2455, opened 2022-07-04
Fixed compilation issue on macOS + Clang.

corwin-of-amber:
Some variables remain unused when the #ifdefs for Linux/BSD are not satisfied. This addresses #2446.


xrmx:
Thanks!

Issue #2457, opened 2022-07-11
Got segmentation fault from py_uwsgi_gevent_request


gavin-jeong:
Some environment info.
This happens only on python3 with websocket, not on python2.
It been reproduced on 2.0.20 as well.

```bash
$ python --version
Python 3.8.13
$ uwsgi --version
2.0.19.1
$ uname -srvmpio
Linux 5.13.0-1031-aws #35~20.04.1-Ubuntu SMP Mon Jun 13 22:33:23 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
```

Stack trace
```bash
$ coredumpctl debug
...
(gdb) bt
#0  0x0000ffff9e69f9b0 in ?? () from /lib/aarch64-linux-gnu/libgcc_s.so.1
#1  0x0000ffff9e6a1100 in _Unwind_Backtrace () from /lib/aarch64-linux-gnu/libgcc_s.so.1
#2  0x0000ffffa067ddbc in __GI___backtrace (array=<optimized out>, size=<optimized out>) at backtrace.c:116
#3  0x0000aaaaac05f38c in uwsgi_backtrace ()
#4  0x0000aaaaac05f754 in uwsgi_segfault ()
#5  <signal handler called>
#6  0x0000aaaaac089368 in py_uwsgi_gevent_request ()
#7  0x0000ffffa07dd534 in cfunction_call_varargs (kwargs=<optimized out>, args=0xffff628d97f0, func=0xffff629370e0) at Objects/call.c:758
#8  PyCFunction_Call (func=0xffff629370e0, args=0xffff628d97f0, kwargs=<optimized out>) at Objects/call.c:773
#9  0x0000ffff9dd3cf40 in __Pyx_PyObject_Call (kw=0xffff62921bc0, arg=0xffff628d97f0, func=0xffff629370e0) at src/gevent/greenlet.c:25420
#10 __pyx_pf_6gevent_17_gevent_cgreenlet_8Greenlet_42run (__pyx_v_self=0xffff628c4260) at src/gevent/greenlet.c:14310
#11 __pyx_pw_6gevent_17_gevent_cgreenlet_8Greenlet_43run (__pyx_v_self=0xffff628c4260, unused=<optimized out>) at src/gevent/greenlet.c:14213
[...]

PR #2458, opened 2022-07-11
Fix segfault from GEVENT_SWITCH

gavin-jeong:
Fix https://github.com/unbit/uwsgi/issues/2457


xrmx:
`PyObject_CallObject` and so `python_call` can return NULL so a check looks legit. Thanks!

PR #2459, opened 2022-07-11
PHP: Fix php-app for php81

cuchac:
On php 8.1 when using php-app insteaad of php-index, PHP always return exception about empty file name - caused by empty real_filename_len


xrmx:
Looks legit, thanks!

PR #2464, opened 2022-07-18
make dev version PEP-0440 compliant

roehnan:
The current dev version string is not PEP-0440 compliant and the pip tooling will not allow creating wheels with a non-compliant version. We use wheels and would like to be able to deploy and test development versions.


roehnan:
This isn't strictly required. `pip` has a `--no-verify` option to work around non-compliant names.

xrmx:
Looks good, thanks!

Issue #2480, opened 2022-09-09
Python uWSGI hang during worker restart on recent Python version


PierreF:
We have an hard to reproduce issue with our uWSGI on production that fail to restart. I've found a way to produce similar bug (hopefully the exact same bug) and do some search on the cause

# Our issue on production

Relevant uWSGI options:
```
[uwsgi]
http=0.0.0.0:8080
module=bleemeo.wsgi:application
master=True
processes = 2
threads = 8
max-worker-lifetime = 3600           ; Restart workers after this many seconds
```

The issue happen on Ubuntu 22.04 (amd64) + Python 3.10.4. Sometime when worker lifetime is reached, the worker fail to stop and hang forever. Once the 2nd worker also fail the whole uWSGI stop processing requests.

 This issue does not happen on Ubuntu 20.04 + Python 3.8.10 or Python 3.9.5.

When the hang occur:
* Log say `worker 1 lifetime reached, it was running for 3601 second(s)`
* Log does NOT say `Respawned uWSGI worker 1 (new pid: 5643)`
* The worker only have one thread remaining:
```
$ ps -L wup 23
USER         PID     LWP %CPU NLWP %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
daemon        23      23  0.0    1  3.6 1062012 284748 ?      S    Aug17   0:16 uWSGI worker 1
```
* The stacktrace of the worker is the following:
```
*** backtrace of 23 ***
uWSGI worker 1(uwsgi_backtrace+0x33) [0x55d8f3241953]
uWSGI worker 1(what_i_am_doing+0x2d) [0x55d8f3241a9d]
/usr/lib/x86_64-linux-gnu/libc.so.6(+0x42520) [0x7fc9ae7a5520]
/usr/lib/x86_64-linux-gnu/libc.so.6(+0x91195) [0x7fc9ae7f4195]
/usr/lib/x86_64-linux-gnu/libc.so.6(pthread_cond_wait+0x211) [0x7fc9ae7f6ac1]
/usr/lib/x86_64-linux-gnu/libpython3.10.so.1.0(+0x1c435e) [0x7fc9aeb4f35e]
/usr/lib/x86_64-linux-gnu/libpython3.10.so.1.0(_PyEval_EvalFrameDefault+0xc3e) [0x7fc9ae9fd87e]
/usr/lib/x86_64-linux-gnu/libpython3.10.so.1.0(+0x1c61ff) [0x7fc9aeb511ff]
[...]

Issue #1894, opened 2018-10-14
Using `max-worker-lifetime` kills workers without waiting for pending requests to finish


cipriancraciun:
It seems that if one uses `max-worker-lifetime`, the master process will trigger a `kill` towards the worker process, which will immediately exit without waiting for the pending request to finish.

----

In order to reproduce this configure the following:
````
master = true
workers = 2
threads = 0
enable-threads = false

# Python related configuration

# NOTE:  See the note bellow about this small limit.
max-worker-lifetime = 60
reload-mercy = 30
worker-reload-mercy = 30
reload-on-exception = true
exit-on-reload = true
reaper = true
die-on-term = true
````

Then start in a loop requests that take say 10% of that `max-worker-lifetime` (a simple Python app that just does `time.sleep(6)` would do).  Thus almost one in 10 requests will be reset due to the master killing the worker process.

Regarding the `max-worker-lifetime = 60` limit:  in production I used `3600` (which is quite enough), however in order to reliably trigger this issue I've set this limit small enough.

timdrijvers:
I ran into the same problem with max-requests and offload threads. The thread as still busy generating the request, while the original process got recycled. This abruptly closed the connection, without response.

timdrijvers:
Had a similar issue with still running requests being killed when using max_requests. It only seems to take the last spawn into account, while there can still be running requests.

cipriancraciun:
@timdrijvers Have you found a workaround?

(In my case I just ignored the issue, and restart the whole application once a couple of weeks, as I don't have any "heavy" memory-leaks.)  :(

brandontksmith:
I am experiencing this as well.

lazydaemon:
Same here, see https://stackoverflow.com/questions/58731398/uwsgi-worker-respawning-although-the-request-is-not-yet-finished

PierreF:
I'm also hitting this issue. And I *think* it the cause of the production issue seen in #2480 

For me, what happen is:
* uWSGI start the worker loop here https://github.com/unbit/uwsgi/blob/066b7fdf1bfa12f95652b8bd8cbd3532c8096b91/core/uwsgi.c#L3714
* For all *additional* threads (id > 0) it will start simple_loop_run function: https://github.com/unbit/uwsgi/blob/066b7fdf1bfa12f95652b8bd8cbd3532c8096b91/core/loop.c#L62
[...]

PR #2484, opened 2022-09-14
Wait for other worker thread on shutdown

PierreF:
When worker stop due to max-lifetime, the first thread will now wait for additional thread to complet before terminating the worker process.

This fix the issue described in #1894. With this change I never get any request error nor request that don't finish.

I think this will also fix the issue seen on production described in #2480

If you want that I make this wait behind an option, just ask and I'll update this PR.

### DIFF
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -3718,6 +3718,10 @@ void uwsgi_ignition() {
 		}
 	}
 
+	if (uwsgi.threads > 1 && !uwsgi_instance_is_dying) {
+		wait_for_threads();
+	}
+
 	// end of the process...
 	end_me(0);
 }


PierreF:
Is there something missing for this PR ?

svanoort:
Any chance of getting a review on this one?  It looks like a straightforward 4-liner bugfix to a very painful issue and I think is impacting my company as well, because the symptoms line up quite well (read: prod outages with the upgrade from Python 3.9.7 to 3.9.8 -- with the problem appearing to happen in uWSGI internal code)  

@xrmx ?

SteveByerly:
This was fixed in #2626 which is now commit 06a22597bd419860904fae6f446d8e3b714f5afa

The fix was shipped in [v2.0.25](https://uwsgi-docs.readthedocs.io/en/latest/Changelog-2.0.25.html), though the version's release notes are less explicit about the change than the commits https://github.com/unbit/uwsgi/compare/2.0.24...2.0.25

Issue #2479, opened 2022-09-07
speed up ci testing


xrmx:
tests ci workflows takes more than 15 minutes which is a lot. At the moment we have all the tests run serialized so we can save minutes just by parallelizing them.
We should split tests/travis.sh into a file with the shared variables and then one file for each of the tests: python, deadlocks and rack. And then split the ci workflow into one for each test. Hopefully we can save at least 5 minutes.

PR #2487, opened 2022-09-21
ci: refactor tests to be run in parallel

MRoci:
starting from `tests/travis.sh` subdivide test suites into:
  * `tests/gh-python.sh`
  * `tests/gh-deadlocks.sh`
  * `tests/gh-rack.sh`
  
that share common variables and functions from `tests/gh-shared.sh`

refactor github test workflow to use a multidimensional matrix in order to run each test suite (unittest, python, deadlocks) on each supported python version and rack tests on each supported rack version.

closes https://github.com/unbit/uwsgi/issues/2479

leave `tests/travis.sh` because it can be useful as is somewhere else.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,10 +7,13 @@ on:
     branches: [ master, uwsgi-2.0 ]
 
 jobs:
-  build:
+  python:
 
     runs-on: ubuntu-18.04
-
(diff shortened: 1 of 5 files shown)


xrmx:
It removes so many minutes it can't be wrong. Thanks!

PR #2510, opened 2022-11-17
plugins/python: Use "backslashreplace" on stderr initialization

### LINKED ISSUE #2509: Unicode strings in logging to stderr result in a 500 error code when using the python3 plugin
Hello,

I used the [code from the documentation](https://uwsgi-docs.readthedocs.io/en/latest/WSGIquickstart.html#the-first-wsgi-application) with the slight change that I added some debugging statement with unicode characters.

```
import logging

logging.basicConfig(level=logging.DEBUG)
logger = logging.getLogger(__name__)


    logger.debug("École")
    start_response('200 OK', [('Content-Type', 'text/html')])
    return [b'Hello World']
```

Once this is run with the following command:

```
LANG= uwsgi --plugins python3 --http-socket :9000 --wsgi-file t.py
```

Accessing the endpoint will result in a 500. If `LANG` is set to `C.UTF-8` it won't and the request will succeed.

Of course I can use this workaround but digging in the uwsgi I notice that this change https://github.com/unbit/uwsgi/commit/5bf55220d48c15ec54ef0a01c073771f2a4a4b10 is replacing python's stderr with it's own implementation for performance reasons. The issue is that python creates `stderr` with the error argument set to `"backslashreplace"` (cfr https://github.com/python/cpython/blob/5cfb7d19f5242c9b8ffd2fe30a24569e85a99e1d/Python/pylifecycle.c#L2497)

So I wonder if the replacement of the stdout and stderr shouldn't use this value for `errors` (second to last parameter). I don't think it would interfer with the buffering in any way (I made a crude patch and it works but I am not knowledgeable enough to know it it interfers or not).

nicoe:
Failing to use this value will result in enconding errors when logging unicode characters to stderr

Fix #2509

### DIFF
--- a/plugins/python/python_plugin.c
+++ b/plugins/python/python_plugin.c
@@ -591,7 +591,7 @@ void init_uwsgi_vars() {
 #ifdef HAS_NO_ERRORS_IN_PyFile_FromFd
 		PyObject *new_stdprint = PyFile_FromFd(2, NULL, "w", _IOLBF, NULL, NULL, 0);
 #else
-		PyObject *new_stdprint = PyFile_FromFd(2, NULL, "w", _IOLBF, NULL, NULL, NULL, 0);
+		PyObject *new_stdprint = PyFile_FromFd(2, NULL, "w", _IOLBF, NULL, "backslashreplace", NULL, 0);
 #endif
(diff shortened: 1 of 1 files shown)


xrmx:
Looks legit, thanks!

PR #2536, opened 2023-05-10
Configurable MAX_GATEWAYS, MAX_TIMERS, MAX_CRONS

idlesign:
Hi,

In some cases default max values (namely for `MAX_CRONS`) won't be enough.

This change allows setting more suitable macro values through CFLAGS, e.g:
```
CFLAGS="-DMAX_CRONS=128"
```


idlesign:
@xrmx 
Hi. Can I do anything for this to be merged?

Issue #2486, opened 2022-09-20
Consider porting to PCRE2


musicinmybrain:
The original PCRE library [has just reached end-of-life with release 8.45](https://pcre.org/), and is now unmaintained. Please consider porting to PCRE2.

Fedora Linux has [deprecated](https://www.fedoraproject.org/wiki/PcreDeprecation) the original PCRE library, meaning no new packages may depend on it, and in the medium term the current maintainers would like to retire it, which would affect the [`uwsgi` package](https://src.fedoraproject.org/rpms/uwsgi) in Fedora.

coolo:
Not that it's not just Fedora. Basically every Linux distribution out there has interest to remove PCRE 8.45

xrmx:
Both RHEL 7 and ubuntu 18.04 seems to have pcre2 so hopefully we can avoid maintaining support for both.

diederikdehaas:
Debian has an issue for this too: https://bugs.debian.org/999936

utkonos:
All the libpcre3 packages in Ubuntu have "Old" in the package description now.

https://packages.ubuntu.com/kinetic/libpcre3-dev

dvzrv:
I have just added a [TODO for Arch Linux to move to pcre2](https://archlinux.org/todo/move-to-pcre2/).

@xrmx it would be great if https://github.com/unbit/uwsgi/pull/2543 could be merged and released! :)

xrmx:
This was released in 2.0.25

PR #2543, opened 2023-07-21
add libpcre2 support

niol:
I've started the work. It compiles.

I need to test if it works but wanted to share to avoid somebody else wasting time doing the same.

This would fix https://github.com/unbit/uwsgi/issues/2486

### DIFF
--- a/core/regexp.c
+++ b/core/regexp.c
@@ -1,4 +1,4 @@
-#ifdef UWSGI_PCRE
+#if defined(UWSGI_PCRE) || defined(UWSGI_PCRE2)
 #include "uwsgi.h"
 
 extern struct uwsgi_server uwsgi;
@@ -13,34 +13,64 @@ void uwsgi_opt_pcre_jit(char *opt, char *value, void *foobar) {
 #endif
 }
 
+#ifdef UWSGI_PCRE2
 int uwsgi_regexp_build(char *re, pcre2_code ** pattern) {
 
 	int errnbr;
 	long unsigned int erroff;
 
 	*pattern = pcre2_compile((const unsigned char *) re, PCRE2_ZERO_TERMINATED, 0, &errnbr, &erroff, NULL);
+#else
+int uwsgi_regexp_build(char *re, pcre ** pattern, pcre_extra ** pattern_extra) {
+
+	const char *errstr;
+	int erroff;
+
+	*pattern = pcre_compile((const char *) re, 0, &errstr, &erroff, NULL);
+#endif
 	if (!*pattern) {
+#ifdef UWSGI_PCRE2
 		uwsgi_log("pcre error: code %d at offset %d\n", errnbr, erroff);
+#else
+		uwsgi_log("pcre error: %s at offset %d\n", errstr, erroff);
+#endif
 		return -1;
 	}
 
+#ifdef UWSGI_PCRE2
 	if (uwsgi.pcre_jit) {
 		errnbr = pcre2_jit_compile(*pattern, PCRE2_JIT_COMPLETE);
 		if (errnbr) {
 			uwsgi_log("pcre JIT compile error code %d\n", errnbr);
 			return -1;
 		}
+#else
+	int opt = uwsgi.pcre_jit;
+
+	*pattern_extra = (pcre_extra *) pcre_study((const pcre *) *pattern, opt, &errstr);
+	if (*pattern_extra == NULL && errstr != NULL) {
+		pcre_free(*pattern);
+		uwsgi_log("pcre (study) error: %s\n", errstr);
+		return -1;
+#endif
 	}
 
 	return 0;
 
 }
 
+#ifdef UWSGI_PCRE2
 int uwsgi_regexp_match(pcre2_code *pattern, const char *subject, int length) {
 
 	return pcre2_match(pattern, (const unsigned char *)subject, length, 0, 0, NULL, NULL);
+#else
+int uwsgi_regexp_match(pcre * pattern, pcre_extra * pattern_extra, char *subject, int length) {
+
+	return pcre_exec((const pcre *) pattern, (const pcre_extra *) pattern_extra, subject, length, 0, 0, NULL, 0);
+#endif
 }
 
+#ifdef UWSGI_PCRE2
 int uwsgi_regexp_match_ovec(pcre2_code *pattern, const char *subject, int length, int *ovec, int n) {
 
 	int rc;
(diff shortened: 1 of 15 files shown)


niol:
I've played with the following configuration file and it seems to behave as expected.

```
[uwsgi]
master = False
cheap = True
http-socket = :9090

plugins = python,router_redirect, router_rewrite, router_cache

wsgi-file = helloworld.py

route-user-agent = .*curl.* redirect:http://uwsgi.it
route = ^/test log:someone called /test
route = (.*)\.php$ rewrite:/index.php/$1 log:$1
route = .* addheader:Server: my uWSGI server
route-host = ^localhost$ logvar:local=1
route-uri = ^/foo/(.*)\.jpg\?([0-9]{2})$ log:key=$1.$2.jpg
route-if = equal:${PATH_INFO};/bad break:500 Internal Server Error
```

`helloworld.py`:

```
    return [b"Hello World"]
```

niol:
pcre is not maintained anymore, and being removed in distributions, that's why I did not work on all those `#ifdef`.

xrmx:
But it's still available in the LTS distributions people is using now and uwsgi has been built against.

niol:
pcre2 is also available in LTS distributions. Anyway I'll update this PR with fixes as I get feedback (this just got integrated in Debian). I cannot commit time to update to support both but will hapilly help (answer questions) someone who would.

xrmx:
@niol could you please cherry-pick and rebase on top of https://github.com/xrmx/uwsgi/commit/7835662f76831a76e4cc04791fcf2ee1ea725931 so at least we avoid conflicts?

xrmx:
> pcre2 is also available in LTS distributions.

I know it's available, what I'm saying I don't like to break the build of the majority of uwsgi users that is pip installing uwsgi and not using uwsgi from the distribution.

loqs:
@xrmx is this what you had in mind for retaining pcre1 support https://gist.github.com/loqs/f79a4095875defb294f6617ba06dcd41
I had to add an extra define to differentiate between the two versions of pcre.

xrmx:
@loqs i thought more of two alternative definition (UWSGI_PCRE vs UWSGI_PCRE2) but that's fine too

core/regexp.c
```diff
@@ -66,7 +82,7 @@ char *uwsgi_regexp_apply_ovec(char *src, int src_n, char *dst, int dst_n, int *o
 	int dollar = 0;
 
 	size_t dollars = n;
-	
+
```
loqs, line 85: Unrelated white space change?

core/routing.c
```diff
@@ -1484,38 +1484,27 @@ static int uwsgi_route_condition_regexp(struct wsgi_request *wsgi_req, struct uw
...
 
-	pcre_free(pattern);
-#ifdef PCRE_STUDY_JIT_COMPILE
-	pcre_free_study(pattern_extra);
-#else
-	pcre_free(pattern_extra);
-#endif
-        return 0;
+	pcre2_code_free(pattern);
+	return 0;
```
loqs, line 1507: Accidental formatting change?

core/regexp.c
```diff
@@ -13,49 +13,65 @@ void uwsgi_opt_pcre_jit(char *opt, char *value, void *foobar) {
...
+int uwsgi_regexp_ovector(pcre2_code *pattern) {
 
 	int n;
+	pcre2_match_data *match_data;
 
-	if (pcre_fullinfo((const pcre *) pattern, (const pcre_extra *) pattern_extra, PCRE_INFO_CAPTURECOUNT, &n))
-		return 0;
+	match_data = pcre2_match_data_create_from_pattern(pattern, NULL);
+	n = pcre2_get_ovector_count(match_data);
+	pcre2_match_data_free(match_data);
```
loqs, line 74: Could pcre2_pattern_info not have been used?  Which would have been a smaller change.
```C
       if (pcre2_pattern_info((const pcre2_code *) pattern, PCRE2_INFO_CAPTURECOUNT, &n))
```

loqs (commented):
Please ignore the formatting comments if they are too pedantic.

core/regexp.c
niol: Yep

core/routing.c
niol: Yep

core/regexp.c
niol: You are most probably right, will test and update accordingly.

loqs:
> @loqs i thought more of two alternative definition (UWSGI_PCRE vs UWSGI_PCRE2) but that's fine too

https://gist.github.com/loqs/0aa5f478b0c684c935b0b4f202298aa6 with two alternative definitions.

core/regexp.c
niol: It crashes, this needs more time.

xrmx:
> Any more comments on this?

I'll try to give a review this afternoon

core/regexp.c
```diff
@@ -13,48 +13,111 @@ void uwsgi_opt_pcre_jit(char *opt, char *value, void *foobar) {
...
+	return pcre2_match(pattern, (const unsigned char *)subject, length, 0, 0, NULL, NULL);
+#else
+	return pcre_exec((const pcre *) pattern->p, (const pcre_extra *) pattern->extra, subject, length, 0, 0, NULL, 0);
+#endif
 }
 
-int uwsgi_regexp_match_ovec(pcre * pattern, pcre_extra * pattern_extra, char *subject, int length, int *ovec, int n) {
+int uwsgi_regexp_match_ovec(uwsgi_pcre *pattern, const char *subject, int length, int *ovec, int n) {
 
+#ifdef UWSGI_PCRE2
```
xrmx, line 74: given the two implemetations share only one line of code i think it would be easier to move the pcre2 version into this block altogether

uwsgi.h
```diff
@@ -479,11 +496,10 @@ struct uwsgi_hook {
 	struct uwsgi_hook *next;
 };
 
-#ifdef UWSGI_PCRE
+#if defined(UWSGI_PCRE) || defined(UWSGI_PCRE2)
 struct uwsgi_regexp_list {
 
-	pcre *pattern;
-	pcre_extra *pattern_extra;
+    uwsgi_pcre *pattern;
```
xrmx, line 502: i really hope this won't break anything 😅

BKPepe:
Kind reminder, what needs to be done to have this merged, please?

rockdrilla:
There's one more moment to adjust for pcre2: https://github.com/rockdrilla/sentry-hacks/blob/960d41476665c48ccbe0f87d7a47481ec556476d/patches/uwsgi/1500-libpcre2.patch

matejsp:
I would really like pcre2 support merged in uwsgi. We are currently doing bundling of .so files with auditwheel because pcre is not even installed by default in most if not all latest distros.

```
dnf install patchelf
pip install auditwheel wheel
pip wheel uwsgi-2.0.23.tar.gz
auditwheel repair uWSGI-2.0.23-cp311-cp311-linux_x86_64.whl --plat manylinux_2_34_$(uname -m) --wheel-dir bla --exclude libssl.so.3 --exclude libcrypt.so.2 --exclude libcrypto.so.3
```

RemiCardona:
got bit by this (silently because pip hides the build output by default, it seems) and it took me a long time to figure out why the "Expires" settings for static files was not being set.

So a kind +1 from me. TIA

svenstaro:
@xrmx friendly ping. Could we merge this and get a release out? pcre is going to be dropped sooner or later from all major distros.

xrmx:
Squashed and merged to master in #2624

xrmx:
@niol just noticed that when squashing I erased your authorship, sorry about that. Will fix for 2.0 backport at least.

svenstaro:
@xrmx could we please get a new release including this?

PR #2548, opened 2023-07-25
core/lock: silence glibc warnings against pthread robust mutex functions

xrmx:
Since glibc 2.34 we are gettings warnings that pthread_mutexattr_setrobust_np and pthread_mutex_consistent_np are deprecated.
Problem is that we are checking PTHREAD_MUTEX_ROBUST with the preprocessor but it doesn't work because it's an enum :) So in the end we are using the _np versions of the functions even if the standard ones are available. Since this stuff is implemented on linux libc since 2010-2011 and 2016 in freebsd assume it's here.

### DIFF
--- a/core/lock.c
+++ b/core/lock.c
@@ -91,11 +91,6 @@ retry:
 	}
 
 #ifdef EOWNERDEAD
-#ifndef PTHREAD_MUTEX_ROBUST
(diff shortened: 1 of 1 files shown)


xrmx:
@rdeioris Any chance you can give it a run on macos?

xrmx:
Macos does not use mutexes and illumos has them so I think this is good.

PR #2563, opened 2023-09-17
ci: run compile test on ubuntu 22.04 too

xrmx:


niol:
Maybe use a build matrix for this https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

PR #2578, opened 2023-11-07
properly init cache for purge_lru

niol:
Purging the least recently used cache item relies on a linked list maintained during cache get and set. While uwsgi is running, this works well. But if the uwsgi process is stopped, lru_head and lru_tail are not initialized when loading the existing cache file. As a consequence, least recently used items never get deleted a full cache fails to set any new keys until the cache file is deleted.

This patch properly initializes those variables.


xrmx:
Thanks!

Issue #2283, opened 2021-01-28
PHP 8 plugin build issue


jSML4ThWwBID69YC:
Hello, 

I'm trying to build the uWSGi plugin for PHP 8. I'm using the same steps I do for PHP 7, but with no success. 

OS: FreeBSD 12.2
PHP: PHP 8.0.1 from the ports tree with the following basic options set. 

-  ZTS: Disabled 
-  Embed: Enabled
-  CLI: Enabled

uWSGI version: 2.0.19.1

Build commands:
Extract uWSGI 2.0.19.1.tar.gz to the /root/scripts/uwsgi folder. Then run:

```
cd /root/scripts/uwsgi
make PROFILE=nolang
./uwsgi --build-plugin "plugins/php php"

```

Error:
```
*** uWSGI building and linking plugin from plugins/php ***
[cc -pthread] php_plugin.so
plugins/php/php_plugin.c:77:70: error: expected ')'
static size_t sapi_uwsgi_ub_write(const char *str, size_t str_length TSRMLS_DC)
                                                                     ^
plugins/php/php_plugin.c:77:34: note: to match this '('
static size_t sapi_uwsgi_ub_write(const char *str, size_t str_length TSRMLS_DC)
                                 ^
plugins/php/php_plugin.c:92:70: error: expected ')'
static int sapi_uwsgi_send_headers(sapi_headers_struct *sapi_headers TSRMLS_DC)
                                                                     ^
plugins/php/php_plugin.c:92:35: note: to match this '('
static int sapi_uwsgi_send_headers(sapi_headers_struct *sapi_headers TSRMLS_DC)
                                  ^
plugins/php/php_plugin.c:127:69: error: expected ')'
static size_t sapi_uwsgi_read_post(char *buffer, size_t count_bytes TSRMLS_DC)
                                                                    ^
plugins/php/php_plugin.c:127:35: note: to match this '('
static size_t sapi_uwsgi_read_post(char *buffer, size_t count_bytes TSRMLS_DC)
                                  ^
plugins/php/php_plugin.c:167:66: error: expected ')'
static void sapi_uwsgi_register_variables(zval *track_vars_array TSRMLS_DC)
                                                                 ^
plugins/php/php_plugin.c:167:42: note: to match this '('
static void sapi_uwsgi_register_variables(zval *track_vars_array TSRMLS_DC)
                                         ^
plugins/php/php_plugin.c:171:52: error: expected ')'
[...]

PR #2581, opened 2023-11-13
use libphp on PHP 8+, else libphp7

remicollet:
Since PHP 8 embedded libray don't' have version as suffix

See #2283 

`/usr/bin/ld: cannot find -lphp8: No such file or directory`

This patch is used in Fedora and EPEL for years


xrmx:
Looks good, thanks!

Issue #2580, opened 2023-11-10
harakiri triggered instantly in 2.0.22+ when post-buffering is set to 1 and harakiri > 60


jeromartin:
uwsgi version 2.0.22+
os version: osx, alpine linux, ubuntu
python version 3.11

Following flask application accepting a POST request:

```python
from flask import Flask

app = Flask(__name__)

@app.route("/test", methods=["POST"])
def test():
    return {"data": "value"}
```

And the following simplified configuration:

```
[uwsgi]
master          = true
post-buffering  = 1
harakiri        = 61
harakiri-graceful-timeout = 1
```

launching the wsgi app:

```
uwsgi --http-socket :8080 -w main:app scanner.ini
```

When doing a POST request with 10mb files:

```
curl -X POST -d @data_1mb localhost:8080/test
```

Harakiri is kicking in and killing the worker immediately, preventing the POST to proceed:
```
Fri Nov 10 15:54:01 2023 - *** HARAKIRI ON WORKER 1 (pid: 50158, try: 1, graceful: yes) ***
Fri Nov 10 15:54:01 2023 - HARAKIRI !!! worker 1 status !!!
Fri Nov 10 15:54:01 2023 - HARAKIRI [core 0] 127.0.0.1 - POST /test since 1699628036
Fri Nov 10 15:54:01 2023 - HARAKIRI !!! end of worker 1 status !!!
Fri Nov 10 15:54:01 2023 - HARAKIRI triggered by worker 1 core 0 !!!
DAMN ! worker 1 (pid: 50158) died, killed by signal 9 :( trying respawn ...
Respawned uWSGI worker 1 (new pid: 50174)
```
This is happening only in the following conditions:
- harakiri is set to a value >60 
- post-buffering set to 1
- uploading a big file (1mb is not causing the issue)

The same configuration is working fine in 2.0.21. When enabling the harakiri_graceful_timeout : 
```
[uwsgi]
master          = true
post-buffering  = 1
harakiri        = 61
harakiri-graceful-timeout = 1
```
we can see the following line added to the debugs:
```
Fri Nov 10 15:56:20 2023 - *** HARAKIRI ON WORKER 1 (pid: 50228, try: 1, graceful: yes) ***
Fri Nov 10 15:56:20 2023 - HARAKIRI !!! worker 1 status !!!
Fri Nov 10 15:56:20 2023 - HARAKIRI [core 0] 127.0.0.1 - POST /test since 1699628176
[...]

PR #2587, opened 2023-12-07
use right type for harakiri

sbentov:
see https://github.com/unbit/uwsgi/issues/2580


xrmx:
Nice catch, thanks!

sbentov:
@xrmx any chance of having this in an official 2.0.x release anytime? we would rather not have to run our own version for long

PR #2590, opened 2023-12-16
Rework `uwsgi` module handling

Lalufu:
This partially reverts some of the work done in
18816280a4a91ca39193bba73bc04b047645bbbd.

This removes some code reordering done in the above commit to work around the fact that, starting in 3.12, `PyImport_AppendInittab()` could no longer be called after `Py_Initialize()`.

This patch reverts this reordering, and takes a different approach to add the `uwsgi` python module to the plugin.

This is mainly done to support the https://pypi.org/project/pyuwsgi/ use case where the uwsgi server itself is a python module, and hence the interpreter will always be initialized before uwsgi even starts.

It's also a bit cleaner, there's less code getting rearranged to support 3.12.


plugins/python/python_plugin.c
```diff
@@ -698,10 +695,20 @@ void init_uwsgi_embedded_module() {
 
 
 #ifdef PYTHREE
-#ifndef UWSGI_PY312
+#	ifndef UWSGI_PY312
```
xrmx, line 698: could you keep all the ifdefery at column 0 please?

PR #2597, opened 2024-01-01
Avoid strncpy from null in pyloader

### LINKED ISSUE #2596: Undefined Behavior: `strncpy` from `NULL`
# Steps to reproduce:
0. Compile uwsgi with UBSan enabled:
```
make all CC='clang' CFLAGS='-fsanitize=address,undefined -g -fno-builtin -O0' LDFLAGS='-fsanitize=address,undefined'
```
1. Open uwsgi in gdb:
```
gdb ./uwsgi
```
2. Set a breakpoint for UBSan, then run uwsgi with a Python WSGI file in gdb:
```
>>> b __ubsan_handle_nonnull_arg
>>> r --http=127.0.0.1:8080 --wsgi-file=./examples/welcome.py
```
3. Observe the UBSan crash due to a 0-sized `strncpy` from `NULL`. It is always undefined behavior to execute a `strncpy` from `NULL`, even when `n==0`.

# Suggested patch
Add a check around the `strncpy` call at `plugins/python/pyloader.c:100` to prevent making calls when `n==0`.

kenballus:
Fixes #2596.


xrmx:
Thanks!

PR #2605, opened 2024-02-10
Update glusterfs io callback function signature

Lalufu:
Starting with glusterfs 6.0, the IO callback function takes two additional parameters for stats structs.

Ideally there'd be a way to detect which API version we're building against, but nothing convenient seems to exist.

### DIFF
--- a/plugins/glusterfs/glusterfs.c
+++ b/plugins/glusterfs/glusterfs.c
@@ -46,7 +46,7 @@ struct uwsgi_glusterfs_async_io {
 	ssize_t rlen;
 };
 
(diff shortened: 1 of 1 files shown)


Lalufu:
I'm not aware what the distribution of glusterfs versions across distributions is, but even CentOS7 (which will be end of life in a few months) came with 6.0.

Lalufu:
I'm also pretty sure that this has never worked properly if the library version was 6.0 or higher, which might be an indication on how many people use the glusterfs plugin.

xrmx:
Thanks! 

> I'm also pretty sure that this has never worked properly if the library version was 6.0 or higher, which might be an indication on how many people use the glusterfs plugin.

Indeed

PR #2607, opened 2024-02-14
Fix default values doc for min-worker-lifetime & legion-skew-tolerance

thomas-riccardi:
I reviewed all options that mention `default` in their help string and found these two discrepancies.

sources:
- https://github.com/unbit/uwsgi/blob/ee46b79a4fe31d0a7f5056bf9db3aff01051e064/core/init.c#L140-L141
- https://github.com/unbit/uwsgi/blob/ee46b79a4fe31d0a7f5056bf9db3aff01051e064/core/legion.c#L492-L493


xrmx:
Good catch!

PR #2615, opened 2024-03-08
graceful stop worker when max_requests/realod_on_*

methane:
worker stops when reached max_requests or reload_on_*.

https://github.com/unbit/uwsgi/blob/39f3ade88c88693f643e70ecf6c36f9b375f00a2/core/utils.c#L1216-L1251

`goodbye_cruel_world()` is not graceful. It caused `atexit` not called.
If atexit stops daemon threads, worker won't stop until killed from master.

## Reproduce

```py
# mt_sample.py
import atexit, time, sys
import uwsgi

@atexit.register
def on_exit():
    print(f"on_exit: {uwsgi.worker_id()=}")
    sys.stdout.flush()

def application(env, start_response):
    time.sleep(1)
    start_response('200 OK', [('Content-Type','text/html')])
    return [b"Hello World"]
```

```sh
./uwsgi --wsgi-file=mt_sample.py --master --http-socket=:8000 --workers=1 --threads=80 --max-requests=20 --min-worker-lifetime=10 -L
```

I used [hey](https://github.com/rakyll/hey) to request. But any other load testing tools is OK.

```sh
./hey -c 80 -z 1m  'http://127.0.0.1:8000/'
```

### master branch

```
*** uWSGI is running in multiple interpreter mode ***
spawned uWSGI master process (pid: 93920)
spawned uWSGI worker 1 (pid: 93921, cores: 80)
...The work of process 93921 is done (max requests reached (641 >= 20)). Seeya!
worker 1 killed successfully (pid: 93921)
Respawned uWSGI worker 1 (new pid: 94019)
...The work of process 94019 is done (max requests reached (721 >= 20)). Seeya!
worker 1 killed successfully (pid: 94019)
Respawned uWSGI worker 1 (new pid: 94099)
...The work of process 94099 is done (max requests reached (721 >= 20)). Seeya!
worker 1 killed successfully (pid: 94099)
Respawned uWSGI worker 1 (new pid: 94179)
...The work of process 94179 is done (max requests reached (721 >= 20)). Seeya!
worker 1 killed successfully (pid: 94179)
Respawned uWSGI worker 1 (new pid: 94260)
...The work of process 94260 is done (max requests reached (721 >= 20)). Seeya!
worker 1 killed successfully (pid: 94260)
Respawned uWSGI worker 1 (new pid: 94340)
```

atexit is not called.

### After this patch

```
*** uWSGI is running in multiple interpreter mode ***
spawned uWSGI master process (pid: 94781)
[...]

PR #2619, opened 2024-03-19
stop using pthread_cancel()

methane:
I used much time to reproducer for #2615.

I saw many strange "defunct" (zombie) workers. The master called `waitpid(-1, ...)` but it return 0 even there are some zombies. Finally, master sends KILL signal (MERCY) and worker is restarted.

I believe this strange zombie was born from pthread_cancel. Subthreads calls pthread_cancel() for main thread and it cause strange process.

pthread_cancel() is very hard to use and debug. I can not even attach the strange zombie with `gdb --pid`. I think it is not maintainable.

### DIFF
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -3608,7 +3608,6 @@ void uwsgi_worker_run() {
 
 	if (uwsgi.cores > 1) {
 		uwsgi.workers[uwsgi.mywid].cores[0].thread_id = pthread_self();
-		pthread_mutex_init(&uwsgi.six_feet_under_lock, NULL);
 	}
 
 	uwsgi_ignition();
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -1226,8 +1226,9 @@ void warn_pipe() {
 	}
 }
 
-// in threading mode we need to use the cancel pthread subsystem
-void wait_for_threads() {
+// This function is called from signal handler or main thread to wait worker threads.
+// `uwsgi.workers[uwsgi.mywid].manage_next_request` should be set to 0 to stop worker threads.
+static void wait_for_threads() {
 	int i, ret;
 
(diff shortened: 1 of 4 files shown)


core/uwsgi.c
```diff
@@ -1230,53 +1230,21 @@ void warn_pipe() {
...
-
-		ret = pthread_join(uwsgi.workers[uwsgi.mywid].cores[0].thread_id, NULL);
-		if (ret) {
-			uwsgi_log("pthread_join() = %d on initial thread\n", ret);
-		}
-	}
-
-end:
-
-	pthread_mutex_unlock(&uwsgi.six_feet_under_lock);
```
xrmx, line 1279: Shouldn't we keep this?

core/uwsgi.c
```diff
@@ -1230,53 +1230,21 @@ void warn_pipe() {
...
-			}
-		}
-	}
-
-	if (sudden_death)
-		goto end;
 
 	// wait for thread termination
-	for (i = 1; i < uwsgi.threads; i++) {
+	for (i = 0; i < uwsgi.threads; i++) {
```
xrmx, line 1240: why this?

xrmx: Ah i see

core/uwsgi.c
methane: Nice catch. But I removed `six_feet_under_lock` instead of keep this.

six_feet_under_lock is used only here. And if this function is called from multiple worker threads, the lock doesn't avoid deadklock. (thread A waits thread B, and therad B waits the lock).

When signal handler and main thread call this function simultaneously, six_feet_under_lock doesn't needed.
signal handler just waits pthread_join() for main thread.

xrmx:
Squashed and merged in #2627

Issue #2630, opened 2024-04-14
uWSGI 2.0.25: TypeError: sequence item 14: expected str instance, NoneType found


chickahoona:
Hi,

I am running into an issue with the newest uwsgi (2.0.25). I am pinning things now to 2.0.24 so its not really a big deal for now, yet this of course prevents me from upgrading to newer patches which (especially for seucity updates) will be a problem for me :(

# Steps to reproduce the issue

```
docker run -it python:3.11.3-alpine3.16 /bin/sh
apk add build-base && pip3 install uwsgi
```

# The full error output:

```
Collecting uwsgi
  Using cached uwsgi-2.0.25.tar.gz (811 kB)
  Preparing metadata (setup.py) ... done
Building wheels for collected packages: uwsgi
  Building wheel for uwsgi (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [49 lines of output]
      /usr/local/lib/python3.11/site-packages/setuptools/_distutils/dist.py:264: UserWarning: Unknown distribution option: 'descriptions'
        warnings.warn(msg)
      running bdist_wheel
      running build
      running build_py
      creating build
      creating build/lib
      copying uwsgidecorators.py -> build/lib
      /usr/local/lib/python3.11/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
        warnings.warn(
      installing to build/bdist.linux-x86_64/wheel
      running install
      using profile: buildconf/default.ini
      detected include path: ['/usr/local/include', '/usr/include/fortify', '/usr/include', '/usr/lib/gcc/x86_64-alpine-linux-musl/11.2.1/include']
      Patching "bin_name" to properly install_scripts dir
      detected CPU cores: 16
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/tmp/pip-install-rpmde5o5/uwsgi_bf87333a916945eea9e199e0dca8ae21/setup.py", line 117, in <module>
          setup(
        File "/usr/local/lib/python3.11/site-packages/setuptools/__init__.py", line 87, in setup
          return distutils.core.setup(**attrs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2631, opened 2024-04-15
Ensure every cflag is not None and string type when printing

iKlotho:
Fixes https://github.com/unbit/uwsgi/issues/2630

### DIFF
--- a/uwsgiconfig.py
+++ b/uwsgiconfig.py
@@ -336,8 +336,9 @@ def build_uwsgi(uc, print_only=False, gcll=None):
     if 'APPEND_CFLAGS' in os.environ:
         cflags += os.environ['APPEND_CFLAGS'].split()
 
+    clags = [str(flag) for flag in cflags if flag is not None]
     print("detected CPU cores: %d" % CPUCOUNT)
-    print("configured CFLAGS: %s" % ' '.join([str(flag) for flag in cflags if flag is not None]))
+    print("configured CFLAGS: %s" % ' '.join(cflags))
 
     if sys.version_info[0] >= 3:
         import binascii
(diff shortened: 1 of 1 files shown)


uwsgiconfig.py
```diff
@@ -336,6 +336,7 @@ def build_uwsgi(uc, print_only=False, gcll=None):
     if 'APPEND_CFLAGS' in os.environ:
         cflags += os.environ['APPEND_CFLAGS'].split()
 
+    cflags = [str(flag) for flag in cflags if flag is not None]
```
Wwarrior1, line 339: LGTM

Wwarrior1 (approved):

xrmx:
I think the bug is in the pcre handling and this is just papering over the bug

iKlotho:
I checked the failed build, and the `pcre_cflags` returns None. Checking the value before appending it to the list could fix the problem. 

```py
if pcre_flags:
    self.cflags.append(pcre_cflags)
```

Another fix would be to modify the `append` method of the existing `list` type.  This way we won't need to check every flag if it is `None` but I am not so sure about this approach.

```py
class SafeList(list):
    def append(self, value):
        if value is not None:
            super().append(item)
```

xrmx:
@iKlotho https://github.com/unbit/uwsgi/issues/2630#issuecomment-2055848014

iKlotho:
> @iKlotho see [#2630 (comment)](https://github.com/unbit/uwsgi/issues/2630#issuecomment-2055848014)

Okay, I see the `pcre` is set to `auto` and `pcre_libs` is not defined. I opened another [PR](https://github.com/unbit/uwsgi/pull/2632) but I am not sure if this case should exit the build. I tried to build with the guard and it succeeded.

PR #2632, opened 2024-04-15
Check if the `pcre_libs` exists

iKlotho:
This handles the case where the `required_pcre` is `auto` and the `pcre_libs` is not found.

https://github.com/unbit/uwsgi/issues/2630


xrmx:
Thanks!

PR #2655, opened 2024-07-21
plugins/python: use PyOS_*Fork stable API functions on 3.7+

xrmx:
To avoid calling functions made private from 3.13+. And probably fixing issues with C extensions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


Lalufu:
I've build some packages for the upcoming Fedora 41 with this (which will ship 3.13 and currently has RC versions), and I can confirm that the code builds, and it so far hasn't blown up with the test cases that I have.

As for the patch itself, may I suggest not nesting the `#ifdef` statements for 3.12 and 3.13, but make 3.13 a separate section as the other major versions are? I know this leads to a bit more code duplication, as little changes between the two, but it makes it much easier to see what the code/structures look like for different versions.

xrmx:
@Lalufu thanks for checking, I'll merge this as is and then we can revise it later.

PR #2660, opened 2024-08-19
avoid interleaving pyuwsgi threadstate

### LINKED ISSUE #2659: pyuwsgi + enable-threads + python3.12: workers do not recover after a reload
reproduction:

```bash
python3.12 -m venv venv
. venv/bin/activate
pip install pyuwsgi
bash t.sh
```

```python
# wsgi.py
    print('.', end='', flush=True)

    status = '200 OK'
    output = b'!'

    response_headers = [('Content-type', 'text/plain'),
                        ('Content-Length', str(len(output)))]

    return [output]
```

```bash
# t.sh
UWSGI_MASTER=true \
UWSGI_BINARY_PATH=$PWD/venv/bin/python3 \
UWSGI_VIRTUALENV=$PWD/venv \
UWSGI_WORKERS=2 \
UWSGI_ENABLE_THREADS=true \
UWSGI_MAX_REQUESTS=5 \
UWSGI_HTTP_SOCKET=localhost:9001 \
UWSGI_WSGI_FILE=wsgi.py \
UWSGI_DISABLE_LOGGING=1 \
    venv/bin/uwsgi
```

```bash
# some other tab / window
yes | head -300 | xargs --replace curl localhost:9001/health-check/; echo
```

output:

```console
$ bash t.sh 
*** Starting uWSGI 2.0.26 (64bit) on [Mon Aug 12 16:57:55 2024] ***
compiled with version: Apple LLVM 15.0.0 (clang-1500.3.9.4) on 12 August 2024 17:24:49
os: Darwin-23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000
nodename: FJJ4YYCWYX.attlocal.net
machine: arm64
clock source: unix
pcre jit disabled
detected number of CPU cores: 10
current working directory: /private/tmp/y
detected binary path: /tmp/y/venv/bin/python3
your processes number limit is 5333
your memory page size is 16384 bytes
detected max file descriptor number: 2560
lock engine: OSX spinlocks
uwsgi socket 0 bound to TCP address localhost:9001 fd 3
Python version: 3.12.2 (main, Feb  6 2024, 20:19:44) [Clang 15.0.0 (clang-1500.1.0.2.5)]
--- Python VM already initialized ---
Python main interpreter initialized at 0x10154c4f0
python threads support enabled
your server socket listen backlog is limited to 100 connections
your mercy for graceful operations on workers is 60 seconds
mapped 218688 bytes (213 KB) for 2 cores
*** Operational MODE: preforking ***
WSGI app 0 (mountpoint='') ready in 0 seconds on interpreter 0x10154c4f0 pid: 63841 (default app)
spawned uWSGI master process (pid: 63841)
spawned uWSGI worker 1 (pid: 63842, cores: 1)
spawned uWSGI worker 2 (pid: 63843, cores: 1)
................................................................................................................................................................................................................................................................................................................................................................................................................................................The work of process 63843 is done. Seeya!
....The work of process 63842 is done. Seeya!
worker 2 killed successfully (pid: 63843)
Respawned uWSGI worker 2 (new pid: 64286)
worker 1 killed successfully (pid: 63842)
Respawned uWSGI worker 1 (new pid: 64287)
```

(and then no additional output) -- the uwsgi processes also become relatively unkillable (not responding to `^C` / `^\`, `SIGTERM` etc.)

the other terminal output:

```console
$ yes | head -300 | xargs --replace curl localhost:9001/health-check/; echo
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
$ yes | head -300 | xargs --replace curl localhost:9001/health-check/; echo
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

this seems to not be an issue with python 3.11 or without `pyuwsgi`

asottile:
resolves #2659

so there's quite a lot going on here so I'll try to explain the change

- in all versions of pyuwsgi at the moment the *first* fork has a `NULL` threadstate 
    - due to `uwsgi_python_master_fixup` which calls `UWSGI_RELEASE_GIL` (expanded to `PyEval_SaveThread` -- which drops the GIL and sets threadstate to `NULL`)
    - this is called during `uwsgi_setup`
    - after `uwsgi_setup` was returning, `PyThreadState_Swap` was restoring the pyuwsgi threadstate (in both the original and worker processes)
- future forks would have the pyuwsgi threadstate active (from the restoration at `PyThreadState_Swap`)
    - in python versions < 3.12 this wasn't an issue (for a reason I'm not completely understanding -- but my notes are below this)
    - in 3.12+ the `PyEval_RestoreThread` would attempt to `take_gil` and then block forever on the GIL mutex (despite it actually holding it?  due to the fork state from the parent process)
    - bisecting cpython showed that https://github.com/python/cpython/commit/92d8bfffbf377e91d8b92666525cb8700bb1d5e8 slightly changed behaviour of `PyThreadState_Swap` (it now additionally manages GIL state: unlocking the previous threadstate and locking the new threadstate)
    - putting a log line in the `PyThreadState_Swap` showed a suspicious `swapping from oldts=123123 to newts=123123` (swapping from its own threadstate to itself?)
    - this is because after forking control would be given back to the original threadstate (which mostly worked but was in UB territory given the GIL state)
- in the new version of the code, once we call `uwsgi_setup` we never give control back to the original `pyuwsgi` threadstate avoiding the `Swap` dance entirely!

### DIFF
--- a/plugins/pyuwsgi/pyuwsgi.c
+++ b/plugins/pyuwsgi/pyuwsgi.c
@@ -115,12 +115,6 @@ pyuwsgi_setup(PyObject *self, PyObject *args, PyObject *kwds)
         return NULL;
     }
 
-    //TODO: ...???
-    // actually do the thing!
-    PyThreadState *_tstate = PyThreadState_Get();
-    uwsgi_setup(orig_argc, orig_argv, environ);
-    PyThreadState_Swap(_tstate);
-
     Py_INCREF(self);
     return self;
 }
@@ -133,6 +127,7 @@ pyuwsgi_init(PyObject *self, PyObject *args, PyObject *kwds)
         return NULL;
     }
 
+    uwsgi_setup(orig_argc, orig_argv, environ);
     int rc = uwsgi_run();
 
     // never(?) here
@@ -149,6 +144,7 @@ pyuwsgi_run(PyObject *self, PyObject *args, PyObject *kwds)
         return NULL;
     }
 
+    uwsgi_setup(orig_argc, orig_argv, environ);
     int rc = uwsgi_run();
 
     // never(?) here


asottile:
I did end up figuring out why 3.12's subtle behaviour change caused the deadlock as well

- in 3.11 the threadstate that was restored after the `PyThreadState_Swap` did *not* have the GIL locked (technically this _could_ have allowed a data race if threads existed before starting uwsgi via pyuwsgi -- however I think this is exceedingly unlikely)
- in 3.12 since `PyThreadState_Swap` was changed to release the old threadstate's GIL and acquire the GIL in the new threadstate this meant that the saved threadstate had `->locked = 1` (which is sort of an invalid state?)
    - as far as I can tell there aren't any public apis to *undo* this and "restore" the 3.11 behaviour precisely -- though I think my patch is still an improvement on that idea even if it were possible
- then later it would try and lock (despite already being `-> locked = 1`) and deadlock against itself
    - this is actually called out on the [docs](https://docs.python.org/3/c-api/init.html#c.PyEval_RestoreThread):
      > If the lock has been created, the current thread must not have acquired it, otherwise deadlock ensues.

xrmx:
Thanks, added the explanation to the commit message in #2670

Issue #2270, opened 2021-01-03
pip3 install fails on macOS Big Sur with M1


jordanmclain:
I don't know if this is an M1 issue, as it compiles with homebrew.  Looks to be a path error with pip3 install...

Relevant error line:

clang: error: no such file or directory: '/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/config-3.8/libpython3.8.a'

Correct path:

/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/config-3.8-darwin/libpython3.8.a

Full error message:

Collecting uwsgi
  Downloading uWSGI-2.0.19.1.tar.gz (803 kB)
     |████████████████████████████████| 803 kB 18.7 MB/s 
Building wheels for collected packages: uwsgi
  Building wheel for uwsgi (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: /Library/Developer/CommandLineTools/usr/bin/python3 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/tmp/pip-install-9m0dhho2/uwsgi_bd9181bf477e43c0aba1c101bae5d6f5/setup.py'"'"'; __file__='"'"'/private/tmp/pip-install-9m0dhho2/uwsgi_bd9181bf477e43c0aba1c101bae5d6f5/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /private/tmp/pip-wheel-0wfln_r7
       cwd: /private/tmp/pip-install-9m0dhho2/uwsgi_bd9181bf477e43c0aba1c101bae5d6f5/
  Complete output (153 lines):
  /Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/distutils/dist.py:274: UserWarning: Unknown distribution option: 'descriptions'
    warnings.warn(msg)
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib
  copying uwsgidecorators.py -> build/lib
  installing to build/bdist.macosx-10.14.6-arm64/wheel
  running install
  using profile: buildconf/default.ini
[...]

Issue #2109, opened 2019-12-27
uwsgi on mac catalina (10.15).


swuecho:
after upgrade to mac 10.15.2 (19C57). uwsgi fails. I reinstall using pip, but still no luck.  There is no error  msg, just do not work any more. 

I wonder if uwsgi works on catalina or not.

sampullman:
I'm not sure if this is related, but I was unable to install uwsgi on OSX 10.15.3, the error was:
```
clang: error: no such file or directory: '<venv-dir>/lib/python3.7/config-3.7m/libpython3.7.a\'
*** error linking uWSGI ***
```
My temporary workaround is:
```
ln -s <venv-dir>/lib/python3.7/config-3.7m-darwin <venv-dir>/lib/python3.7/config-3.7m
```

swuecho:
thanks @sampullman. I migrate to use uwsgi in docker after hitting the wall too many times.

richardcrichardc:
I have the same issue on Mac OS X Catalina:

```
(flask-hello) bash-3.2$ pip install uwsgi
Collecting uwsgi
  Using cached uwsgi-2.0.18.tar.gz (801 kB)
Building wheels for collected packages: uwsgi
...
...
...
    *** uWSGI linking ***
[...]

PR #2683, opened 2024-10-05
Fix `pip install` static library not found errors for conda

shaform:
Under certain circumstances `Py_ENABLE_SHARED` would be `False` but the static library cannot be found. For example, if you use conda environments or if you use specific versions of MacOS #2109, #2270. `pip install` would fail in such a situation.

In this PR, we adjust the configurations so that the installation script would attempt to use shared library instead of the static one if no static library is found. This fixes `pip install` for `Python 3.10.15 | packaged by conda-forge`. It's not clear if this would fix the MacOS problems mentioned in the above issues though, as I am unable to verify those problems.


xrmx:
Thanks!

PR #2685, opened 2024-10-09
fix: Bit more friendly log messages

de1987:
These messages were fixed in Oct 2019 on the master branch and it's not present in the  2.0 branch.


elyezer:
@unbit any idea when we would be able to get these changes in?

elyezer:
Thank you so much!

PR #2686, opened 2024-10-09
fix: Bit more friendly log messages 

de1987:
These messages were fixed in Oct 2019 on the master branch and it's not present in the 3.0 branch.


xrmx:
No need to update uwsgi-3.0 branch

Issue #2656, opened 2024-07-22
gracefully_kill_them_all is invalid.


junhao69535:
python version: 3.10.12
uwsgi version: 2.0.23

uwsgi.ini
```
wsgi = app:app
http = :8000
master = true
processes = 2
harakiri = 15
harakiri-verbose = true
harakiri-graceful-timeout = 15
harakiri-graceful-signal = 15
max-requests = 100000
memory-report = true
enable-threads = true
threads = 4
enable-thread = true
showconfig = true
listen = 1024
post-buffering = 8192
buffer-size = 32768
lazy = true
http-keepalive = 1
add-header = Connection: Keep-Alive
http-timeout = 70
socket-timeout = 75
hook-master-start = unix_signal:15 gracefully_kill_them_all
vacuum = true
```

app.py
```
import time
from flask import Flask

app = Flask(__name__)

@app.route("/")
def home():
    time.sleep(10)
    return "OK"
```

kill -s 15 master-pid while request is imcompleted.fllowing is uwsgi log:

```
[2024-07-22 09:13:16] - compiled with version: 11.4.0 on 22 July 2024 09:12:33                                                                
[2024-07-22 09:13:16] - os: Linux-5.15.0-113-generic #123-Ubuntu SMP Mon Jun 10 08:16:17 UTC 2024                                             
[2024-07-22 09:13:16] - nodename: zhbubuntu                                                                                                   
[2024-07-22 09:13:16] - machine: x86_64                                                                                                       
[2024-07-22 09:13:16] - clock source: unix                                                                                                    
[2024-07-22 09:13:16] - pcre jit disabled                                                                                                     
[2024-07-22 09:13:16] - detected number of CPU cores: 2                                                                                       
[2024-07-22 09:13:16] - current working directory: /home/zhb/flask                                                                            
[2024-07-22 09:13:16] - detected binary path: /home/zhb/venv3/bin/uwsgi                                                                       
[2024-07-22 09:13:16] - your processes number limit is 15191                                                                                  
[2024-07-22 09:13:16] - your memory page size is 4096 bytes                                                                                   
[2024-07-22 09:13:16] - detected max file descriptor number: 1024                                                                             
[2024-07-22 09:13:16] - lock engine: pthread robust mutexes                                                                                   
[2024-07-22 09:13:16] - thunder lock: disabled (you can enable it with --thunder-lock)                                                        
[2024-07-22 09:13:16] - uWSGI http bound on :8000 fd 4                                                                                        
[...]

PR #2689, opened 2024-10-22
Fix reload-on-touch/change

### LINKED ISSUE #2681: 20.0.27 attach-daemon not responding to touching .ini file
<= 20.0.26 we've never had an issue with a couple of attach-daemon entries in our .ini files.
Now with 20.0.27 touching the file doesn't restart the daemons. We've changed nothing.

**UPDATE: Confirmed, touching an ini that used to restart python app and attach-daemon's no longer works on .27. We downgraded to .26 no problems at all.**

Of course, restarting uWSGI works just fine.

Lalufu:
As part of a fix for #2656, commit 8f1d0e5 introduced a measure to wait for processes to finish work while reloading. The code to do this affects other logic flows was well, breaking the ability to reload uwsgi when a config file changes.

This moves the code around a bit so #2656 stays fixed, but reload-on-touch/change works again.

Fixes #2681.

### DIFF
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -1356,7 +1356,7 @@ void kill_them_all(int signum) {
 // gracefully destroy
 void gracefully_kill_them_all(int signum) {
 
-	    int waitpid_status;
+        int waitpid_status;
 
         if (uwsgi_instance_is_dying) return;
         uwsgi.status.gracefully_destroying = 1;
@@ -1379,12 +1379,12 @@ void gracefully_kill_them_all(int signum) {
                 }
         }
 
-
         for (i = 1; i <= uwsgi.numproc; i++) {
             if (uwsgi.workers[i].pid > 0) {
                 waitpid(uwsgi.workers[i].pid, &waitpid_status, 0);
             }
         }
+
         uwsgi_destroy_processes();
 }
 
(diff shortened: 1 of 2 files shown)


core/uwsgi.c
```diff
@@ -1356,6 +1356,8 @@ void kill_them_all(int signum) {
 // gracefully destroy
 void gracefully_kill_them_all(int signum) {
 
+	    int waitpid_status;
```
xrmx, line 1359:
```suggestion
        int waitpid_status;
```

core/uwsgi.c
```diff
@@ -1377,6 +1379,12 @@ void gracefully_kill_them_all(int signum) {
                 }
         }
 
+
```
xrmx, line 1382:
```suggestion
```

core/uwsgi.c
```diff
@@ -1377,6 +1379,12 @@ void gracefully_kill_them_all(int signum) {
...
+        for (i = 1; i <= uwsgi.numproc; i++) {
+            if (uwsgi.workers[i].pid > 0) {
+                waitpid(uwsgi.workers[i].pid, &waitpid_status, 0);
+            }
+        }
```
xrmx, line 1387:
```suggestion
        }

```

xrmx:
Thanks a lot!

karolyi:
Thank you very much for this! This change has fixed my spoolers not being killed (staying in defunct state), while trying to do a touch-reload. It resulted in my entire uwsgi stack hanging on reload.

PR #2695, opened 2024-11-15
fix for #1143

### LINKED ISSUE #1143: Setting reload-os-env to True causes TypeError on python3.4
If I have reload-os-env = true in my uwsgi.conf, starting uwsgi will result in

```
  File "/srv/knip/development/KnipKnap/lib/python3.4/os.py", line 637, in __setitem__
    key = self.encodekey(key)
  File "/srv/knip/development/KnipKnap/lib/python3.4/os.py", line 706, in encode
    raise TypeError("str expected, not %s" % type(value).__name__)
TypeError: str expected, not bytes
```

Apparently, plugins/python/pyloader.c 

https://github.com/unbit/uwsgi/blob/master/plugins/python/pyloader.c#L147

passes a bytestring in stead of a normal string (I don't know enough Python C api to say anything more sensible than this :)
## reproduce:

```
$ virtualenv -p /usr/bin/python3.4 .; bin/pip install uwsgi==2.0.12
...
$ bin/uwsgi -s s --reload-os-env --wsgi-file=anything
....
  File "/srv/knip/development/KnipKnap/lib/python3.4/os.py", line 637, in __setitem__
    key = self.encodekey(key)
  File "/srv/knip/development/KnipKnap/lib/python3.4/os.py", line 706, in encode
    raise TypeError("str expected, not %s" % type(value).__name__)
TypeError: str expected, not bytes
(repeated)
```

wszak:
replace strings with unicode for python3
fixes #1143

### DIFF
--- a/plugins/python/pyloader.c
+++ b/plugins/python/pyloader.c
@@ -129,21 +129,33 @@ int init_uwsgi_app(int loader, void *arg1, struct wsgi_request *wsgi_req, PyThre
                         		p = strchr(*e, '=');
                         		if (p == NULL) continue;
 
+#ifdef PYTHREE
+					k = PyUnicode_FromStringAndSize(*e, (int)(p-*e));
+#else
 					k = PyString_FromStringAndSize(*e, (int)(p-*e));
+#endif
 					if (k == NULL) {
                                 		PyErr_Print();
                                 		continue;
 					}
 
-                        		env_value = PyString_FromString(p+1);
+#ifdef PYTHREE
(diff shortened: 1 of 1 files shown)


xrmx (commented):
Patch looks fine but could you please add the PYTHREE guards instead?

```
+#ifdef PYTHREE
+                                       k = PyUnicode_FromStringAndSize(*e, (int)(p-*e));
+#else
                                        k = PyString_FromStringAndSize(*e, (int)(p-*e));
+#endif
```

PR #2703, opened 2025-01-30
Fix memory leak in py_uwsgi_spooler_pids

henyu6:
When we call PyInt_FromLong, the reference count is increased and PyList_Append increasing the reference count again because when using PyList_Append, the reference count is incremented and not taken. 

The change made is to set the result of PyInt_FromLong, append it to the list, then decrement the reference count.


methane (approved):

plugins/python/uwsgi_pymodule.c
```diff
@@ -2278,7 +2278,9 @@ PyObject *py_uwsgi_spooler_pids(PyObject * self, PyObject * args) {
     PyObject *ret = PyList_New(0);
     struct uwsgi_spooler *uspool = uwsgi.spoolers;
     while (uspool) {
-        PyList_Append(ret, PyInt_FromLong(uspool->pid));
+        PyObject *spool_pid = PyInt_FromLong(uspool->pid)
```
luchr, line 2281: There is a semicolon ";" missing at the end.

PR #2708, opened 2025-02-18
fix build with gcc-15 and -Wincompatible-pointer-types

niol:
see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1098052
    https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1098054


core/fifo.c
```diff
@@ -78,25 +78,25 @@ void uwsgi_master_fifo_prepare() {
 	uwsgi_fifo_table['8'] = uwsgi_fifo_set_slot_eight;
 	uwsgi_fifo_table['9'] = uwsgi_fifo_set_slot_nine;
 
-	uwsgi_fifo_table['-'] = uwsgi_cheaper_decrease;
-	uwsgi_fifo_table['+'] = uwsgi_cheaper_increase;
-	uwsgi_fifo_table['B'] = vassal_sos; 
-	uwsgi_fifo_table['c'] = uwsgi_chain_reload;
-	uwsgi_fifo_table['C'] = uwsgi_go_cheap;
+	uwsgi_fifo_table['-'] = (void (*)(int))uwsgi_cheaper_decrease;
```
xrmx, line 81: These changes are not in master, could you please create a PR against that first so we can apply the bakcport to 2.0 later? thanks

niol: Proposed #2713 against master.

PR #2713, opened 2025-03-30
fix build with gcc-15

niol:
see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1098052
    https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1098054

### DIFF
--- a/core/init.c
+++ b/core/init.c
@@ -3,7 +3,7 @@
 extern struct uwsgi_server uwsgi;
 
 struct http_status_codes {
-        const char      key[3];
+        const char      key[4];
         const char      *message;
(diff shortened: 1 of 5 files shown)


core/init.c
```diff
@@ -3,7 +3,7 @@
 extern struct uwsgi_server uwsgi;
 
 struct http_status_codes {
-        const char      key[3];
+        const char      key[4];
```
xrmx, line 6: I don't see this from the issues mentioned above, is it for something else?

niol: I made an specific commit for this.

PR #2734, opened 2025-08-18
Some minor build fixes

niol:
Here are some minor build fixes we've been using in Debian for years.


niol:
Sorry, removed that commit.

Issue #2710, opened 2025-03-21
uWSGI fails to start with "--py-call-osafterfork" on Python 3.12 (work with Python 3.12)


TheBlusky:
Using uwsgi 2.0.28, simply running the `uwsgi --master --http=8080 --py-call-osafterfork` command on an empty environment.

- With python 3.13

```
$ docker run --rm -it python:3.13 bash
root@a2a21e201c85:/# pip install uwsgi
[...]
Successfully installed uwsgi-2.0.28
root@a2a21e201c85:/# uwsgi --master --http=8080 --py-call-osafterfork
*** Starting uWSGI 2.0.28 (64bit) on [Fri Mar 21 16:40:33 2025] ***
[...]
*** uWSGI is running in multiple interpreter mode ***
spawned uWSGI master process (pid: 609)
spawned uWSGI worker 1 (pid: 610, cores: 1)
Fatal Python error: PyMutex_Unlock: unlocking mutex that is not locked
Python runtime state: initialized

Current thread 0x00007ff7a3c42480 (most recent call first):
spawned uWSGI http 1 (pid: 611)
  <no Python frame>
DAMN ! worker 1 (pid: 610) died, killed by signal 6 :( trying respawn ...
Respawned uWSGI worker 1 (new pid: 612)
Fatal Python error: PyMutex_Unlock: unlocking mutex that is not locked
Python runtime state: initialized

Current thread 0x00007ff7a3c42480 (most recent call first):
  <no Python frame>
DAMN ! worker 1 (pid: 612) died, killed by signal 6 :( trying respawn ...
worker respawning too fast !!! i have to sleep a bit (2 seconds)...
^CSIGINT/SIGTERM received...killing workers...
gateway "uWSGI http 1" has been buried (pid: 611)
Respawned uWSGI worker 1 (new pid: 613)
Fatal Python error: PyMutex_Unlock: unlocking mutex that is not locked
Python runtime state: initialized
```

The module is looping on the last error: `Fatal Python error: PyMutex_Unlock: unlocking mutex that is not locked`

- With python 3.12

```
$ docker run --rm -it python:3.12 bash
root@7af8ba226d73:/# pip install uwsgi
[...]
root@7af8ba226d73:/# uwsgi --master --http=8080 --py-call-osafterfork
*** Starting uWSGI 2.0.28 (64bit) on [Fri Mar 21 16:38:51 2025] ***
[...]
spawned uWSGI master process (pid: 607)
[...]

PR #2736, opened 2025-09-05
py-call-osafterfork isn't needed in python 3.13+

duzumaki:
fixes https://github.com/unbit/uwsgi/issues/2710

in 3.13, there's a new [mutex type](https://docs.python.org/3/c-api/init.html#c.PyMutex) that now cares if the mutex is unlocked while not currently locked

A warning should be emitted to advise to use `py-call-uwsgi-fork-hooks` if running against 3.13 (+ a doc update that suggests to use `py-call-uwsgi-fork-hooks` over `py-call-osafterfork`

since maintainers want to support py-call-osafterfork for LTS reasons, my suggestion would be to just remove `py-call-osafterfork` entirely otherwise

### DIFF
--- a/plugins/python/python_plugin.c
+++ b/plugins/python/python_plugin.c
@@ -460,13 +460,18 @@ void uwsgi_python_post_fork() {
 
 	// reset python signal flags so child processes can trap signals
 	// Necessary if uwsgi fork hooks not called to update interpreter state
-	if (!up.call_uwsgi_fork_hooks && up.call_osafterfork) {
-#ifdef HAS_NOT_PYOS_FORK_STABLE_API
-		PyOS_AfterFork();
-#else
-                PyOS_AfterFork_Child();
-#endif
-	}
+    #if PY_VERSION_HEX < 0x030D0000
+		if (!up.call_uwsgi_fork_hooks && up.call_osafterfork) {
+			// before python 3.7
+			#ifdef HAS_NOT_PYOS_FORK_STABLE_API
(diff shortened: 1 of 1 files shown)


plugins/python/python_plugin.c
```diff
@@ -460,13 +460,18 @@ void uwsgi_python_post_fork() {
...
 	// reset python signal flags so child processes can trap signals
 	// Necessary if uwsgi fork hooks not called to update interpreter state
-	if (!up.call_uwsgi_fork_hooks && up.call_osafterfork) {
-#ifdef HAS_NOT_PYOS_FORK_STABLE_API
-		PyOS_AfterFork();
-#else
-                PyOS_AfterFork_Child();
-#endif
-	}
+    #if PY_VERSION_HEX < 0x030D0000
```
xrmx, line 463: Please move preprocessor directives to column 0, also you can use `#ifdef UWSGI_PY313`

plugins/python/python_plugin.c
```diff
@@ -460,13 +460,18 @@ void uwsgi_python_post_fork() {
...
+		if (!up.call_uwsgi_fork_hooks && up.call_osafterfork) {
+			// before python 3.7
+			#ifdef HAS_NOT_PYOS_FORK_STABLE_API
+					PyOS_AfterFork();
```
xrmx, line 467: looks like there's one tab too much

shadchin:
We also encountered this issue when updating Python to version 3.13.

PR #2737, opened 2025-09-08
Fix failure to build with -Wincompatible-pointer-type

niol:


niol:
Same as https://github.com/unbit/uwsgi/pull/2733, closing

Issue #2722, opened 2025-04-29
Old master stuck on graceful shutdown (2q) in master-fifo reload since uWSGI 2.0.27+


harmant:
In versions of uWSGI from **2.0.27 and above**, the **fork-master graceful reload approach using `master-fifo` and `2q` (as described in #1617)** is **broken**.

After sending `f` (to fork a new master), and the newly forked master sending `2q` to shut down the old one:

- **Workers of the old master exit cleanly** (verified via `ps`),
- But the **old master process hangs indefinitely**,
- Logs every 20 seconds (based on the `reload-mercy` option) repeatedly show:
```
worker 1 (pid: XXXX) is taking too much time to die...NO MERCY !!!
```

Despite the fact that the worker PID no longer exists.

### Config (minimal reproduction)

```ini
master = true
no-orphans = true
reload-mercy = 20
enable-threads = true

master-fifo = /tmp/new-app.fifo
master-fifo = /tmp/running-app.fifo
master-fifo = /tmp/quit-app.fifo

if-exists = /tmp/running-app.fifo
    hook-accepting1-once = writefifo:/tmp/running-app.fifo 2q
endif =
hook-accepting1-once = spinningfifo:/tmp/new-app.fifo 1P
```

Reload triggered via:

```bash
echo f > /tmp/running-app.fifo
```

This works correctly up to **2.0.26**, but breaks on **2.0.27 and later**.

### Observed behavior

- `2q` is received correctly.
- The old master enters the graceful shutdown path.
- All workers exit (`ps` confirms this).
- But the old master logs `NO MERCY` messages and never exits.

### Expected behavior

- After all workers terminate, the old master should exit cleanly.
- No lingering zombie master process should remain.

### Possibly related commits

From the changelog for 2.0.27:

> “Fix `gracefully_kill_them_all` with running requests” (#2656)

That change appears to have **introduced this regression**, possibly due to incorrectly handling the `worker_table` state or failing to detect that all workers have already exited.

Our production deployments rely on **FIFO-based zero-downtime reloads**, and we are currently unable to upgrade to newer uWSGI versions due to this issue.

Please consider reviewing the changes introduced in 2.0.27 related to worker reap logic in the old master process after forking the new one.

**OS**: Debian 11 (oldstable)  
**Python**: 3.9.2 (system-wide)  
**App**: Django 4.2.20

wynnw:
I can confirm the same behavior that's impacted my deployment workflows - had to revert back to an older release.

[...]

PR #2740, opened 2025-09-23
Fix for gracefully_kill_them_all() which broke setups using fifo-base…

wynnw:
This is a fix for https://github.com/unbit/uwsgi/issues/2722 which is a critical bug blocking existing users from upgrading. It keeps the same logic as the original fix in #2656 (included in the 2.0.27 release) by blocking the gracefully_kill_them_all() function from finishing until all the child workers exit. BUT, it uses the WNOHANG approach which seems to allow existing signal handler logic to do the needed cleanup for the old master process to exit cleanly.

Please consider this - thanks!

### DIFF
--- a/core/uwsgi.c
+++ b/core/uwsgi.c
@@ -1319,38 +1319,57 @@ void kill_them_all(int signum) {
 
 // gracefully destroy
 void gracefully_kill_them_all(int signum) {
+	if (uwsgi_instance_is_dying) return;
+	uwsgi.status.gracefully_destroying = 1;
 
-        int waitpid_status;
+	// unsubscribe if needed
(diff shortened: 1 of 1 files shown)


core/uwsgi.c
```diff
@@ -1319,38 +1319,57 @@ void kill_them_all(int signum) {
 
 // gracefully destroy
 void gracefully_kill_them_all(int signum) {
+  if (uwsgi_instance_is_dying) return;
```
xrmx: Could you please convert to tabs or keep the previous 8 spaces indentation?

wynnw: Done - changed the whole function to use tabs.

xrmx:
Thanks!

PR #2746, opened 2025-11-03
fix: use `_exit()` instead of `exit()` from `uwsgi_exit()`

taegyunkim:
When a native Python extension module uses thread-local storage and an `atexit` hook using that TLS is installed, `uwsgi` results in a segmentation fault. 

A minimal repro code is available at https://github.com/taegyunkim/uwsgi-tls-segfault


taegyunkim:
Do you mind taking a look @xrmx?

KowalskiThomas (approved):

taegyunkim:
Gentle reminder @xrmx

PR #2763, opened 2026-05-06
[SAST] Fix multiple handle leaks

### LINKED ISSUE #2762: [SAST] Several potential handle leaks
Running static analysis on UWSGi identified several potential handle leaks that I believe are confirmed. They are quite minor, but still could be fixed.

The findings are:

1. Handle leak in `create_server_socket`
**File:** /core/socket.c

**Problem:**
Socket desciptor `serverfd` is acquired on line 82:
https://github.com/unbit/uwsgi/blob/8d116f7ea2b098c11ce54d0b3a561c54dcd11929/core/socket.c#L82

On certain error paths (e.g. lines 94-98) `uwsgi_nuclear_blast` is called:
https://github.com/unbit/uwsgi/blob/8d116f7ea2b098c11ce54d0b3a561c54dcd11929/core/socket.c#L94-L98
Which terminates the process with `exit` in most cases, however, it may simply return in some scenarios (e.g. when running as Emperor):
https://github.com/unbit/uwsgi/blob/8d116f7ea2b098c11ce54d0b3a561c54dcd11929/core/uwsgi.c#L1455-L1458

If the process survives, the function returns without closing `serverfd`, resulting in a handle leak.

**Suggested patch:**
Close `serverfd` before returning if the process survived nuclear blast:
```
if (setsockopt(serverfd, SOL_SOCKET, SO_REUSEADDR, (const void *) &reuse, sizeof(int)) < 0) {
	uwsgi_error("SO_REUSEADDR setsockopt()");
	uwsgi_nuclear_blast();
	close(serverfd);
	return -1;
}
```

2. Handle leak in `uwsgi_load_plugin`
**File:** /core/plugins.c

**Problem:**
Plugin handle `plugin_handle` is acquired via `dlopen` (e.g. on line 157):
https://github.com/unbit/uwsgi/blob/8d116f7ea2b098c11ce54d0b3a561c54dcd11929/core/plugins.c#L157

Then, under the `success` label, the first `dlsym` call on line 175 may fail:
https://github.com/unbit/uwsgi/blob/8d116f7ea2b098c11ce54d0b3a561c54dcd11929/core/plugins.c#L175

And the subsequent `dlsym` call on line 197 may also fail:
https://github.com/unbit/uwsgi/blob/8d116f7ea2b098c11ce54d0b3a561c54dcd11929/core/plugins.c#L197
(Seems like an unlikely scenario, but I believe it is still possible if the user provided a regular .so file instead of a valid uwsgi plugin. Please correct me if I'm wrong though).

If both `dlsym` calls fail, the function reaches the `end` block at lines 268–275 and returns without closing `plugin_handle`:
https://github.com/unbit/uwsgi/blob/8d116f7ea2b098c11ce54d0b3a561c54dcd11929/core/plugins.c#L268-L275
This results in a handle leak.

**Suggested patch:**
Close `plugin_handle` in the `end` block if needed:
```
end:
	if (need_free)
		free(plugin_name);
	if (plugin_filename)
		free(plugin_filename);
    if (plugin_handle && dlclose(plugin_handle)) {
        uwsgi_error("dlclose()");
    }

	return NULL;
}
```

Please feel free to get back to me and **reject any of the findings** described above if you believe they are false positives, or **discuss any of my suggested patches**. I will adjust the linked PR accordingly.

Thank you for your time and consideration.

Found by [Linux Verification Center](https://portal.linuxtesting.ru/) with SVACE

Canned-pineapple-8:
Fixes 2 handle leaks found by Svace static analyzer. See issue #2762 for details.

Commits:
1. `fix(socket): close serverfd after uwsgi_nuclear_blast()`
2. `fix(plugins): close plugin_handle on dlsym() failure`

Closes #2762 

If any fix is incorrect, please comment on the specific commit. I will update the PR accordingly.


Canned-pineapple-8:
Closed in favor of #2764 (branch was renamed).  Sorry for the mess...

